PR #84, opened 2018-12-17
Establishment api

wozzer72:
Morning Nasir

Firstly, please *do NOT delete* the branch.

I've not worked with `sequelize` before (first time with SQL databases and node.js - I've always used noSQL with node.js). So I got a head start on the faux login API, not least because the registration did not define the Establishment, Login and User types and their corresponding relationships, but had defined Services.

This pull request is my implementation for the faux login.  It does minimum validation on username, which it uses to resolve user, establishment and establishment's main service (if it is defined).

I've updated the JSON schema definition as I've gone along; not enforcing schema validation just yet though. Really need to separate the backend into its own repo first (not wanting to pollute the UI `package.json` with backend dependencies - which includes unit tests.

You'll be needing most of the changes here for your true `login` API.

No unit tests yet; I want to move onto employer type now, which includes validating Authorization header in a common utility (which you'll replace when the true `login` API is available).

I want to get some breadth across the backend services so Denny is playing catch up, rather than a waiting game, and so Maria/Pete can see agree the results before they go on leave for Xmas.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ To get more help on the Angular CLI use `ng help` or go check out the [Angular C
 # Backend Server
 Unfortunately, the backend server code is part of the client project & repo. Until refactoring:
 * `npm run build`
-* `node server` or `PORT=<p1> node server`
+* `node server` or `PORT=<p1> node server` or `nodemon server` or `PORT=<p1> nodemon server`
 
 Database connection parameters can be overridden using environment variables:
 * `DB_HOST` - hostname or IP address
--- a/resources/db/cqc_ddl.sql
+++ b/resources/db/cqc_ddl.sql
@@ -1,0 +1,478 @@
+--
+-- PostgreSQL database dump
+--
+
+-- Dumped from database version 11.0
+-- Dumped by pg_dump version 11.0
+
+SET statement_timeout = 0;
+SET lock_timeout = 0;
+SET idle_in_transaction_session_timeout = 0;
+SET client_encoding = 'UTF8';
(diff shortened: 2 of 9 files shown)


NMDSdevopsServiceAdm:
> Morning Nasir
> 
> Firstly, please _do NOT delete_ the branch.
> 
> I've not worked with `sequelize` before (first time with SQL databases and node.js - I've always used noSQL with node.js). So I got a head start on the faux login API, not least because the registration did not define the Establishment, Login and User types and their corresponding relationships, but had defined Services.
> 
> This pull request is my implementation for the faux login. It does minimum validation on username, which it uses to resolve user, establishment and establishment's main service (if it is defined).
> 
> I've updated the JSON schema definition as I've gone along; not enforcing schema validation just yet though. Really need to separate the backend into its own repo first (not wanting to pollute the UI `package.json` with backend dependencies - which includes unit tests.
> 
> You'll be needing most of the changes here for your true `login` API.
> 
> No unit tests yet; I want to move onto employer type now, which includes validating Authorization header in a common utility (which you'll replace when the true `login` API is available).
> 
> I want to get some breadth across the backend services so Denny is playing catch up, rather than a waiting game, and so Maria/Pete can see agree the results before they go on leave for Xmas.
> 
> Thanks
Morning Warren,

I have not worked with Sequalize earlier so i need to get the understanding as well

Yes you are the the folder structure need correcting front and back end but we stay for while to get this sprint done because it might break my CI/CD.

I will catch with you on login soon.

Thanks

PR #100, opened 2018-12-25
Add Current Vacancies

adamfaryna:
Hi Denny,

this is implementation of "Add Current Vacancies" page:
https://trello.com/c/Hp5HzbzG/146-add-current-vacancies-question-151-non-functional and 
https://trello.com/c/FFq0tywl/145-add-current-vacancies-question-151-front-end.

Please review. I hope you like it.

Adam


src/app/app-routing.module.ts
```diff
@@ -15,6 +15,10 @@ import { RegistrationCompleteComponent } from './features/registration-complete/
 import { EnterWorkplaceAddressComponent } from './features/enter-workplace-address/enter-workplace-address.component';
 import { SelectMainServiceComponent } from './features/select-main-service/select-main-service.component';
 import { ContinueCreatingAccountComponent } from './features/continue-creating-account/continue-creating-account.component';
+import { VacanciesComponent } from "./features/vacancies/vacancies.component";
+import { ConfirmVacanciesComponent } from "./features/confirm-vacancies/confirm-vacancies.component"
+import { AddNewStartersComponent } from "./features/add-new-starters/add-new-starters.component"
```
wozzer72: Regarding naming, starters and vacanies (along with leavers later) are the equivalent features; what's the significance of naming Starters with the prefix "AddNew" compared with not adding the prefix for Vacanies?

src/app/features/confirm-vacancies/confirm-vacancies.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  styleUrls: ['./confirm-vacancies.component.scss']
+})
+export class ConfirmVacanciesComponent implements OnInit {
+
+  constructor(private jobService: JobService) { }
+
+  private total = 0
+  
+  submitHandler() {
+      // this.establishmentService.postStaff(establishmentId, 0)
```
wozzer72: The workflow is such that on the "vacancies" screen (component) it should consume the `establishmentService` service posting to 'api/establishment/:id/jobs` posting JSON with the content:
```
{
  "jobs": {
     "vacancies": [
         {}. {}, ..., {}
     ]
  }
}
```

The "vacancies" component needs to total the number of vacancies across all job titles as application (client state) data, and on success of the post, navigate on to the "confirm vacancies" screen (component).

The "confirm vacancies" component to render needs to issue a GET to `api/establishment/:id/jobs` (via `establishmentService`) and read the JSON response attribute `totalVacancies`, comparing that value with the one stored in state from the "Vacancies" component.

There is a very explicit act of:
1. POST vacancies
2. GET and confirm vacancies

If on the "confirm vacancies" screen all is good and the numbers match, the next screen in the workflow is "Starters".
If on the "confirm vacancies" screen all is not good and the numbers do match, or the user thinks they have missed a job title, they can return back to the "vacancies" screen.

src/app/shared/number.directive.ts
```diff
@@ -0,0 +1,75 @@
+import { Input, Directive, ElementRef, HostListener } from "@angular/core"
```
wozzer72: I can now see where this relates to the generic "Number" in `src/app/app.module.ts` above. But the naming is still too generic. 

I can see how this is adding keyboard shortcuts into the 'input type=text' boxes, to make it easier for keyboard level adjustment of values.

Nice touch; although given the IT experience of the users, unlikely to be used and only introduces additional test effort.

PR #101, opened 2018-12-26
Feedback and minor update to "Local Authorities Share"

wozzer72:
Hey Nasir

Hans is at work today, so I've used this morning to finish of my APIs. It also means I'm totally available to help you with the login but also pair up with Adam on the UIs.

This PR includes the new feedback API.

It also includes a minor update to "Local Authorities Share" to return the primary Authority which is then shown on the UI.


nasirmids:
Hi Warren, thats great; i think we can go through my stuff tomorrow and i will than do pull request.

PR #102, opened 2018-12-27
Loginapi with passportjs and jwt implementation

nasirmids:
Hi Warren, 

This is the basic authentication implemented

- registration should now add hash values (i still keep password field temporarily)
- templogin replace with login; which validate uname/pwd and issue token. also build response copied from tmplogin
- set the header to estb id

What not??
currently  on proctecting the route.

kindly review and let me know what you think

Thanks


server/config/passport.js
```diff
@@ -0,0 +1,18 @@
+const JwtStrategy = require('passport-jwt').Strategy,
+    ExtractJwt = require('passport-jwt').ExtractJwt;
+
+// load up the user model
+const User = require('../models').User;
+
+module.exports = function(passport) {
+  const opts = {
+    jwtFromRequest: ExtractJwt.fromAuthHeaderWithScheme('JWT'),
+    secretOrKey: 'nodeauthsecret',
```
wozzer72: We cannot embed secrets within the code - the repo is an open respoitory.

server/models/login.js
```diff
@@ -51,14 +52,30 @@ module.exports = function(sequelize, DataTypes) {
...
   }, {
     tableName: '"Login"',
     schema: 'cqc',
     createdAt: false,
     updatedAt: false
   });
 
+  Login.prototype.comparePassword = function (passw, cb) {
+    console.log(this.password);
+    bcrypt.compare(passw, this.Hash, function (err, isMatch) {
```
wozzer72: Thought you were generating and storing a `salt` too?

server/routes/login.js
```diff
@@ -0,0 +1,93 @@
...
+
+            // check if this is the first time logged in and if so, update the "FirstLogin" timestamp
+            if (!login.firstLogin) {
+               login.update({
+                firstLogin: new Date()
+              });
+            }
+      
+          const headers = {
+            'Authorization': login.user.establishment.id
```
wozzer72: No. This needs to be the JWT itself. It is the token that must be passed back and forth, and to be OAuth compliant, the token must be prefixed with the word 'bearer'.

server/routes/registration.js
```diff
@@ -196,7 +196,7 @@ router.route('/')
     var EstablishmentSelect = 'SELECT * FROM cqc."Establishment" where "Name" = $1 Limit 1';
     var UserInsert = 'INSERT INTO cqc."User"("FullName", "JobTitle", "Email", "Phone", "DateCreated", "EstablishmentID", "AdminUser") VALUES ($1,$2,$3,$4,$5,$6,$7) RETURNING "RegistrationID"';
     var UserSelect = 'SELECT * FROM cqc."User" where "FullName" = $1 Limit 1';
-    var LoginInsert = 'INSERT INTO cqc."Login"("RegistrationID", "Username", "Password", "SecurityQuestion", "SecurityQuestionAnswer", "Active", "InvalidAttempt") VALUES ($1,$2,$3,$4,$5,$6,$7) RETURNING "ID"';
+    var LoginInsert = 'INSERT INTO cqc."Login"("RegistrationID", "Username", "Password", "SecurityQuestion", "SecurityQuestionAnswer", "Active", "InvalidAttempt", "Hash") VALUES ($1,$2,$3,$4,$5,$6,$7,$8) RETURNING "ID"';
```
wozzer72: About the generating and storing the `salt`?

wozzer72: And we must not store the password - it is a massive faux pas.

server/routes/registration.js
```diff
@@ -265,7 +265,8 @@ router.route('/')
       try {
         // forced error - in absence of unit tests
         //throw new Error("Totally forced")
-        const result = await client.query(LoginInsert,[registrationID, Logindata.UserName, Logindata.Password,Logindata.SecurityQuestion,Logindata.SecurityQuestionAnswer,1,0]);
+        const hash = await bcrypt.hashSync(Logindata.Password, bcrypt.genSaltSync(10), null);
+        const result = await client.query(LoginInsert,[registrationID, Logindata.UserName, Logindata.Password,Logindata.SecurityQuestion,Logindata.SecurityQuestionAnswer,1,0, hash]);
```
wozzer72: Don't pass "password".

server/routes/tmpLogin.js
```diff
@@ -84,7 +84,7 @@ router.route('/').post(async (req, res) => {
 
         // check if this is the first time logged in and if so, update the "FirstLogin" timestamp
         if (!results.firstLogin) {
-          await results.update({
+           results.update({
```
wozzer72: You need the await, else the function will return before the update has completed

server/routes/login.js
```diff
@@ -0,0 +1,93 @@
...
+ const passport = require('passport');
+var router = express.Router();
+require('../config/passport')(passport);
+const Login = require('../models').login;
+// console.log("this is console user");
+// //console.log(models.user.username);
+
+router.post('/',async function(req, res) {
+  console.log('inside login');
+  await Login
```
wozzer72: await is not necessary because you are explicitly handling the promise (with `.then`)

server/models/login.js
nasirmids: bcrypt insert the salt with hash first time hence not needed  ; i previously using node crypto package which need salt.  I will remove the salt  and password field when we have happy with this solution.

server/routes/login.js
wozzer72: Also, the JWT needs 'claims' defined within it:
1. EstablishmentID
2. username
3. isAdmin

This will become the impudicated facts that the API can call upon later.

server/routes/registration.js
nasirmids: i reply above salt not needed

server/routes/registration.js
nasirmids: ok it is temp; i will tidy up

server/routes/tmpLogin.js
nasirmids: gotchas

server/routes/login.js
wozzer72: The token also needs an expiry. A default expiry but also allow for an env variable to override it. A default over 12 hours would be good to make it easy to work with in dev.

server/config/passport.js
```diff
@@ -0,0 +1,20 @@
+const JwtStrategy = require('passport-jwt').Strategy,
```
wozzer72: This file is not config, but is authentication utility. It would sit better under `.../util/security`.

server/routes/login.js
```diff
@@ -0,0 +1,101 @@
+ var express = require('express');
+ const models = require('../models/index');
+ const jwt = require('jsonwebtoken');
+ const passport = require('passport');
+var router = express.Router();
+require('../config/passport')(passport);
+const Login = require('../models').login;
+
+Token_Secret = process.env.Token_Secret ? process.env.Token_Secret : "nodeauthsecret";
```
wozzer72: Ideally, would like to only see the processing of the env variable once - duplication of code does not help (what is somebody changed the name of the ENV vairable - but only changed it in one place?).

Any chance you can encapsulated within the `server/config/passport.js` export above.

server/config/passport.js
```diff
@@ -0,0 +1,20 @@
...
+    ExtractJwt = require('passport-jwt').ExtractJwt;
+
+// load up the user model
+const User = require('../models').User;
+
+Token_Secret = process.env.Token_Secret ? process.env.Token_Secret : "nodeauthsecret";
```
wozzer72: If you make this an export, you can reuse through imports.

server/routes/login.js
```diff
@@ -0,0 +1,101 @@
...
+      .then((login) => {
+        if (!login) {
+          return res.status(401).send({
+            message: 'Authentication failed. User not found.',
+          });
+        }
+    login.comparePassword(req.body.password, (err, isMatch) => {
+      if (isMatch && !err) {
+
+        var claims = {
```
wozzer72: This will have to do for now, but with OAuth JWT, there are specific fields that are required in a JWT:
* `sub` - this would be the username
* `iss` - this is a unique identifier for the endpoint - typically, the URL
* `exp` - this is given by the 'expiresIn' parameter - I presume
* `aud` - self identity, e.g. ADS-WDS.

Attributes we can explicitly ignore are:
* `nbf`
* `iat`
* `jti`

server/routes/login.js
```diff
@@ -0,0 +1,101 @@
...
+  return {
+    fullname,
+    isFirstLogin: firstLoginDate ? false : true,
+    establishment: {
+      id: establishment.id,
+      name: establishment.name,
+      isRegulated: establishment.isRegulated
+    },
+    mainService: {
+      id: mainService ? mainService.id : null,
```
wozzer72: We need to return 'expiry' here too; best done as an ISO 8601 date format (or JSON/XML format as I call it).

server/utils/security/isAuthenticated.js
```diff
@@ -1,11 +1,27 @@
-const AUTH_HEADER='authorization';
+const jwt = require('jsonwebtoken');
+const AUTH_HEADER = 'authorization';
+
+Token_Secret = process.env.Token_Secret ? process.env.Token_Secret : "nodeauthsecret";
```
wozzer72: Again. Import from the `server/config/passport.js` module.

server/utils/security/isAuthenticated.js
```diff
@@ -17,22 +33,37 @@ exports.isAuthorised = (req, res , next) => {
...
-      console.error('isAuthenticated - missing establishment id parameter');
-      return res.status(400).send(`Unknown Establishment ID: ${req.params.id}`);
-    }
-    if (req.establishmentId !== parseInt(req.params.id)) {
-      console.error('isAuthenticated - given and known establishment id do not match');
-      return res.status(403).send(`Not permitted to access Establishment with id: ${req.params.id}`);
+    let token = req.headers[AUTH_HEADER];
+
+    if (token.startsWith('Bearer')) {
+      token = token.slice(7, token.length);
```
wozzer72: I'm guilty of this all the time;l but should factor this repetition of code in to a local function.

server/utils/security/isAuthenticated.js
```diff
@@ -17,22 +33,37 @@ exports.isAuthorised = (req, res , next) => {
...
+      if (err) {
+        return res.json({
+          sucess: false,
+          message: 'token is invalid'
+        });
+      } else {        
+
+        req.establishmentId = data.establishmentId; //parseInt(req.headers[AUTH_HEADER]);     
+        // must provide the establishment ID and it must be a number
+        if (!req.params.id || isNaN(parseInt(req.params.id))) {
```
wozzer72: This needs to be the `claim.EstblishmentId` taken from the token.

server/routes/login.js
NMDSdevopsServiceAdm: i have seen those variable but will make it more tidy as we go along

server/utils/security/isAuthenticated.js
```diff
@@ -1,11 +1,31 @@
...
-  if (req.headers[AUTH_HEADER] && Number.isInteger(parseInt(req.headers[AUTH_HEADER]))) {
-    next();
+  let token = getToken(req.headers[AUTH_HEADER]);
+
+  if (token) {
+    var dec = getverify(token, Token_Secret);
+
+    jwt.verify(token, Token_Secret, function (err, claim) {
+      if (err) {
+        return res.json({
```
wozzer72: YOu must set the header error code not just the response body!

server/utils/security/isAuthenticated.js
```diff
@@ -1,11 +1,31 @@
...
+          sucess: false,
+          message: 'token is invalid'
+        });
+      } else {
+        req.establishmentId = claim.EstblishmentId;
```
wozzer72: The claims are not needed of isAuthenticated.

server/utils/security/isAuthenticated.js
```diff
@@ -15,24 +35,52 @@ exports.isAuthorised = (req, res , next) => {
...
-      console.error('isAuthenticated - given and known establishment id do not match');
-      return res.status(403).send(`Not permitted to access Establishment with id: ${req.params.id}`);
-    }
+  let token = getToken(req.headers[AUTH_HEADER]);
   
-    next();
+  if (token) {
+    jwt.verify(token, Token_Secret, function (err, claim) {
+      if (err) {
+        return res.json({
```
wozzer72: Set the http code to 401.

PR #111, opened 2018-12-28
Functional 'staff' component, although needs some further work on error messaging for validation

wozzer72:
Wanting to merge back down to keep the common branch in good shape.

### DIFF
--- a/src/app/features/staff/staff.component.scss
+++ b/src/app/features/staff/staff.component.scss
@@ -1,0 +1,4 @@
+.page-comment {
+    color: darkgrey
+}
+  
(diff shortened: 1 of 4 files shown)


src/app/features/staff/staff.component.scss
```diff
@@ -0,0 +1,4 @@
+.page-comment {
+    color: darkgrey
```
adamfaryna: We use color codes and variables. Check variables.scss.

wozzer72: Good catch and thank you for the suggestion

PR #117, opened 2018-12-31
Establishment ui login

wozzer72:
This pulls in the login and welcome.

After welcome it jumps to 'other services'; it will need to jump to your employer type once done.

And this includes a workaround for not being able to get the headers in the post response to `api/login`; have therefore assumed the authorization header. One for Denny on his return.

And finally, rather than introducing the Angular intercepter, because you had alreadyd encapsulated the header in your `establishment` service and it is the only place we need to pass the header, I've simply updated the establishment service to be given an auth token.

Have ran through all remaining screens 'other services', 'share', 'share with LA', 'vacancies', 'starters', 'leavers' and 'staff'. All updating via the backend.


src/app/core/services/auth-service.ts
```diff
@@ -15,25 +15,88 @@ const initialRegistration: LoginApiModel = {
...
 
 
-  constructor(private http: HttpClient, private router: Router) { }
+  constructor(private http: HttpClient,
+              
+              private router: Router) { }
+
+  // returns true if logged in; otherwise false
+  public get isLoggedIn(): boolean {  
+    return this._session ? true : false;
```
adamfaryna: Can we just `return !!this._session` ?

src/app/core/services/auth-service.ts
```diff
@@ -15,25 +15,88 @@ const initialRegistration: LoginApiModel = {
...
+      return false;
+    }
+  }
+
+  // sets the session token - note; this should proberly be saved to local storage - but also with an expiry.
+  public set token(authorization:string) {
+    this._token = authorization;
+  }
+
+  // return establishment
```
adamfaryna: clean up pls

src/app/features/login/login.component.ts
```diff
@@ -76,16 +76,27 @@ export class LoginComponent implements OnInit {
...
-
-            this._loginService.updateState(data);
-
+        (response) => {
+          // TODO: despite passing the 'observe' option
+          //       in the auth-service postLogin, the callback
+          //       here is still the JSON data and not the full
+          //       response. Hence, cannot get at the
+          //       headers.
+          console.log("DEBUG response: ", response)
```
adamfaryna: No debug code in repo pls.

src/app/features/select-other-services/select-other-services-list/select-other-services-list.component.ts
```diff
@@ -138,10 +137,9 @@ export class SelectOtherServicesListComponent implements OnInit {
           console.log(err);
         },
         () => {
-          console.log('Successfully posted other services');
-          this.router.navigate(['/type-of-employer']);
+          // TODO - update navigation to "service capacity"
+          this.router.navigate(['/shareOptions']);
```
adamfaryna: It's '/share-options' now.

src/app/core/services/auth-service.ts
wozzer72: Done

src/app/features/login/login.component.ts
wozzer72: Done

src/app/features/select-other-services/select-other-services-list/select-other-services-list.component.ts
wozzer72: But it's not on this branch. When it's merged into `establishment_ui` it will pick up on your changes.

src/app/core/services/auth-service.ts
wozzer72: that will require testing, and I want Denny to review and update the code accordingly to use those fancy Observables he was working on.

PR #119, opened 2018-12-31
Employer type

adamfaryna:


src/app/core/services/establishment.service.ts
```diff
@@ -28,6 +28,12 @@ interface ShareWithLocalAuthorityResponse extends EstablishmentApiResponse {
   localAuthorities: LocalAuthorityModel[];
 };
 
+interface EmployerTypeResponse {
+  id: number
+  name: string
+  employerType: string
```
wozzer72: Ideally, this should be a typescript enum: https://www.typescriptlang.org/docs/handbook/enums.html. But leave as is, I've raised an issue to go back and correct.

src/app/core/services/establishment.service.ts
```diff
@@ -202,4 +209,24 @@ export class EstablishmentService {
...
+      .pipe(
+        debounceTime(500),
+        catchError(this.httpErrorHandler.handleHttpError))
+  }
+
+  /*
+   * POST /api/establishment/:establishmentId/employerType
+   */
+  postEmployerType(data) {
+    return this.http.post<EmployerTypeResponse>(`/api/establishment/${this.establishmentId}/employerType`, data, this.getOptions())
```
wozzer72: For type safely, would be best to define an EmployerTypeRequest interface. I'll add it to issue to tidy up later.

PR #123, opened 2019-01-02
Fixed deployment issues

adamfaryna:
Fixed styles & establishmentId & establishmentToken & API call issues.


src/app/features/vacancies/vacancies.component.ts
```diff
@@ -106,7 +106,7 @@ export class VacanciesComponent implements OnInit, OnDestroy {
 
     this.subscriptions.push(
       this.establishmentService.getVacancies().subscribe(vacancies => {
-        if (vacancies) {
+        if (vacancies.length) {
```
wozzer72: Why only vacancies? Habit suggests this is leavers and starters too.

adamfaryna: Good point, I fixed there as well.

PR #127, opened 2019-01-02
Adam couple small fixes

adamfaryna:


src/app/core/footer/footer.component.html
```diff
@@ -9,7 +9,7 @@
         </div>
 
         <ul class="footer-nav-list col-12 col-lg-8">
-          <li class="d-inline-block"><a routerLink="/" title="Help">Help</a></li>
+          <li class="d-inline-block"><a routerLink="/" skipLocationChange="true" title="Help">Help</a></li>
```
wozzer72: I don't understand why help returns the user back to login? Was this feedback from yesterday's review?

PR #129, opened 2019-01-02
uat related config changes

NMDSdevopsServiceAdm:
Hi Warren, 

Kindly review; few UAT related changes needed in config files. 

Thanks
Nasir


NMDSdevopsServiceAdm:
Hi Warren, I have not added the gov paas in to config.json and they were just dummy password already there. i only add hostname and few others. I will still define all of them as env variable in gov paas as discussed. 

Secondly the gov paas i dont think provide AWS secret manager (they dont provide full fledged aws features out of box as of now but i will cross check it again with them) and we have you use our aws instance for those purposes.

PR #142, opened 2019-01-03
Version 1.0 from Test merge into Master

NMDSdevopsServiceAdm:


NMDSdevopsServiceAdm:
This is the first ever release and It has following major modules

- Registration
- Login
- Establishment

PR #154, opened 2019-01-05
IE11 Compatibility (Functional)

wozzer72:
I've been through various scenarios testing IE11 functionally against that with Chrome and Firefox.

The initial set of errors Adam fixed with the includsion of "core-js" module and I fixed with forced "no-cache" headers on transactional data (strangely enough only IE11 was demonstrating caching of data even though the others should).

The introduce of "core-js" had a bizarre impact on the solution I initially adopted on login to store the authorisation header (you'll need to read the Trello ticket for reasoning - https://trello.com/c/DVZYvkaG). That shortcut though had to be removed, so I've implemented the required HttpInterceptor as part of this ticket. And yes, I have intentionally left in a comment with a todo in the login component that we need to introduce an expiry of the token in to local storage too - I've already fed back necessity on token expiry to Nasir for the formal backend login API. It's not yet available hence the TODO comment.

The Vacancy, Starters and Leavers, not technically an IE11 compatibility issue, simply required matching the default 'null' form control created in `ngOnInit` with the options drop in the HTML.

And finally, the "enter-workplace-address" observation that I thought was IE11, is not just IE11. It's across all browsers, so I've raised a subsequent trello ticket (https://trello.com/c/UeJ1oUxo).

I'm not a visual man, but visually, having worked through various scenarios, it's not a bad looker compared to Chrome and Firefox. Consequently, I'm raising this PR to get the code reviewed and merged in to `develop` branch and deployed into DEV, so we can put it in Ann's more thorough hands to see if we're confident to be able to push through TEST into UAT and have IE11 included in the beta trial (which the client would very much like).

Thanks

### DIFF
--- a/src/app/core/services/auth-interceptor.ts
+++ b/src/app/core/services/auth-interceptor.ts
@@ -1,0 +1,34 @@
+import { Injectable } from '@angular/core';
+import {
+  HttpRequest,
+  HttpHandler,
+  HttpEvent,
+  HttpInterceptor
+} from '@angular/common/http';
+import { Observable } from 'rxjs/Observable';
+
+@Injectable()
+export class AuthInterceptor implements HttpInterceptor {
+
+  constructor() {}
+
+  intercept(request: HttpRequest<any>,
+            next: HttpHandler): Observable<HttpEvent<any>> {
+    
+    // read the token from local storage
+    const idToken = localStorage.getItem("auth-token");
+
+    if (idToken) {
+        // TODO: add bearer upon integrating JWT server side auth
+        const cloned = request.clone({
+            //headers: request.headers.set("Authorization", "Bearer " + idToken);
+            headers: request.headers.set("Authorization", idToken)
+        });
+
+        return next.handle(cloned);
+    }
+    else {
+        return next.handle(request);
+    }
+  }
+}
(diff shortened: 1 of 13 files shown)


src/app/core/services/auth-interceptor.ts
```diff
@@ -0,0 +1,34 @@
...
+@Injectable()
+export class AuthInterceptor implements HttpInterceptor {
+
+  constructor() {}
+
+  intercept(request: HttpRequest<any>,
+            next: HttpHandler): Observable<HttpEvent<any>> {
+    
+    // read the token from local storage
+    const idToken = localStorage.getItem("auth-token");
```
adamfaryna: We don't have to access localStorage every time we need token. Because of the performance reasons, we should reach localStorage only if the token is not cached yet. Please add cache for token.

src/app/features/login/login.component.ts
```diff
@@ -88,8 +88,13 @@ export class LoginComponent implements OnInit {
 
           // // update the establishment service state with the given establishment oid
           this.establishmentService.establishmentId = response.establishment.id;
+
+          // store the authorization token
+          // TODO: add expiry of token const expiresAt = moment().add(response.expiresIn,'second');
+          localStorage.setItem("auth-token", response.establishment.id)
```
adamfaryna: This operation should be handled by some service, at least similar as we do which `establishmentId` above. Component controller shouldn't be involved directly in any operations on storages, but use services instead.

PR #161, opened 2019-01-07
Handling DB SSL as a environment variable override

wozzer72:
HI Nasir

This is to backfill the SSL option for database config - to be controlled as env variable.

Note, the Trello card is: https://trello.com/c/OoE63MqJ. There is a task on there for you to update the Travis deployment to pass `DB_SSL=1` to UAT. You can (and should) do that in advance of a deploy to UAT. Adding it now has no adverse consequence.


wozzer72:
Note, no changes are required to travis job after - the env variables are set in UAT using cloud factory command line. Trello card has been updated.

PR #181, opened 2019-01-10
337 api error handling, headers setting interceptor

adamfaryna:
I created generic interceptor invoked at the beginning of the interception chain. It sets all the common headers, debounceTime, as well as error handling for us, so we can just write basic API call and everything else is covered.


adamfaryna:
> I like the idea, as it simplifies the code. I can see it has been applied to Establishment, Jobs (reference service) and Feedback (service).
> 
> But there are more services that are interacting with the API not yet covered. I'm not comfortable approving this until all services have been tackled, and not without a more thorough review by Denny too.
> 
> Also, there is no Trello card associated with this work, and the impact on the deployed solution could be greater than expected, without significant testing. Consequently, I also would not recommend this being merged as a bugfix, but instead, tested more thoroughly under the 'workers' epic.

I agree, it has to be tested and I did that already, testing our app back and forth, through all the screens.

adamfaryna:
[API error handling, headers setting interceptor](https://trello.com/c/3klHa0aV/337-api-error-handling-headers-setting-interceptor)

adamfaryna:
> So I have to agree with Warren..
> 
> * Firstly there is no ticket for this work so that will need to be created and approved before any this can be, as we shouldn't be working on ad-hock stuff.
> * Secondly without properly testing it, I cannot approve it. The reason being is it impacts across so much, and will need to be tested across the whole application.

* The related ticket is created https://trello.com/c/3klHa0aV.

* How this testing should look like, how can we measure the results and who should proceed with them?

wozzer72:
Following sprint planning, we decided to close this PR without merging any changes, because the work is only part done.

PR #183, opened 2019-01-11
Bugfix/update registration workplace validation

denny13:
Updating validation on registration 'workplace name' field from 40 to 120 max-length

### DIFF
--- a/src/app/features/enter-workplace-address/enter-workplace-address.component.html
+++ b/src/app/features/enter-workplace-address/enter-workplace-address.component.html
@@ -120,7 +120,7 @@
 
           <span class="invalid-feedback">
             <span *ngIf="submittedWpNameInput && getWpName.hasError('maxlength')" class="main-body-text mb-2">
-              The Workplace name must be no longer than 40 characters.
+              The Workplace name must be no longer than 120 characters.
             </span>
             <span *ngIf="submittedWpNameInput && getWpName.hasError('required')" class="main-body-text mb-2">
               Please enter your Workplace name.
--- a/src/app/features/enter-workplace-address/enter-workplace-address.component.ts
+++ b/src/app/features/enter-workplace-address/enter-workplace-address.component.ts
@@ -274,21 +274,23 @@ export class EnterWorkplaceAddressComponent implements OnInit {
 
   loadExistingValues() {
 
-    if ((this.registration.locationdata[0].hasOwnProperty('locationName')) && (this.registration.locationdata[0].locationName === '')) {
-      const postcodeValue = this.registration.locationdata[0].postalCode;
-      const address1Value = this.registration.locationdata[0].addressLine1;
-      const address2Value = this.registration.locationdata[0].addressLine2;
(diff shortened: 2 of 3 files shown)


src/app/features/enter-workplace-address/enter-workplace-address.component.ts
```diff
@@ -274,21 +274,23 @@ export class EnterWorkplaceAddressComponent implements OnInit {
...
-
-      this.enterWorkplaceAddressForm.setValue({
-        postcodeInput: postcodeValue,
-        address1Input: address1Value,
-        address2Input: address2Value,
-        townCityInput: townCityValue,
-        countyInput: countyValue,
-        wpNameInput: '',
-      });
+    if (this.registration.hasOwnProperty('locationdata')) {
```
wozzer72: I know why you has introduced this, because when coming from non-CQC registration, locationdata is null. This fix stops the error in the console, but it doesn't solve the problem of when coming to this component from non-CQC, you can't add a manual address.

src/app/features/enter-workplace-address/enter-workplace-address.component.html
```diff
@@ -120,7 +120,7 @@ <h1 class="h1">Enter your workplace address</h1>
 
           <span class="invalid-feedback">
             <span *ngIf="submittedWpNameInput && getWpName.hasError('maxlength')" class="main-body-text mb-2">
-              The Workplace name must be no longer than 40 characters.
+              The Workplace name must be no longer than 120 characters.
```
wozzer72: I presume it was just the error message that was incorrect? Please confirm that the field validation does indeed enforce up to 120 characters.

src/app/features/enter-workplace-address/enter-workplace-address.component.ts
```diff
@@ -274,21 +274,23 @@ export class EnterWorkplaceAddressComponent implements OnInit {
...
+      if ((this.registration.locationdata[0].hasOwnProperty('locationName')) && (this.registration.locationdata[0].locationName === '')) {
```
adamfaryna: Similar as above. Checking property if the property value is provided by using `hasOwnProperty` is over verbose. Simple `this.registration.locationdata[0].locationName` will do the job.

Also `locationdata` should be camel-case `locationData`.

src/app/features/enter-workplace-address/enter-workplace-address.component.ts
wozzer72: @adamfaryna . Please try the following:
1. Create an Account on login page.
2. Select "no" to CQC.
3. Enter postcode and search.
4. Then select the "I can't find my address"; which takes you to the "enter-workplace-address".
5. Enter some address details.
6. Open the browser debugger and put a breakpoint in `webpack:///src/app/features/enter-workplace-address/enter-workplace-address.component.ts::save` - around line 315.
7. Press "Continue".

You'll see locationdata is undefined. As per the screenshot attached.
image

You'll see

PR #188, opened 2019-01-15
339-log-out-user-is-allowed-to-navigate-back-into-the-establishment-f…

adamfaryna:
https://trello.com/c/vOJpPknP/339-log-out-user-is-allowed-to-navigate-back-into-the-establishment-flow-using-the-browser-back-button-without-logging-in-again


wozzer72:
Merging down to `develop` branch because this change is a direct result of fix currently under test.

PR #189, opened 2019-01-15
Merge develop into worker_epic.

adamfaryna:
Our `develop` branch have some nice features that we should have available in `worker_epic`, at least: authentication & authenticated routing.


adamfaryna:
This one doesn't make sense, I created new one https://github.com/NMDSdevopsServiceAdm/SopraSteria-SFC/pull/190.

PR #192, opened 2019-01-16
107-google analytics integration

nasirmids:
Hi Warren, 

Another for review and below is the summary

- Three GA tracking for dev, test and uat
- Initially i tried to put into custom js file as we discussed but there is script tag in the tracking code didnt trigger if i put this in external js file <script async src="https://www.googletagmanager.com/gtag/js?id=UA-132324399-1"></script>
- Denny also had a look but didnt find workable solution.
- Alternatively, There is custom js code where you can call the js file via js code rather via script tag above but that didnt work with our angular like this below
     // var s = document.createElement('script');
    // s.type = 'text/javascript';
    // s.async = true;
    // s.src = 'https://www.googletagmanager.com/gtag/js?id=UA-132324399-1';
    // var x = document.getElementsByTagName('script')[0];
    // x.parentNode.insertBefore(s, x);

- finally, i stop spending much time and i have created 3 index files and in the build i ran a mv command to rename it by environment/branch and it works well. 

The GA dashboard runs off gmail account NMDSdevopsServiceAdm@gmail.com and we can see who is active.

If i find some solution i will do refactoing of that.

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,8 @@ before_deploy:
     #- if [ "$TRAVIS_BRANCH"=develop ]; then B_NAME=develop; fi
     #- if [ "$TRAVIS_BRANCH"=test ]; then B_NAME=test; fi
     #- echo $B_NAME
-   - if [ "$TRAVIS_BRANCH" = master ]; then echo "master"; mv index_uat.html index.html; fi
-   - if [ "$TRAVIS_BRANCH" = test ]; then echo "test"; mv index_staging.html index.html; fi
+   #- if [ "$TRAVIS_BRANCH" = master ]; then echo "master"; mv index_uat.html index.html; fi
+   #- if [ "$TRAVIS_BRANCH" = test ]; then echo "test"; mv index_staging.html index.html; fi
  deploy:
   - provider: cloudfoundry
     edge: true  
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ before_deploy:
     #- echo $B_NAME
    #- if [ "$TRAVIS_BRANCH" = master ]; then echo "master"; fi
    #- if [ "$TRAVIS_BRANCH" = test ]; then echo "test"; fi
-   - if [ "$TRAVIS_BRANCH" = 107_google_analytics_dev ]; then echo "107_google_analytics_dev"; fi
+   - if [ "$TRAVIS_BRANCH" = 107_google_analytics_dev ]; then echo "107_google_analytics_dev"; mv index_staging.html index.html  fi
 deploy:
   - provider: cloudfoundry
     edge: true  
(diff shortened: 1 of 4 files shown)


adamfaryna:
If we would have something like [dotenv](https://www.npmjs.com/package/dotenv), run variables or env variables we could take leverage of placeholders. The base concept was the best - injecting script element dynamically from the code. 

This solution which moving html around is quite invasive, though.

nasirmids:
> If we would have something like [dotenv](https://www.npmjs.com/package/dotenv), run variables or env variables we could take leverage of placeholders. The base concept was the best - injecting script element dynamically from the code.
> 
> This solution which moving html around is quite invasive, though.

@adamfaryna  correct Adam i have seen DotEnv but not in Anger and can do something like this going forward when i do refactoring at some point.  The first thing would be make the async script work then think about replacement of variable with this dotenv . does it make sense?

.travis.yml
```diff
@@ -47,25 +47,13 @@ script:
...
 before_deploy:    
-    #- if [ "$TRAVIS_BRANCH"=master ]; then B_NAME=master; fi
+    - if [ "$TRAVIS_BRANCH"=master ]; then B_NAME=master; fi
     #- if [ "$TRAVIS_BRANCH"=develop ]; then B_NAME=develop; fi
     #- if [ "$TRAVIS_BRANCH"=test ]; then B_NAME=test; fi
     #- echo $B_NAME
-   #- if [ "$TRAVIS_BRANCH" = master ]; then echo "master"; fi
-   #- if [ "$TRAVIS_BRANCH" = test ]; then echo "test"; fi
-   #- if [ "$TRAVIS_BRANCH" = develop ]; then echo "develop"; fi
+    - if [ "$TRAVIS_BRANCH" = master ]; then echo "master"; mv index_uat.html index.html; fi
```
wozzer72: And whatabout production? Are we expecting an index_prod.html?

wozzer72: I take it we're not using Google Analytics in dev then? What about when we want to demonstrate specific analytic traps going forward?

index.html
```diff
@@ -12,5 +12,12 @@
 </head>
 <body>
   <app-root></app-root>
-  
+  <script async src="https://www.googletagmanager.com/gtag/js?id=UA-132324399-1"></script>
```
wozzer72: The tag here is the only thing changing yes?

.travis.yml
wozzer72: Or is dev implied by the default `index.html`???

nasirmids: my fault; the default index file contains dev tracking code

index.html
nasirmids: yes thats right; as discussed there are two ways
1- either i replace it as text  from travis build
2- use angular plugin

I opt for 1 one.

.travis.yml
nasirmids: well; hopefully this is not needed when i do proper refactoring and we get rid of all files.

PR #200, opened 2019-01-17
Rebaseline worker_epic with latest changes from `develop`

wozzer72:
Significanty, this is to include database schema changes that break the code from attaching to the central DEV and TST databases (namely, the dropping of `location.cqcid` column).

I've now also applied those database changes to my local database instances, and consequently, need the reflecting backend model in the worker_epic branch. Given you're now also creating a local database instance (from a backup of dev db yesterday), you'll also need the latest backend sequelize model.

Please review and approve with some urgency as I am working at risk in my branch with localised changes.

Thanks


adamfaryna:
> Significanty, this is to include database schema changes that break the code from attaching to the central DEV and TST databases (namely, the dropping of `location.cqcid` column).
> 
> I've now also applied those database changes to my local database instances, and consequently, need the reflecting backend model in the worker_epic branch. Given you're now also creating a local database instance (from a backup of dev db yesterday), you'll also need the latest backend sequelize model.
> 
> Please review and approve with some urgency as I am working at risk in my branch with localised changes.
> 
> Thanks

I see merge of entire current `develop` branch is included here as well. I believe you though about `server/models/location.js` change only which could be cherry-picked, but it's fine to have child branch in sync which parent.

PR #201, opened 2019-01-17
Added gender and disablility extended properties

wozzer72:
Extended properies for gender and disability.

This observes the latest UI screen for gender which has descope the 'other' drop down.


adamfaryna:
All good and neat.

PR #204, opened 2019-01-18
Generic 404 page.

adamfaryna:


wozzer72:
This one needs to be reviewed with Maria and Richard before it is merged, as there is no requirement/acceptance criteria defined.

adamfaryna:
It will be handy on Sprint 6, when we plan to implement 404 error page.

PR #208, opened 2019-01-21
Worker Change and Audit logging

wozzer72:
This is a refactoring of the extended property framework and the existing #10 Worker properties to record saved/change activity by property and at Worker level, in addition to recording audit events (again by property and at Worker level).

Within this scope is the retrieval of a Worker with four different presentations:


server/models/classes/worker/properties/contractProperty.js
```diff
@@ -1,40 +1,55 @@
...
     }
 
-    toJSON() {
-        return {
-            contract: this.property
-        }
+    isEqual(currentValue, newValue) {
+        // contract is a simple (enum'd) string
+        if (currentValue && newValue && currentValue === newValue) return true;
+        else return false;
```
adamfaryna: This can be one-liner `return currentValue && newValue && currentValue === newValue`

server/models/classes/worker/properties/contractProperty.js
```diff
@@ -1,40 +1,55 @@
...
+                contract: this.property
+            }
+        } else {
+            return {
+                contract : {
+                    currentValue: this.property,
+                    ... this.changePropsToJSON(showPropertyHistoryOnly)
+                }
+            }
+        }
```
adamfaryna: When we have `if` with `else`, it's better to check if the value is true first and left false for `else`. It's small performance improvement and readiness.

server/models/classes/worker/properties/dateOfBirthProperty.js
```diff
@@ -25,35 +25,42 @@ exports.WorkerDateOfBirthProperty = class WorkerDateOfBirthProperty extends Prop
...
+        };
     }
 
-    get valid() {
-        return this.property && this.property.isValid();
+    isEqual(currentValue, newValue) {
+        // a moment date
+        if (currentValue && newValue && currentValue.isSame(newValue, 'day')) return true;
+        else return false;
+    }
```
adamfaryna: This can be one-liner `return currentValue && newValue && currentValue.isSame(newValue, 'day')`.

server/models/classes/worker/properties/dateOfBirthProperty.js
```diff
@@ -25,35 +25,42 @@ exports.WorkerDateOfBirthProperty = class WorkerDateOfBirthProperty extends Prop
...
+                dateOfBirth: this.property.toJSON().slice(0,10)
+            }
+        } else {
+            return {
+                dateOfBirth : {
+                    currentValue: this.property.toJSON().slice(0,10),
+                    ... this.changePropsToJSON(showPropertyHistoryOnly)
+                }
+            }
+        }
```
adamfaryna: As mentioned 2 comments above.

server/models/classes/worker/properties/disabilityProperty.js
```diff
@@ -1,40 +1,55 @@
...
+        };
     }
 
-    toJSON() {
-        return {
-            disability: this.property
-        }
+    isEqual(currentValue, newValue) {
+        // a simple (enum'd) string compare
+        if (currentValue && newValue && currentValue === newValue) return true;
```
adamfaryna: As mentioned in comments above.

server/models/classes/worker/properties/disabilityProperty.js
```diff
@@ -1,40 +1,55 @@
...
+                disability: this.property
+            }
+        } else {
+            return {
+                disability : {
+                    currentValue: this.property,
+                    ... this.changePropsToJSON(showPropertyHistoryOnly)
+                }
+            }
+        }
```
adamfaryna: As mentioned in comments above.

server/models/classes/worker/properties/mainJobProperty.js
```diff
@@ -1,60 +1,67 @@
...
     }
 
-    toJSON() {
-        return {
-            mainJob: this.property
-        }
+    isEqual(currentValue, newValue) {
+        // main job is an object where jobId is the primary key
+        if (currentValue && newValue && currentValue.jobId === newValue.jobId) return true;
+        else return false;
```
adamfaryna: As mentioned in comments above.

server/models/classes/worker/properties/postcodeProperty.js
```diff
@@ -1,41 +1,59 @@
...
 
-    toJSON() {
-        return {
-            postcode: this.property
-        }
+
+    isEqual(currentValue, newValue) {
+        // a simple string compare
+        if (currentValue && newValue && currentValue === newValue) return true;
+        else return false;
```
adamfaryna: Could be one-liner.

adamfaryna:
I went through this PR. I saw a lot of repeated code, that it looks like could be externalised to common class. I focused mostly on semantic because I don't know entire picture of this solution.

server/models/classes/worker/properties/contractProperty.js
wozzer72: That is true; I will reduce because it makes this easier to read and it leverages on to the properties that are yet to be written.

I repeat this accordingly across all properties.

server/models/classes/worker/properties/contractProperty.js
wozzer72: I didn't know it was a performance hit. But with a backend database, shaving a couple of clocks off the response time will not see much difference. But I like the suggestion as the code does become easier to read, and it's a good principle for the properties yet to write.

Again. I'll repeat on all properties.

server/models/classes/worker/properties/dateOfBirthProperty.js
wozzer72: Code changed accordingly.

server/models/classes/worker/properties/disabilityProperty.js
wozzer72: Code changed accordingly.

server/models/classes/worker/properties/disabilityProperty.js
wozzer72: Code changed accordingly.

server/models/classes/worker/properties/genderProperty.js
```diff
@@ -1,40 +1,55 @@
...
     }
 
-    toJSON() {
-        return {
-            gender: this.property
-        }
+    isEqual(currentValue, newValue) {
+        // gender is a simple (enum'd) string
+        if (currentValue && newValue && currentValue === newValue) return true;
+        else return false;
```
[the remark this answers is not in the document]
wozzer72: Code changed accordingly.

server/models/classes/worker/properties/mainJobProperty.js
wozzer72: Code changed accordingly.

server/models/classes/worker/properties/mainJobStartDateProperty.js
```diff
@@ -39,35 +42,41 @@ exports.WorkerMainJobStartDateProperty = class WorkerMainJobStartDateProperty ex
...
     }
 
-    save() {
-        if (this.valid) {
-            return {
-                mainJobStartDate: this.property
-            };
-        }
+    restorePropertyFromSequelize(document) {
+        return new moment(document.MainJobStartDateValue);
```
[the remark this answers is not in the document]
wozzer72: It works with the new too. I'll leave this one in unless you have strong reasoning for me to remove it and then run through my tests.

server/models/classes/worker/properties/nationalInsuranceProperty.js
```diff
@@ -1,43 +1,60 @@
...
     }
 
-    toJSON() {
-        return {
-            nationalInsuranceNumber: this.property
-        }
+    isEqual(currentValue, newValue) {
+        // a simple string compare
+        if (currentValue && newValue && currentValue === newValue) return true;
+        else return false;
```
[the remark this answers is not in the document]
wozzer72: The code looks similar, but the value to change depends on the actual concrete type.  It could be argued that the base type `ChangePropertyPrototype` could provide a default isEqual implementation.

But it is not wise in my view, because someone could skip over the actual thought process required to determine the equality of an object. Just look at `server\models\classes\worker\properties\mainJobProperty.js` to see how easy it would have been to have forgotten this.

I've made code changes as per other properties.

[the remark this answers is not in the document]
wozzer72: Code changed accordingly.

server/models/classes/worker/properties/postcodeProperty.js
wozzer72: Code changed accordingly.

wozzer72:
Thanks for your feedback. The code you referenced as "repeated" is similar rather than duplicated. Repetition is expected because all properties are based on a property framework. The logic though similar, is based on specifics of each property, only known by the property.

Although the framework could provide default implementations, with specialised classes only overriding when necessary, I believe that approach carries more risk because the person implementing the concrete class may overlook the more obvious.

PR #218, opened 2019-01-22
187 erroneous messages when validating username front

adamfaryna:
https://trello.com/c/CAMeYU5U/187-erroneous-messages-when-validating-username#

### DIFF
--- a/src/app/core/services/registration.service.ts
+++ b/src/app/core/services/registration.service.ts
@@ -160,7 +160,15 @@ export class RegistrationService {
 
   private handleHttpError(error: HttpErrorResponse): Observable<RegistrationTrackerError> {
     const dataError = new RegistrationTrackerError();
-    dataError.message = error.message;
+
+    if (error.message) {
+      dataError.message =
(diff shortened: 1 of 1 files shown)


src/app/core/services/registration.service.ts
```diff
@@ -160,7 +160,15 @@ export class RegistrationService {
 
   private handleHttpError(error: HttpErrorResponse): Observable<RegistrationTrackerError> {
     const dataError = new RegistrationTrackerError();
-    dataError.message = error.message;
+
+    if (error.message) {
+      dataError.message =
```
wozzer72: It would be better to trap the error explicitly in src/app/core/services/registration.service.ts::getUsernameDuplicate pipe before calling out to handleError.

adamfaryna: These kind of errors can occur in more places than invocation of `/api/registration/username/:username`. The "old way" of handling errors which is used across entire register path need to be secured and these kind of errors to be properly handled.

PR #219, opened 2019-01-22
325-remove-establishmentid-on-logout: done

adamfaryna:
https://trello.com/c/9Ao5YaUk

### DIFF
--- a/src/app/core/header/header.component.ts
+++ b/src/app/core/header/header.component.ts
@@ -2,6 +2,7 @@ import { Component, OnInit } from '@angular/core';
 import { Router } from "@angular/router"
 
 import { AuthService } from "../services/auth-service"
+import { EstablishmentService } from "../services/establishment.service"
 
 
 @Component({
@@ -12,7 +13,8 @@ export class HeaderComponent {
 
   constructor(
     private router: Router,
-    private authService: AuthService
+    private authService: AuthService,
+    private establishmentService: EstablishmentService
     ) { }
 
   isLoggedIn() {
@@ -22,6 +24,7 @@ export class HeaderComponent {
   signOut(event) {
     event.preventDefault()
     this.authService.logout()
+    this.establishmentService.clearSession()
     this.router.navigate(["/sign-out"])
   }
 }
--- a/src/app/core/services/establishment.service.ts
+++ b/src/app/core/services/establishment.service.ts
@@ -46,7 +46,13 @@ export class EstablishmentService {
 
   public set establishmentId(value:number) {
     this._establishmentId = value
-    localStorage.setItem("establishmentId", value.toString())
+
+    if (value) {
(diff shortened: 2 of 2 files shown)


wozzer72:
The code changes are clear. We have however introduced a dependency an additional dependency(Establishment service) into the header component. I understand Angular support RxJS. There must be a better to implement this without such as rigid (brittle) dependency.

Something like the Establishment service listening for notification from the AuthService there is a logout activity.
Or to simplify the implementation, the AuthService and Establishment service having a dependency. But to make the header dependent on Establishment service does not make a lot of sense here.

And finally, the destination for this change is identified as develop. It is not clear from the Trello task of this PR what the priority of the work is. Need clearer direction on which release this change is being approved for.

adamfaryna:
[Home postcode #152](https://trello.com/c/SPakfIMh/44-home-postcode-152)

adamfaryna:
> The code changes are clear. We have however introduced a dependency an additional dependency(Establishment service) into the header component. I understand Angular support RxJS. There must be a better to implement this without such as rigid (brittle) dependency.
> 
> Something like the Establishment service listening for notification from the AuthService there is a logout activity.
> Or to simplify the implementation, the AuthService and Establishment service having a dependency. But to make the header dependent on Establishment service does not make a lot of sense here.
> 
> And finally, the destination for this change is identified as develop. It is not clear from the Trello task of this PR what the priority of the work is. Need clearer direction on which release this change is being approved for.

We are still waiting for as refactoring of frontend code base by @denny13 as anticipated. The solution for mentioned problem was already addressed in https://github.com/NMDSdevopsServiceAdm/SopraSteria-SFC/pull/181, before this is approved we have to copy and paste the same common code between service files (related to setting headers and handling errors) or reuse existing code base.

Hopely entire code base will be cleared soon so these kind of workarounds wouldn't be necessary.

wozzer72:
I'm not seeing how the interceptor, the goal of which is to ensure a reference set of data to initiate RESTful API calls can be used to decouple logout and establishment service, with the aim of deleting the establishment id from local storage. Logout and the removing of the local storage items is not an API call.

adamfaryna:
I misunderstood your previous comment.

Yes, I don't like it either, it's compromise. I don't want to make EstablishmentService, nor AuthService dependent on one another. Subscribing to observers will create this kind of dependency, similar as just injecting one service into another.

The best solution would be to make AuthService emit logout event, than anyone interested would subscribe. I will try to refactor this solution to use JS custom events.

adamfaryna:
[Remove establishmentId on logout](https://trello.com/c/9Ao5YaUk/325-remove-establishmentid-on-logout)

wozzer72:
This has been superseeded by https://github.com/NMDSdevopsServiceAdm/SopraSteria-SFC/pull/257.

PR #227, opened 2019-01-24
63 ethnicity front

adamfaryna:
https://trello.com/c/8XIckTL9

### DIFF
--- a/src/app/core/services/ethnicity.service.ts
+++ b/src/app/core/services/ethnicity.service.ts
@@ -1,0 +1,36 @@
+import { Injectable } from "@angular/core"
+import { HttpClient } from "@angular/common/http"
+import { Observable } from "rxjs"
+import { map, debounceTime, catchError } from "rxjs/operators"
+
+import { HttpErrorHandler } from "./http-error-handler.service"
+import { EstablishmentService } from "./establishment.service"
+import { Ethnicity } from "../model/ethnicity.model"
+
+
+@Injectable({
+  providedIn: "root"
+})
+export class EthnicityService {
+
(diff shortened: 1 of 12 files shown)


adamfaryna:
[Ethnicity? #152](https://trello.com/c/8XIckTL9/63-ethnicity-152)

src/app/features/workers/ethnicity/ethnicity.component.ts
```diff
@@ -1,14 +1,101 @@
...
+  private workerId: string
+  private worker: Worker
+  private subscriptions = []
+
+  ethnicitiesUngrouped() {
+    return this.ethnicities[""]
+  }
+
+  ethnicityGroups() {
+    return Object.keys(this.ethnicities).filter(e => e.length)
```
wozzer72: @adamfaryna You do know the API Usage document and the Trello card both emphasise that the Ethnicity response includes both a straight forward list of ethnicities, along with the separate set of ethnicity by group

src/app/core/services/ethnicity.service.ts
```diff
@@ -0,0 +1,36 @@
...
+})
+export class EthnicityService {
+
+  constructor(
+    private http: HttpClient,
+    private httpErrorHandler: HttpErrorHandler
+  ) {}
+
+  /*
+   * GET /api/ethnicity
```
wozzer72: There are two separate lists within the ethnicity API response; you could easily encapsulate this with two separate methods - one to return list Of and the other to return groupBy, by calling upon the same API but dereferencing the response accordingly.

PR #230, opened 2019-01-25
68 recruited from front

adamfaryna:
https://trello.com/c/a01Sa1vO


adamfaryna:
[Recruited from #152](https://trello.com/c/a01Sa1vO/68-recruited-from-152)

PR #235, opened 2019-01-29
Hotfix v1.3 local authorities

wozzer72:
This includes the correction to properly handle a null primary authority when fetching the list of Local Authorities to share with.

https://trello.com/c/muzhQiT2

Raising as a hotfix to deploy into "test", because this will affect user experience in UAT (cohort 1).

Thanks


wozzer72:
[Sharing data with local authorities fails on fetch if the Establishment's postcode does not match on a primary authority](https://trello.com/c/muzhQiT2/349-sharing-data-with-local-authorities-fails-on-fetch-if-the-establishments-postcode-does-not-match-on-a-primary-authority)

PR #238, opened 2019-01-30
Other job role front

adamfaryna:
https://trello.com/c/XVvDFixW


adamfaryna:
[Other job role #152](https://trello.com/c/XVvDFixW/62-other-job-role-152)

PR #239, opened 2019-01-31
Wortker Property Corrections following updated automated integration test

wozzer72:
1. For handling when a property has changed if it has never been set, and for saving property values only when the property has changed.
2. Correcting the JSON response on Days Sick.
3. Correcting the JSON reponse and incorrect isEqual logic on Weekly Average & Contracted Hours.

### DIFF
--- a/server/models/classes/properties/changePrototype.js
+++ b/server/models/classes/properties/changePrototype.js
@@ -195,10 +195,6 @@ class ChangePropertyPrototype extends PropertyPrototype {
         }
 
         return {
-            // properties: {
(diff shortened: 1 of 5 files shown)


wozzer72:
Regression testing on all Worker properties after having made changes to the core property frameowrk and how it handles "change" assertion against a property and writes that change to the Worker record.
image

wozzer72:
DB Schema on `sfcdevdb` has been rebuilt. Merging the code to initiate deployment to DEV.

PR #247, opened 2019-02-01
Grouped frontend branch for Workers

adamfaryna:
The name doesn't reflect the purpose, I didn't wanted to change it to protect existing references if there are any.

Frontend for Workers:
- adult-social-care-started
- create-staff-record
- ethnicity
- main-job-start-date
- nationality
- year-arrived-uk
- british-citizenship 
- date-of-birth
- gender
- mental-health
- other-job-roles
- country-of-birth
- disability
- home-postcode
- national-insurance-number
- recruited-from


adamfaryna:
> There is nothing I can see here that will break develop.
> 
> There are some changes to non-workers - scope really should be controlled or highlighted:
> `vacancies, starters and leavers - max number of digits change - not within scope of Workers, and owing to deployment into `develop` target branch, won't make UAT until Workers do.`
> 
> I'm accepting the message service as is - not been party to the discussions with Denny or those agreed outcomes.
> 
> _Cautionary Comments:_
> 
> 1. Common shared component changes: not clear to me the impact of those changes (I didn't want to have to click through all the Trello card to identify which if any had summarised the reason for changing these):
> 
> * src/app/shared/number.directive.ts
> * src/app/shared/number-positive-only.directive.ts
> * src/app/shared/number-max.directive.ts
> 

It was necessary, because of date-picker component, which had some issues with these directives.

> 1. Common styles:
> 
> * src/assets/scss/type.scss - has added 10px to the line height to everything - this will need confirmation on target browsers and platforms on deployment!!!

The previous setting, didn't consider splitting heading into two lines so there was no proper interline between lines.

> 
> _Advisory comments:_
> 
> 2. My quick glance on the implementation across all the Worker components (saveHandler) suggests it is only committing to backend API only if the value has changed from 'fetched' value; the expectation is that regardless of whether the value has changed or not, the API must be invoked, to be able to updated the "last saved timestamp and by whom". A necessary on audit tracking.

Ok, I will amend that.

> 3. Every component is doing a fetch of the Worker in "ngOnInit"; I recall a conversation at the start of the Worker epic in sprint planning, where is was suggested that the Worker would be fetched once, and each component update a cached representation of a Worker on successful update, thus reducing the overall load on backend (especially if most users are only ever going to click through the 26 individual screens or so and update only one value). With this original approach, that would have resulted in one fetch and one update on the API. With this current approach, its 27 fetches (one for the initial summary screen) and one update - an unnecessary 26 fetches all of which are the same.

Good point, I will add this.

>    4.Approved Mental Health Worker - the navigation logic is not consistently applying check on main job and/or other jobs being/including Social Worker.

The check if `Social Worker` was selected in other-jobs screen is sufficient requirement to go directly to NIN page, otherwise the normal routing apply. It's consistent which Worker Flow document.

> 4. What happens in the components "saveHandler" and forward navigation if there an error reported from the backend?

It will show "Server error. Please try again later, sorry." message and prevent from navigating further. 

> 5. Other Jobs component; I can't see where it is filtering out their current main job from the list of "jobs".

Good point, I will add this.

> 6. Auto suggest component? I don't recall autoi suggest being a requirement in any of the Trello cards; this is a UI change of scope.

It's one of the requirements for Nationality screen https://trello.com/c/eBXpK76c.

> 7. Date picker component is good encapsulation; would have been good to include momennt's isValid within that component, to encapsulate those awkward months and different number of days thing.

It's already done in the component, by `DateValidator.dateValid()` call.

PR #251, opened 2019-02-04
Country of birth front

adamfaryna:
https://trello.com/c/W7XnNxhH

### DIFF
--- a/src/app/core/services/country.service.ts
+++ b/src/app/core/services/country.service.ts
@@ -1,0 +1,40 @@
+import { Injectable } from "@angular/core"
+import { HttpClient, HttpHeaders } from "@angular/common/http"
+import { Observable } from "rxjs"
+import { catchError, debounceTime, map } from "rxjs/operators"
+
+import { HttpErrorHandler } from "./http-error-handler.service"
+
+@Injectable({
+  providedIn: "root"
+})
+export class CountryService {
+
+  constructor(
+    private http: HttpClient,
+    private httpErrorHandler: HttpErrorHandler
+  ) {}
+
+  // TODO remove when merged https://github.com/NMDSdevopsServiceAdm/SopraSteria-SFC/pull/181
+  private getOptions() {
+    let headers = new HttpHeaders()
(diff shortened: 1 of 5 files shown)


src/app/core/services/country.service.ts
```diff
@@ -0,0 +1,40 @@
...
+    return { headers }
+  }
+
+  /*
+   * GET /api/country
+   */
+  getCountries(): Observable<CountryResponse[]> {
+    return this.http.get<any>("/api/country", this.getOptions())
+      .pipe(
+        debounceTime(500),
```
wozzer72: Why is a debounce here? It's not as if this country request is initiated by username activity.

src/app/features/workers/country-of-birth/country-of-birth.component.html
```diff
@@ -1,3 +1,60 @@
...
+            <input type="radio" formControlName="cobKnown" class="form-control" value="Other" for="cob_known_other">
+            <label for="cob_known_other">Other</label>
+          </div>
+        </div>
+      </div>
+
+      <div *ngIf="form.get('cobKnown').value === 'Other'" class="panel panel-border-narrow col-10 col-lg-6 ml-0 mt-5 mb-5 mt-md-0">
+        <div class="form-field">
+          <label class="form-label col-12 col-md-9">Country
+            <span class="example">Optional</span>
```
wozzer72: "example" class?

src/app/core/services/country.service.ts
adamfaryna: Ok, I will remove it.

src/app/features/workers/country-of-birth/country-of-birth.component.html
adamfaryna: Yes, it's for displaying input examples.

wozzer72: Thanks for the clarification.

src/app/core/services/country.service.ts
wozzer72: Cool.

PR #255, opened 2019-02-05
Update after resetting on develop_v1.3-beta-wa

denny13:
So in completing this task I had to fix a few other things also. Here is what has been done:

* Created registration guard for all registration pages, if you go to them directly you will be kicked back to 'registered-question'.
* Cleared down default values to empty strings and null values.
* Fix validation and error checks on 'select-workplace-address'. Also if the postcode changed on this page it will hide the dropdown until the 'invalid postcode' error has been fixed.
* Have added functionality to each entry of the 'select-workplace-address' page to reset the 'isRegulated' to false.
* Also due to apparent fixes made by other developers I had to update those fixes and then also fix the knock on effects of them not checking and testing the actual impact of their changes across multiple pages.
* Update back button functionality across registration, but will need to be have a full service created later.

I have tested it as best I can but after merging I noticed there were some issues caused by that. I hope I have been able to fix all of those, but if you spot any give me a shout and im happy to re-fix :)

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -1,10 +1,11 @@
 import { NgModule } from '@angular/core';
 import { Routes, RouterModule } from '@angular/router';
 
-import { AuthGuard } from "./core/services/auth-guard.service"
+import { AuthGuard } from './core/services/auth-guard.service';
+import { RegisterGuard } from './core/guards/register/register.guard';
 
 import { LoginComponent } from './features/login/login.component';
-//import { RegisterComponent } from './features/register/register.component';
+// import { RegisterComponent } from './features/register/register.component';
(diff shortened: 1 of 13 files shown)


src/app/app-routing.module.ts
```diff
@@ -31,62 +32,82 @@ import { ServicesCapacityComponent } from './features/services-capacity/services
...
-  },
   {
     path: 'sign-out',
     component: LogoutComponent
   },
   {
-    path: 'select-workplace',
-    component: SelectWorkplaceComponent
+    path: 'registered-question',
+    component: CqcRegisteredQuestionComponent,
```
wozzer72: Does this not need the new register guard?

denny13: It's the page the guard kicks back to so not needed there :)

wozzer72: Understood.

PR #269, opened 2019-02-07
Revised default uninitialised job role for None and Don't know options.

wozzer72:
This follows feedback from Ann on testing https://trello.com/c/d4uwAajW yesterday.

Even when the user had previously selected "none" or "don't know" options, we still need to display a single job role with "Select job title" option shown.

This also fixed a bug I had introduced on starters and leavers default view (the else clause using the wrong control to set).

I've taken the opporutnity to remove unused variable in vacancies and starters (`noVacanciesReasonControl`).


wozzer72:
Approved by Denny earlier today. Now we have a stable `develop` branch, merging these fixes to `develop` for subsequent deployment.

PR #275, opened 2019-02-08
Slack to develop

wozzer72:
Hi Nasir

This is against Trello card (https://trello.com/c/PTIz2Ghs), which during stand up this morning (Friday 7th Feb), Pete agreed the work needed to be prioritised.

He wants to demo this on Monday to Jackie while he's in Leeds.

Hence this PR to `develop`. I shall also raise a separate PR against test, because Pete is looking to get it into UAT as soon as. The low impact of the changes makes it possible to merge straight into test.

But note - you'll need two environment variables to each environment. I've sent you an updated set of release notes. I can walk you through.

Thanks

### DIFF
--- a/server/routes/registration.js
+++ b/server/routes/registration.js
@@ -4,6 +4,7 @@ var concatenateAddress = require('../utils/concatenateAddress').concatenateAddre
 const bodyParser = require('body-parser');
 var bcrypt = require('bcrypt-nodejs');
 
+const slack = require('../utils/slack/slack-logger');
 const { Client } = require('pg');
 
 // TODO - copied from model/index.js; this will be removed when registration is refactored to using sequelize''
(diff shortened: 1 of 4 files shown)


server/routes/registration.js
```diff
@@ -286,6 +287,13 @@ router.route('/')
         throw errOnLogin;
       }
 
+      // post via Slack, but remove sensitive data
+      const slackMsg = req.body[0];
+      delete slackMsg.user.password;
+      delete slackMsg.user.securityQuestion;
+      delete slackMsg.user.securityAnswer;
+      slack.info("Registration", JSON.stringify(slackMsg, null, 2));
```
nasirmids: Just to check; number 2 in slack info represent warning; but it more seems like info which is 3

wozzer72: The 2 here is a parameter of `JSON.stringify` which controls the tab size. It has no relevance to `slack.info` logging level.

nasirmids: got it thanks

PR #281, opened 2019-02-11
183: Scroll to top on navigate

ChazUK:
Added an subscription to the Router events which will return the page position to the top on successful NavigationEnd event


ChazUK:
[Page needs to load at the top](https://trello.com/c/HPxtDnSV/183-page-needs-to-load-at-the-top)

PR #283, opened 2019-02-11
262: Average Contracted Hours Card

ChazUK:
Adding the Average Contracted Hours Card.

@wozzer72 There's API calls on line 34 and 75 of `src/app/features/workers/average-contracted-hours/average-contracted-hours.component.ts`.

### DIFF
--- a/src/app/core/model/worker.model.ts
+++ b/src/app/core/model/worker.model.ts
@@ -54,8 +54,8 @@ export interface Worker {
   };
   zeroHoursContract: string;
   weeklyHoursAverage: {
-    value: number;
-    days?: number;
+    value: string;
+    hours: number;
   };
   weeklyHoursContracted: {
     value: string;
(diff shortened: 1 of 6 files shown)


ChazUK:
[Contracted Weekly hours](https://trello.com/c/LjrgKPFt/262-contracted-weekly-hours)

src/app/core/model/worker.model.ts
```diff
@@ -53,4 +53,12 @@ export interface Worker {
     days?: number;
   };
   zeroHoursContract: string;
+  weeklyHoursAverage: {
```
wozzer72: Follows on from my previous comment to normalise to just `weeklyHours` and reuse across average and contracted. But I would still approve as is.

wozzer72: There are no need to make any changes here; it's just an observation.

PR #285, opened 2019-02-12
263: Salary Card

ChazUK:
Adding Salary Card, asking the question "Are they paid hourly or annually?"

Validation has only been implemented to a basic functionality.

@wozzer72 you can see the API PUT call in the `saveHandler()` method.

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -32,6 +32,7 @@ import { TypeOfEmployerComponent } from './features/type-of-employer/type-of-emp
 import { UserDetailsComponent } from './features/user-details/user-details.component';
 import { VacanciesComponent } from './features/vacancies/vacancies.component';
 import { AdultSocialCareStartedComponent } from './features/workers/adult-social-care-started/adult-social-care-started.component';
+import { BlankCardComponent } from './features/workers/blank-card.component';
 import { BritishCitizenshipComponent } from './features/workers/british-citizenship/british-citizenship.component';
 import { ContractWithZeroHoursComponent } from './features/workers/contract-with-zero-hours/contract-with-zero-hours.component';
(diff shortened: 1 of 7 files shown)


ChazUK:
[Pay Hourly or Annually #152](https://trello.com/c/el8TER0J/263-pay-hourly-or-annually-152)

src/app/app-routing.module.ts
```diff
@@ -32,13 +32,16 @@ import { TypeOfEmployerComponent } from './features/type-of-employer/type-of-emp
...
+import { AverageWeeklyHoursComponent } from './features/workers/average-weekly-hours/average-weekly-hours.component';
+import { BlankCardComponent } from './features/workers/blank-card.component';
 import { BritishCitizenshipComponent } from './features/workers/british-citizenship/british-citizenship.component';
 import { ContractWithZeroHoursComponent } from './features/workers/contract-with-zero-hours/contract-with-zero-hours.component';
 import { CountryOfBirthComponent } from './features/workers/country-of-birth/country-of-birth.component';
 import { CreateStaffRecordComponent } from './features/workers/create-staff-record/create-staff-record.component';
 import { DateOfBirthComponent } from './features/workers/date-of-birth/date-of-birth.component';
 import { DaysOfSicknessComponent } from './features/workers/days-of-sickness/days-of-sickness.component';
 import { DisabilityComponent } from './features/workers/disability/disability.component';
+import { EditWorkerComponent } from './features/workers/edit-worker/edit-worker.component';
```
wozzer72: I can see this import is new, but there is no complemenary route defined below.

ChazUK: There doesn't need to be a route as it's used as the base component for all the child routes whilst editing a worker, This allows us to add a guard to all the child routes which I think you commented on before. I think why you're seeing this is because the imports have been organised so it's moved from the bottom.

PR #290, opened 2019-02-12
Non archived locations

wozzer72:
Hi Nasir

This is a small change to the lambda function that fetches the locations. I've reintroduced the `populateAllLocations` function previously commented out and I've introduced predicate logic in to the main handler function to choose which of `REFRESH` or `BULK_UPLOAD` to perform.

I've added a condition to `getIndividualLocations` function which checks the presence of `deregistrationDate` and if not present (not archived) copies the location data.

The `location` table structure has changed since this lambda function was first written; so I've had to update the sequelize model.

And the `local.js` I've introduced to run the handler locally; no need to deploy the lambda and run in the cloud.

### DIFF
--- a/CQCLocationAPICall/handler.js
+++ b/CQCLocationAPICall/handler.js
@@ -3,45 +3,60 @@ const models = require('./models/index');
 //CQC Endpoint
 const url = 'https://api.cqc.org.uk/public/v1';
 
+const BULK_UPLOAD=0;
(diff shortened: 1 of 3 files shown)


nasirmids:
Hi Warren, we have upload this to aws lambda because we initially thought to run it like schedule job and refresh it but local.js is a good add so save time.

PR #295, opened 2019-02-13
Worker journey optional questions

ChazUK:
There are a few questions that send objects with null values to the API when the question answer has not been selected. This will send a null instead.


src/app/features/workers/average-contracted-hours/average-contracted-hours.component.ts
```diff
@@ -70,7 +70,9 @@ export class AverageContractedHoursComponent implements OnInit, OnDestroy {
 
       if (this.form.valid) {
         const worker = this.worker || ({} as Worker);
-        worker.weeklyHoursContracted = { value: weeklyHoursContractedKnown.value, hours: weeklyHoursContracted.value };
+        worker.weeklyHoursContracted = weeklyHoursContractedKnown.value
```
wozzer72: I didn't notice before, but you have "average contracted" rather than "weekly contracted" hours as the filename for the component.

wozzer72:
None of the APIs are expecting nulls. It's expecting an empty (not NULL) PUT.

wozzer72:
Appreciate that the incumbent code is already passing NULLs, and therefore, this approach to optional property change, is based on consistency. I've raised a Trello card (https://trello.com/c/He9IduC7), to record this as a defect, because the API is not expecting NULLs. Although it's working now, this is not intentional, and can stop working at any time.

src/app/features/workers/average-contracted-hours/average-contracted-hours.component.ts
ChazUK: I'll rename this component

wozzer72: Indeed. Looks good.

ChazUK:
I've added a Details component which works like the GDS equivalent (https://design-system.service.gov.uk/components/details/) This is a shared component which can be reused

PR #296, opened 2019-02-13
169: Staff Record Start Screen

ChazUK:
Adding the Staff Record start screen.
Updated the link on the welcome screen to point to start screen.
Adding styling for bulleted links.

### DIFF
--- a/src/assets/scss/type.scss
+++ b/src/assets/scss/type.scss
@@ -100,10 +100,12 @@ strong {
   @include rem(line-height, 24px);
 }
 
-.fs-19 {
-
+%fs-19 {
   @include rem(font-size, 19px);
   @include rem(line-height, 19px);
+}
+.fs-19 {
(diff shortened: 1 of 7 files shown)


ChazUK:
[Create Staff records start screen](https://trello.com/c/b2AuZZXW/169-create-staff-records-start-screen)

src/assets/scss/type.scss
```diff
@@ -131,3 +133,38 @@ strong {
     }
   }
 }
+
+ul.list {
```
wozzer72: Need @denny13 input to confirm whether the style should be global or localised to the component.

denny13: Yeah these things should be checked before updating globally. What he's done seems unnecessary.. but will work. The whole global styles need refactoring after Adam doing the same thing, without communicating.

PR #307, opened 2019-02-18
Dev proxy

wozzer72:
Hi Nasir

This is the local proxy I have made available in the main repo, using the same code base we had demonstrated previously for the repo split.

https://trello.com/c/bckC3UOM

You'll see I've introduced `proxy.js` - rather than changing `server.j`.

You'll also see I've introduced `npm-run-all`, using it in preference to `concurrently`, because the latter does not combine the logs of multiple processes, whereas the former does.

You'll also see that I've introduced a couple of additional "scripts" to `package.json` for the purpose of running this local proxy.

*BUT, you'll also see I've modified the standard `run start` script line too, removing the watch on `ng` and concurrently. This is because when deploying to DEV, TST, and UAT, you don't need to monitor for any changes in UI files; you have what you are given. Simply build them, then run node.


wozzer72:
[Main Repo - frontend local proxy to API backend](https://trello.com/c/bckC3UOM/332-main-repo-frontend-local-proxy-to-api-backend)

wozzer72:
Merging down to `develop` to correct the already deployed worker list, and also make available the local dev proxy.

PR #308, opened 2019-02-19
Tech Debt: Creating a Shared Module

ChazUK:
I have created a Shared Module and reorganised the folder, so this PR will show a lot of moving files into their relevant folders. This will help us create reusable components, such as the Details Component I created last week.

I've also updated the tsconfig to include import path aliases so that we don't get `import { } from '../../../../shared/'` in a lot of places, instead imports will be displayed as:

```
@core -> src/app/core/
@features -> src/app/features/
@shared -> src/app/shared/
```


ChazUK:
The changes to the `src/app/features/cqc-registered-question/cqc-registered-question-edit/cqc-registered-question-edit.component.ts` are updating the import of the shared CustomValidator, the rest is Prettier formatting.

Prettier should format on save. If you want to look into changing some of the formatting rules, then you can add them to the `.prettierrc.yml` and then we can agree on a point in time to run Prettier on all files across the Angular application.

ChazUK:
For future reference this can be done by:
`prettier --debug-check src/**/*.ts` to check the files first and then `prettier --write src/**/*.ts`

https://prettier.io/docs/en/cli.html

PR #309, opened 2019-02-19
Workers Module

ChazUK:
Creating the Workers Modules so that the app module and routing is overloaded.
Removed Blank Card Component as it's not needed anymore.
Moved NoPaste Directive and MessagesComponent to Shared.


src/app/shared/shared.module.ts
```diff
@@ -4,7 +4,9 @@ import { ReactiveFormsModule } from '@angular/forms';
 import { AutoSuggestComponent } from './components/auto-suggest/auto-suggest.component';
 import { DatePickerComponent } from './components/date-picker/date-picker.component';
 import { DetailsComponent } from './components/details/details.component';
+import { MessagesComponent } from './components/messages/messages.component';
 import { SubmitButtonComponent } from './components/submit-button/submit-button.component';
+import { NoPaste } from './directives/no-paste.directive';
```
wozzer72: nopaste to be removed in subsequent PR.

PR #328, opened 2019-02-21
344: Fixing back button on Create Staff Record Component

ChazUK:


ChazUK:
[Create staff record screen: Back button going to Days of sickness page](https://trello.com/c/HaeJ6Bwa/344-create-staff-record-screen-back-button-going-to-days-of-sickness-page)

PR #329, opened 2019-02-21
343: Updating copy on Staff Record Start Screen

ChazUK:


ChazUK:
[Create staff records start screen: CTA button should be 'Start now >'](https://trello.com/c/izWihryY/343-create-staff-records-start-screen-cta-button-should-be-start-now)

src/app/shared/components/start-button/start-button.component.html
```diff
@@ -0,0 +1,3 @@
+<a [routerLink]="link" role="button" draggable="false" class="govuk-button govuk-button--start">
+  {{ title }}
```
wozzer72: Nice encapsulation. I can see where the `link` attribute is coming from on `create-staff-record-start-screen.component.html` above, but I can't see where the `title` attribute is coming from.

ChazUK: A default is set on the `@Input() title` to set it to Start now

PR #330, opened 2019-02-21
341: Removing Purple navigation bar and updating header navigation

ChazUK:


ChazUK:
[Purple navigation bar should not be on the designs- revert back](https://trello.com/c/HwNeuheA/341-purple-navigation-bar-should-not-be-on-the-designs-revert-back)

PR #331, opened 2019-02-22
Change password built on top of reset-password

wozzer72:
https://trello.com/c/pByUKSW3 - change password is just one of the User's properties to change. However, unlike name, email et al, changing a password has unique handling.

Change Password is like reset password, except, it can only be called when logged in (so needs to verify against login JWT) and must give the current password which must be authenticated first.

This PR includes the biggest risk of all, and that is:
1. Moving Login security question/answer to User
2. Renaming existing User aliases in preparation for Extended Change Property - this is the biggest risk because any reference to old column will cause a failure.

Whilst regression testing the changes to User sequelizie model Column aliases my automated tests were failing. To get tests working again I noticed (and fixed):
1. Transactions were not properly rolling back; turns out, have to pass the `transaction` object into those sequelize calls (create, update and destroy) to be tied back to the transaction.
2. The unique constraints on Establishment were topsy turvy.
3. The relationship (association) from Login to User was joining on the wrong Login field - only noticed this after a transaction failure on registration and my login/user record ids got out of sync!
4. `server/routes/la.js` - `api/localAuthority/:postcode` was returning different results to `api/establishment/:id/la` for the primary authority, as a result of updating the latter last week whilst in the to resolve first on full postcode, but not the former which I use in my tests.

Thanks


NMDSdevopsServiceAdm:
Gone through the code and changes with warren; seems all perfect!

wozzer72:
Ready to merge back down to `reset-password`, for subsequent merge to `develop` for deployment in dev.

PR #338, opened 2019-02-26
Merge from Develop to Test

NMDSdevopsServiceAdm:
Mainly worker front end along with backend for user password, reset and change.


wozzer72:
This is an accurate reflection of what is in Ann's test backlog.

PR #347, opened 2019-03-02
151: Staff Skeleton Records

ChazUK:
Users can now create skeleton records.
Added a Status Component similar to https://home-office-digital-patterns.herokuapp.com/components/status
Added Error Summary Component https://design-system.service.gov.uk/components/error-summary/
Added blue button style

### DIFF
--- a/src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
+++ b/src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
@@ -23,15 +23,7 @@ export class StaffRecordsTabComponent implements OnInit, OnDestroy {
       })
     );
 
-    if (localStorage.getItem('staffcreated') || localStorage.getItem('stafffailed')) {
-      this.createStaffResponse = {
-        success: parseInt(localStorage.getItem('staffcreated'), 10),
-        failed: parseInt(localStorage.getItem('stafffailed'), 10),
-      };
-
-      localStorage.removeItem('staffcreated');
-      localStorage.removeItem('stafffailed');
-    }
+    this.createStaffResponse = this.workerService.getCreateStaffResponse();
   }
 
   ngOnDestroy() {
--- a/src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
+++ b/src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
@@ -1,0 +1,40 @@
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { Worker } from '@core/model/worker.model';
+import { WorkerService } from '@core/services/worker.service';
+import { Subscription } from 'rxjs';
+
+@Component({
+  selector: 'app-staff-records-tab',
+  templateUrl: './staff-records-tab.component.html',
+})
+export class StaffRecordsTabComponent implements OnInit, OnDestroy {
+  public createStaffResponse = null;
+  public workers: Worker[];
+  public incomplete = 0;
+  private subscriptions: Subscription = new Subscription();
+
+  constructor(private workerService: WorkerService) {}
+
+  ngOnInit() {
+    this.subscriptions.add(
+      this.workerService.getAllWorkers().subscribe(data => {
+        this.workers = data;
+        this.incomplete = this.workers.filter(worker => worker.created === worker.updated).length;
+      })
+    );
+
+    if (localStorage.getItem('staffcreated') || localStorage.getItem('stafffailed')) {
+      this.createStaffResponse = {
+        success: parseInt(localStorage.getItem('staffcreated'), 10),
+        failed: parseInt(localStorage.getItem('stafffailed'), 10),
+      };
+
+      localStorage.removeItem('staffcreated');
+      localStorage.removeItem('stafffailed');
+    }
+  }
+
+  ngOnDestroy() {
+    this.subscriptions.unsubscribe();
+  }
+}
--- a/src/app/features/workers/create-staff-record/create-staff-record.component.ts
+++ b/src/app/features/workers/create-staff-record/create-staff-record.component.ts
@@ -1,56 +1,82 @@
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { FormBuilder, FormGroup, Validators } from '@angular/forms';
-import { ActivatedRoute, Router } from '@angular/router';
+import { FormArray, FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { Router } from '@angular/router';
 import { Contracts } from '@core/constants/contracts.enum';
 import { Job } from '@core/model/job.model';
 import { Worker } from '@core/model/worker.model';
+import { EstablishmentService } from '@core/services/establishment.service';
 import { JobService } from '@core/services/job.service';
 import { MessageService } from '@core/services/message.service';
 import { WorkerEditResponse, WorkerService } from '@core/services/worker.service';
+import { reject } from 'q';
 import { Subscription } from 'rxjs';
 
 @Component({
   selector: 'app-create-staff-record',
   templateUrl: './create-staff-record.component.html',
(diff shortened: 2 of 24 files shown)


src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  ngOnInit() {
+    this.subscriptions.add(
+      this.workerService.getAllWorkers().subscribe(data => {
+        this.workers = data;
+        this.incomplete = this.workers.filter(worker => worker.created === worker.updated).length;
+      })
+    );
+
+    if (localStorage.getItem('staffcreated') || localStorage.getItem('stafffailed')) {
```
wozzer72: We're back to using local storage then? Why does number of staff created/failed need to be persisted?

src/app/features/workers/create-staff-record/create-staff-record.component.ts
```diff
@@ -1,93 +1,203 @@
...
   ) {
-    this.saveHandler = this.saveHandler.bind(this);
+    this.addStaffRecord = this.addStaffRecord.bind(this);
+    this.totalStaffValidator = this.totalStaffValidator.bind(this);
+  }
+
+  get displayAddMore() {
+    const staffRecordsControl = <FormArray>this.form.controls.staffRecords;
+
+    return staffRecordsControl.controls.some(control => {
```
wozzer72: NIce. I like this. You can resolve on this comment.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
```diff
@@ -1,93 +1,203 @@
...
+    this.subscriptions.add(
+      this.workerService.getAllWorkers().subscribe(workers => {
+        this.totalWorkers = workers.length;
+        this.form.updateValueAndValidity();
+      })
+    );
+    this.subscriptions.add(
+      this.jobService.getJobs().subscribe(jobs => {
+        this.jobsAvailable = jobs;
+        this.jobsAvailable.push({ jobId: 500, id: 500, title: 'FAKE JOB' });
```
wozzer72: I can see the logic of why - and it's better than a NULL value; and although there are low double digits jobs here, we cannot guarantee there won't be more than 500 in the future. "0" is a safer bet as job Id starts from 1.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
```diff
@@ -1,93 +1,203 @@
...
+        localStorage.setItem('staffcreated', `${complete.length}`);
+        localStorage.setItem('stafffailed', `${res.length - complete.length}`);
+      });
 
-      if (this.isSocialWorkerSelected()) {
-        this.router.navigate(['/worker', res.uid, 'mental-health']);
-      } else {
-        this.router.navigate(['/worker', res.uid, 'main-job-start-date']);
-      }
+      await this.saveTotalStaff().catch(error => false);
```
wozzer72: Not being an angular person, it's not clear to me how any errors (exceptions) thrown by waiting upon the saveHandlers promises to finish and captured against `this.saveTotalStaff`.

But this I can assume this is why you asked me how to return an error from API, so I just need confirmation that this has been tested.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
```diff
@@ -1,93 +1,203 @@
...
   async submitHandler() {
     try {
-      const res = await this.saveHandler();
+      const promises = this.form.controls.staffRecords['controls'].map(control =>
+        this.saveHandler(control).catch(error => false)
+      );
+
+      await Promise.all(promises).then(function(res) {
+        const complete = res.filter(value => !!value);
+        localStorage.setItem('staffcreated', `${complete.length}`);
```
wozzer72: Still not clear why it is necessary to use local storage for the staff created/failed, when all is running within the same component.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
```diff
@@ -1,93 +1,203 @@
...
+      Object.keys(staffRecord.controls).forEach(key => {
+        staffRecord.get(key).markAsTouched();
+      });
+    }
   }
 
   async submitHandler() {
     try {
-      const res = await this.saveHandler();
+      const promises = this.form.controls.staffRecords['controls'].map(control =>
```
wozzer72: This suggests a deviation away from from infer requirements (previous discussions) which says that a staff record is created on each "add" as it happens, rather than all grouped together.

I can see why you have grouped them to allow for the add/delete given there is no delete API yet.

I take it this revised approach has been agreed with Maria & Pete and accordingly agreed in the trello card?

src/app/features/workers/create-staff-record/create-staff-record.component.ts
ChazUK: This should have been removed as it was for my testing purposes only to display how only the request can be made adding multiple and 1 can fail but the others will still be processed.

src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
ChazUK: I would love to move this to a state management solution like NGRX or NGXS (Redux) but I can't see that being implemented any time soon.
The data needs to be stored when navigating from creating the skeleton records to the Dashboard, and is then immediately removed.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
ChazUK: This has now been removed.

ChazUK:
[Create staff records (Skeleton Record) - #151](https://trello.com/c/Zhiyd7lc/105-create-staff-records-skeleton-record-151)

src/app/features/workers/create-staff-record/create-staff-record.component.ts
ChazUK: I do have them grouped because there is no DELETE API endpoint for Workers.

I'm happy to change Total Staff how you see fit, either being updated every time it is changed, or moving the call to after the workers have been added via the API, or any other suggestions you might have.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
ChazUK: This is the same as the above comment, I am happy to remove this from localstorage. I was originally going to send it as a URL parameter, but that means it could be reloaded and the message displayed which is not ideal.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
ChazUK: I have made sure that even if one of the promises fail, the others are still attempted.

The ticket states that if the declared Total Staff is less than or equal to the current amount of workers, and then another worker is added, that Total Staff will be incremented. This would mean that it is both declarative and inferred. Happy to change this with your suggestions.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
wozzer72: It's not "how I see it". It's based on conversations that have previously been had. I'll stop arguing the point, and simply accept whatever you give me.

src/app/features/workers/create-staff-record/create-staff-record.component.ts
wozzer72: This is as agreed with Pete and Maria in video conference this morning. A backlog ticket has been raised to cover any changes to approach if necessary.

src/app/features/dashboard/staff-records-tab/staff-records-tab.component.ts
wozzer72: Local storage has been removed

PR #348, opened 2019-03-04
Worker "no change" assertions

wozzer72:
Hi Nasir

Ann raised observation against the Worker change history, where some properties were showing as changed, when their data had not.

I've taken the opportunity to improve upon my Worker integration to assert on "no change", using a method I prove while delivering upon the User Change Details you've just reviewed.

It highlighted there five properties where change was being incorrectly cited. I've fixed the properties and once again I have a full successful regression run against all Worker properties.

Complementary PR for the automated integration tests: https://github.com/wozzer72/restful-api-qa/pull/9


wozzer72:
[Worker Change History - change record created for NI when there is no change](https://trello.com/c/m512XJLN/351-worker-change-history-change-record-created-for-ni-when-there-is-no-change)

PR #363, opened 2019-03-07
214: Nationality is not required if user selects Other

ChazUK:


ChazUK:
[Nationality: Drop list shouldn't include British as this is a separate radio button. Optional (blank Nationality) should also be allowed.](https://trello.com/c/CqgkWzya/214-nationality-drop-list-shouldnt-include-british-as-this-is-a-separate-radio-button-optional-blank-nationality-should-also-be-allo)

ChazUK:
@wozzer72 The required Validation rules has been removed. The only error that would be triggered is if the Nationality name does not match one returned by the API.

Lines 120, 121 removed

PR #364, opened 2019-03-07
215: Country of Birth is not required if user selects Other

ChazUK:


ChazUK:
[Country of birth: Actual country is optional](https://trello.com/c/UqLUIYhD/215-country-of-birth-actual-country-is-optional)

PR #371, opened 2019-03-07
348: Fixing display of salary when answer is set to Don't Know

ChazUK:


ChazUK:
[Staff record summary: Navigation back from changed screens](https://trello.com/c/TDMjbhwF/226-staff-record-summary-navigation-back-from-changed-screens)

ChazUK:
[Staff record summary: Pay rates displayed incorrectly](https://trello.com/c/iW1zffQk/348-staff-record-summary-pay-rates-displayed-incorrectly)

src/app/features/workers/worker-summary/worker-summary.component.html
```diff
@@ -485,7 +485,7 @@ <h3>Employment</h3>
           Salary
         </dt>
         <dd class="govuk-summary-list__value">
-          {{ worker.annualHourlyPay?.value ? salary : '-' }}
+          {{ worker.annualHourlyPay?.rate ? salary : worker.annualHourlyPay?.value || '-' }}
```
wozzer72: This hasn't been tested, because the "rate" field has been changed on the left of the predicate, but not on the right of it.

ChazUK: The logic behind this is:

If `rate` is set, then display the `salary` which is a get function to return the formatted salary with a currency symbol.
Otherwise display just the `value` (Hourly, Annually, Don't Know)
If `value` is null then display `-`

wozzer72: Doh. Good point; salary here is the true case.

PR #372, opened 2019-03-08
Feature/b47 reset password ui

denny13:
Forgotten Password and Password Reset components created.

The error handling has been removed from this task, as we do not have all the information available yet.

### DIFF
--- a/src/app/core/services/password-reset.service.ts
+++ b/src/app/core/services/password-reset.service.ts
@@ -1,0 +1,75 @@
+import { HttpClient, HttpHeaders } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, empty, Observable } from 'rxjs';
+import { catchError, debounceTime, map, tap } from 'rxjs/operators';
+import { EstablishmentService } from './establishment.service';
+import { HttpErrorHandler } from './http-error-handler.service';
+
+export interface RequestPasswordResetResponse {
+  usernameOrEmail: string;
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+export class PasswordResetService {
+
+  private _resetPasswordUUID$: BehaviorSubject<string> = new BehaviorSubject<string>(null);
+  public resetPasswordUUID$: Observable<string> = this._resetPasswordUUID$.asObservable();
+
+  constructor(
+    private http: HttpClient,
+    private httpErrorHandler: HttpErrorHandler,
+  ) { }
+
+  requestPasswordReset(usernameOrEmail: string) {
+    return this.http
+      .post<RequestPasswordResetResponse>(
+        '/api/registration/requestPasswordReset',
+        usernameOrEmail,
+        EstablishmentService.getOptions(),
+      )
+      .pipe(
+        debounceTime(500),
+        catchError(this.httpErrorHandler.handleHttpError),
+      );
+  }
+
+  validatePasswordReset(data) {
+    const uuidData = { uuid: data };
+    const options = { 'Content-type': 'application/json' };
+
+    return this.http
+      .post(
+        '/api/registration/validateResetPassword',
+        uuidData,
+        { headers: options, observe: 'response' }
+      )
+      .pipe(
+        debounceTime(500),
+        catchError(this.httpErrorHandler.handleHttpError),
+      );
+  }
+
+  resetPassword(data, token) {
+
+    const newPassword = { password: data };
+    const requestHeaders = new HttpHeaders({ 'Content-type': 'application/json', 'Authorization': token });
+
+    return this.http
+      .post<any>(
+        '/api/user/resetPassword',
+        newPassword,
+        { headers: requestHeaders, responseType: 'text' }
+      )
+      .pipe(
+        debounceTime(500),
+        catchError(this.httpErrorHandler.handleHttpError),
+      );
+  }
+
+  updateState(data) {
+    this._resetPasswordUUID$.next(data);
+  }
+
+}
--- a/src/app/core/services/password-reset.service.ts
+++ b/src/app/core/services/password-reset.service.ts
@@ -2,7 +2,6 @@ import { HttpClient, HttpHeaders } from '@angular/common/http';
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, empty, Observable } from 'rxjs';
 import { catchError, debounceTime, map, tap } from 'rxjs/operators';
-import { EstablishmentService } from './establishment.service';
 import { HttpErrorHandler } from './http-error-handler.service';
 
 export interface RequestPasswordResetResponse {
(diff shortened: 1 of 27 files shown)


src/app/core/services/password-reset.service.ts
```diff
@@ -0,0 +1,75 @@
...
+    private http: HttpClient,
+    private httpErrorHandler: HttpErrorHandler,
+  ) { }
+
+  requestPasswordReset(usernameOrEmail: string) {
+    return this.http
+      .post<RequestPasswordResetResponse>(
+        '/api/registration/requestPasswordReset',
+        usernameOrEmail,
+        EstablishmentService.getOptions(),
```
ChazUK: Is it possible to remove this call from this service and use it's own getOptions method?

src/app/core/services/password-reset.service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  validatePasswordReset(data) {
+    const uuidData = { uuid: data };
+    const requestHeaders = new HttpHeaders({ 'Content-type': 'application/json' });
+
+    return this.http
+      .post(
+        '/api/registration/validateResetPassword',
+        uuidData,
+        { headers: requestHeaders, observe: 'response' }
```
ChazUK: For instance here you're using different headers

src/app/features/forgot-your-password/confirmation/confirmation.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, OnInit, Input } from '@angular/core';
+
+import { Router } from '@angular/router';
+
+@Component({
+  selector: 'app-fp-confirmation',
+  templateUrl: './confirmation.component.html',
+  styleUrls: ['./confirmation.component.scss']
+})
+export class ForgotYourPasswordConfirmationComponent implements OnInit {
```
ChazUK: Remove empty ngOnInit
Router doesn't need to be injected
Remove commented goToResetPassword function

src/app/features/forgot-your-password/forgot-your-password.component.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { PasswordResetService } from '@core/services/password-reset.service';
+import { Subscription } from 'rxjs';
+
+@Component({
+  selector: 'app-forgot-your-password',
+  templateUrl: './forgot-your-password.component.html',
+  styleUrls: ['./forgot-your-password.component.scss']
+})
+export class ForgotYourPasswordComponent implements OnInit {
```
ChazUK: Need to add an OnDestroy to unsubscribe from the Subscriptions

src/app/features/reset-password/confirmation/confirmation.component.ts
```diff
@@ -0,0 +1,15 @@
...
+@Component({
+  selector: 'app-rp-confirmation',
+  templateUrl: './confirmation.component.html',
+  styleUrls: ['./confirmation.component.scss']
+})
+export class ResetPasswordConfirmationComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
```
ChazUK: Remove empty ngOnInit

src/app/features/reset-password/edit/edit.component.html
```diff
@@ -0,0 +1,57 @@
...
+         [ngClass]="{'is-invalid': displayError && getPasswordInput.errors }">
+      <label class="form-label" for="createPassword">Password</label>
+      <p class="field-comment-text">This should be at least 8 characters long, contain at least one uppercase and one lowercase letter and a number</p>
+
+      <span class="invalid-feedback">
+        <span *ngIf="displayError && getPasswordInput.hasError('required')" class="main-body-text mb-2">
+          Please enter your Password.
+        </span>
+        <span *ngIf="displayError && getPasswordInput.hasError('pattern')" class="main-body-text mb-2">
+          Invalid Password.
```
ChazUK: I'm know this is the same validation message when creating a user, but could we get someone to update the copy to be more helpful and inline with GDS?

src/app/features/reset-password/edit/edit.component.ts
```diff
@@ -0,0 +1,115 @@
...
+  styleUrls: ['./edit.component.scss']
+})
+export class ResetPasswordEditComponent implements OnInit {
+  public resetPasswordForm: FormGroup;
+  @Input() validatePasswordResetResponse;
+  @Input() headerToken: string;
+  public name: string;
+  public displayError: boolean;
+  passwordMessage: string;
+  confirmPasswordMessage: string;
```
ChazUK: Are you using these `passwordMessage` and `confirmPasswordMessage` variables anywhere in the template?

If not then can these and their set methods be removed?

src/app/features/reset-password/reset-password.component.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { PasswordResetService } from '@core/services/password-reset.service';
+import { Subscription } from 'rxjs';
+
+@Component({
+  selector: 'app-reset-password',
+  templateUrl: './reset-password.component.html',
+  styleUrls: ['./reset-password.component.scss']
+})
+export class ResetPasswordComponent implements OnInit {
```
ChazUK: Same here to add the OnDestroy lifecycle hook to unsubscribe from the Subscriptions

ChazUK:
Can you remove all the empty SCSS files from each component as well?

src/app/core/services/password-reset.service.ts
denny13: Can do but why duplicate it unnecessarily?

src/app/features/reset-password/edit/edit.component.html
denny13: That will hopefully come with Charlie's updates

src/app/core/services/password-reset.service.ts
denny13: The difference was due to requiring the observe property.

src/app/core/services/password-reset.service.ts
ChazUK: My main argument is that creating a Headers object isn't unique to the Establishment Service, so why should it belong in there and not it's own helper function. Also I believe that most Services should be able to run independent of other Services.

What happens when in the future someone changes the Establishment Service to contain an additional header that could conflict with every service that is injecting it?

src/app/core/services/password-reset.service.ts
ChazUK: You could use the spread operator here to extend the object, instead of duplicating the code for the request headers

src/app/core/services/password-reset.service.ts
denny13: I agree it shouldn't be unique to Establishment service.

src/app/features/forgot-your-password/confirmation/confirmation.component.ts
denny13: Removed

src/app/features/forgot-your-password/forgot-your-password.component.ts
denny13: Done

src/app/features/reset-password/confirmation/confirmation.component.ts
denny13: Removed

src/app/features/reset-password/edit/edit.component.ts
denny13: No good call on these... They were initially added, but not removed.

src/app/features/reset-password/reset-password.component.ts
denny13: Done

PR #376, opened 2019-03-09
Gov.uk notify - reset password

wozzer72:
Hey Nasir

After having basic configuration on the SfC gov.uk notify account and having demonstrated being able to send emails using it, I have now formally introduced it into the backend API.

It has been agreed that emails will only be sent from test (staging) and production (uat). The configuration of what key, what reply to address, what template et al to use is environment specific.

Thus, I have taken this opportunity to introduce a proper configuration capability (convict) - replacing the old "config" implementation. I have taken the opportunity to apply it (convict) across DB (sequelize) including removing the password!!!!, Slack and JWT, by removing all the embedded occurrences of "process.env". Hence the impact is higehr than otherwise expected.

And yes, I have attached release notes with all the necessary environment variable configuration.

I have also update some of the libraries that were raising audit vulenerabilities (get this in now before the next pen test).

Anyhow. Let's review this together on Monday and deploy into dev on Monday (and hopefully in to test too, tso Ann can start to test password reset functionality with sending of emails).


nasirmids:
Looks great! i have defined the env variables from the release note

wozzer72:
`NODE_ENV` and `DB_PASS` env variables configured in dev (and staging).

wozzer72:
[Admin Account - Email confirmation of actions #28](https://trello.com/c/ONiKc7Ck/121-admin-account-email-confirmation-of-actions-28)

PR #377, opened 2019-03-11
Update the reset password to get reset param from url on load of pass…

denny13:
Updating reset page to get 'uuid' from reset param in url, so that it will work from email links.

### DIFF
--- a/src/app/features/reset-password/reset-password.component.ts
+++ b/src/app/features/reset-password/reset-password.component.ts
@@ -1,5 +1,7 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
 
+import { ActivatedRoute } from '@angular/router';
+
 import { PasswordResetService } from '@core/services/password-reset.service';
 import { Subscription } from 'rxjs';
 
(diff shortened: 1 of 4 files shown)


src/app/features/reset-password/reset-password.component.ts
```diff
@@ -17,17 +20,22 @@ export class ResetPasswordComponent implements OnInit, OnDestroy {
...
+    private route: ActivatedRoute,
   ) { }
 
   ngOnInit() {
     this.submitted = false;
 
-    this._passwordResetService.resetPasswordUUID$.subscribe(resetLinkResponse => (
-      this.resetLinkResponse = resetLinkResponse
-    ));
+    this.route.queryParams.subscribe(params => {
```
ChazUK: You could use the snapshot here rather than subscribe, as the URL won't be changing

src/app/features/reset-password/reset-password.component.ts
```diff
@@ -37,7 +45,7 @@ export class ResetPasswordComponent implements OnInit, OnDestroy {
 
   validatePasswordReset(data) {
     this.subscriptions.add(
-      this._passwordResetService.validatePasswordReset(data.uuid).subscribe(res => {
+      this._passwordResetService.validatePasswordReset(data).subscribe(res => {
```
ChazUK: Angular Style Guide suggests to not use an underscore for private properties, can just use passwordResetService

PR #378, opened 2019-03-11
Removing the explicit declaration of NODE_ENV from the 'npm run start' command

wozzer72:


wozzer72:
[Admin Account - Email confirmation of actions #28](https://trello.com/c/ONiKc7Ck/121-admin-account-email-confirmation-of-actions-28)

PR #385, opened 2019-03-14
Add/Edit User

wozzer72:
Hi Nasir

https://trello.com/c/HzYGVltp

With so many changes here, you could ask "What isn't included in this change?".

Essentially, it is to allow adding and edit of new users (explicitly no delete). A user is added in two stages:

* First registered by a logged in user (having Edit role); sends an email with a link
* Second, on receipt of email, the target user completes registration including the more sensitive aspects of their username (can be duplicate) and their security question/answer

There are changes to the GDS (gov.uk) notify to allow sending an email on registration; consequently, there are changes to the config and test/production for the new email template ID.

There is an AddUserTracking database table (similar to the password reset tracking) with the sequelize model.

Of course there is a BIG impact to the User extended change property class: `server/models/classes/user.js`; too many to list individual, but significantly, the introduction of a new property, that being "UserRole" (either Edit or Read).

When getting a list of all users for a given Establishment, they want to see the time the user last logged in; so updated the Login table and sequelize model to store that fact, along with the "login" endpoint. And login now also generates a JWT token with the role as a claim (removing the unused isAdmin claim) and returning in the response for the frontend the role too.

Although the scope does not include deleting of users, I have introduced an "archived" column to `User` table, because my approach to registering/adding users is to first create a user record is only part of the information (but this specifically includes their role); this is done so without auditing. Then, to complete registration, a new User record is created, using the role as initially registered. It means having to archive the original partly added user record now.

### DIFF
--- a/server/config/config.js
+++ b/server/config/config.js
@@ -140,6 +140,11 @@ const config = convict({
           doc: 'The password reset JWT audience',
           format: String,
           default: 'ADS-WDS-password-reset'
+        },
+        addUser: {
+          doc: 'The add user JWT audience',
+          format: String,
+          default: 'ADS-WDS-add-user'
         }
       }
   },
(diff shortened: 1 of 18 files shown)


wozzer72:
[Admin Account - US73 Ability to create other users (edit & read only) #6](https://trello.com/c/HzYGVltp/120-admin-account-us73-ability-to-create-other-users-edit-read-only-6)

nasirmids:
Hi Warren, understand of it as you explain nicely; one question though can we not use some kind of user status instead of archiving?

PR #386, opened 2019-03-14
Special Characters in Password

wozzer72:
https://trello.com/c/CSGha5w5

manual tests using various complex passwords on registration, password reset and change password each followed by successful login.


wozzer72:
[Passwords with special characters no longer work after password reset or change password](https://trello.com/c/CSGha5w5/383-passwords-with-special-characters-no-longer-work-after-password-reset-or-change-password)

PR #395, opened 2019-03-20
Return to Worker Summary

ChazUK:
This work is to navigate the user back to the Workers Summary screen when they have navigated to a Question from Workers Summary.

The Submit buttons have been changed, along with the back link.


ChazUK:
[Staff record summary: Navigation back from changed screens](https://trello.com/c/TDMjbhwF/5-staff-record-summary-navigation-back-from-changed-screens)

PR #399, opened 2019-03-20
Delete Worker

ChazUK:
We can now create Dialog / Popups / Modals using the DialogService. Data can be passed to the Dialog and back to the parent component to be handled. This is handled by the Angular Overlay
 CDK.

`lastDeleted` in the Worker Service stores the name of the user that has just been deleted.


ChazUK:
[Delete/ Reason for leaving and destination question](https://trello.com/c/6bctXaoy/97-delete-reason-for-leaving-and-destination-question)

PR #401, opened 2019-03-21
Recruited From Validation error

ChazUK:
Using form.controls in custom validator over form.value as it is not updated on validation request.

Fix was provided by @wozzer72, but I also changed the validation error to `required` over `recruitedFromIdValid` and added an ID to the select element for future use.


ChazUK:
[Recruited from: first selection is always rejected](https://trello.com/c/qBS26v14/96-recruited-from-first-selection-is-always-rejected)

PR #403, opened 2019-03-22
Save / Delete Skeleton Records

ChazUK:
This PR includes changes to how the Skeleton Records work. When a record is updated and then the Save button is pressed, the record is saved in the DB. The Delete button now either removes the row (if the record has not been saved yet) or sends a request to the DB to delete the record, and then removes the row.

Total Staff is only ever changed if the amount becomes greater than what they entered. There was no specific logic around this described in the AC's.

```
staffRecordsRequiredValidator(control: FormArray) {
    if (!control.controls.some(c => !isNull(c.get('uid').value))) {
      return { required: true };
    }

    return null;
  }
```

This is a weird bit of logic, the double negative is there because sometimes there might not be any items in the FormArray, so it's checking for that in the same statement.

### DIFF
--- a/server/models/classes/worker.js
+++ b/server/models/classes/worker.js
@@ -447,6 +447,7 @@ class Worker {
     // Can throw "WorkerDeleteException"
     async archive(deletedBy) {
         try {
+            
             const updatedTimestamp = new Date();
 
             // need to update the existing Worker record and add an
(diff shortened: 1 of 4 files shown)


ChazUK:
[Skeleton records- Ability to save and delete skeleton records](https://trello.com/c/bGmhgGNM/42-skeleton-records-ability-to-save-and-delete-skeleton-records)

PR #406, opened 2019-03-22
Ensuring Total Staff is not null and only updating if value has changed

ChazUK:


wozzer72:
Added `server/routes/establishments/staff.js` to accept "0" when updating staff count (0 is an expected value - defect in logic which has been resolved with the Establishment full audit coming soon).

ChazUK:
[Incorrect 'total staff' can be saved to the database](https://trello.com/c/gJX5NYwA/90-incorrect-total-staff-can-be-saved-to-the-database)

PR #409, opened 2019-03-22
Delete Reason is no longer mandatory

ChazUK:


ChazUK:
[Change Request - Reason for leaving](https://trello.com/c/ZrjTCAzi/115-change-request-reason-for-leaving)

PR #411, opened 2019-03-23
Establishment full audit cont 2

wozzer72:
This PR completes my backend changes for https://trello.com/c/jaqBQdQg, and builds upon the previous two PRs. But this PR is to `develop`, and therefore includes the changes already reviewed.

This PR  extends the previous two by including refactoring all the Establishment data into individual properties. I've reused much of the JSON formatters.

The biggest change is on Vacancies, Starters and Leavers We have a single endpoint "jobs", but each of Vacancies, Starters and Leavers are a property. I've introduced postgres views having discussed with Shakir which has introduced new sequelize definitions having table name suffixed with `VW`.

Because Establishment has individual POST/GET endpoints with no PUT, unlike Workers/Users single GET/POST/PUT, it has been necessary to update the change property framework to be able to filter properties in the toJSON representation: `server/models/classes/properties/manager.js`. This so, for example, the staff endpoint only returns the staff property.

The git diff view is trying to be clever on all the establishment express route endpoints: `server/routes/establishments/*`. But it's cleverness to show only differences is making the changes hard to read. It's best just to view the raw data; I've used the same GET/POST replacement for each endpoint with a couple of subtle differences:

1. The property filter (typically it's a single property, but on `.../jobs` we have the three Vacancies, Starters and Leavers properties.
2. Post processing of the JSON for `.../services`, `.../capacity` and `.../jobs`.

`server/routes/establishments/worker.js` is included becuase I needed to fix my local environment as a result of the dodgy deployment last Wednesday (HTTP 502 errors on Worker POST/PUT - I needed to fix this on my branch to get the regression tests to work).


wozzer72:
[Audit trail on Establishment- Full audit of who changes what when - Time and Date #42](https://trello.com/c/jaqBQdQg/98-audit-trail-on-establishment-full-audit-of-who-changes-what-when-time-and-date-42)

wozzer72:
Deployment embargo has been lifted on dev. Merging to trigger deployment to dev.

PR #413, opened 2019-03-25
Removing Validation around Total Staff

ChazUK:
Updating Skeleton Records to be less complex when validating against total staff number

* Remove Total Staff validation
* Fix success message on Staff Records tab
* Add prompt to Staff Records tab when Total Staff and Staff Records amount doesn't match
* Change Staff records displayed index to reflect total staff records
* Hide "Provide more information" link when entered Total Staff does not equal Staff Records.
* Added scrolling to the open record

### DIFF
--- a/src/app/core/services/worker.service.ts
+++ b/src/app/core/services/worker.service.ts
@@ -164,11 +164,8 @@ export class WorkerService {
       );
   }
 
-  setCreateStaffResponse(success: number, failed: number) {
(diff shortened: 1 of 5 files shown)


ChazUK:
[Change Request - Skeleton Records Remove Total Staff Validation](https://trello.com/c/cPGmFJ0w/117-change-request-skeleton-records-remove-total-staff-validation)

PR #421, opened 2019-03-27
Redirect with params and fragments

ChazUK:
Redirect after Login now uses an object to pass fragments and query parameters to `router.navigate()`


ChazUK:
[Redirect after logout not working with URL anchors](https://trello.com/c/pfjQHo0o/130-redirect-after-logout-not-working-with-url-anchors)

PR #424, opened 2019-03-27
Establishment additional properties

wozzer72:
Two trello cards in one branch/PR:
* https://trello.com/c/6y2OTQmB
* https://trello.com/c/6j4x0gui

This is because both introduce one or more change properties to Establishment, so it is more efficient to do both changes in one go.

It's pretty much everything you've seen before.

The "Service Users" can be zero or more reflexion records; hence the new `EstablishmentServiceUsers` sequelize model (for the table) and new association from `Establishment` sequelize model.

A little wierdness on the name property, in that because it (the name) is returned on every establishment endpoint (GET and POST), because it's now a managed property, every endpoint needs both it's target endpint and the "Name" endpoint.

Any question, just ask of course.

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -85,7 +85,7 @@ class Establishment {
         return this._username;
     }
     get name() {
-        return this._name;
+        return this._properties.get('Name') ? this._properties.get('Name').property : null;
     };
     get address() {
         return this._address;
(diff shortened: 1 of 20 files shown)


wozzer72:
[Account home page - Workplace](https://trello.com/c/6y2OTQmB/68-account-home-page-workplace)

wozzer72:
[Service Users Question](https://trello.com/c/6j4x0gui/50-service-users-question)

PR #428, opened 2019-03-28
Header bugs

ChazUK:
Fixing display of users first name in the header
Fixing typo on the log out link
Removing title attributes from links

### DIFF
--- a/src/app/core/header/header.component.ts
+++ b/src/app/core/header/header.component.ts
@@ -16,7 +16,7 @@ export class HeaderComponent {
   }
 
   hasFullname() {
-    return this.authService.fullname ? this.authService.fullname.split(' ')[0] : null;
+    return (this.fullname = this.authService.fullname ? this.authService.fullname.split(' ')[0] : null);
   }
 
   signOut(event) {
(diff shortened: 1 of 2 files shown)


ChazUK:
[Home page header review](https://trello.com/c/2WrJbvsA/141-home-page-header-review)

src/app/core/header/header.component.ts
```diff
@@ -16,7 +16,7 @@ export class HeaderComponent {
   }
 
   hasFullname() {
-    return this.authService.fullname ? this.authService.fullname.split(' ')[0] : null;
+    return (this.fullname = this.authService.fullname ? this.authService.fullname.split(' ')[0] : null);
```
wozzer72: Not sure my head is working, but shouldn't that be a `==` between `this.fullname` and `this.authService.fullname`?

The method name "hasFullname" suggests its returning a boolean value, but the right hand side of the predicate is returning a null or the first half of fullname.

ChazUK: yeah this is a strange one, I was just updating the previous code.
I believe if the assignment is made then the boolean value is true, otherwise null which defaults to false.

The issue was that fullname (which is now firstname) was not being set, so when I was trying to echo it in the template it was undefined

wozzer72: As long it works.

PR #432, opened 2019-04-02
Capacities are no longer mandatory

ChazUK:


ChazUK:
[Edit Workplace data: changing capacities from the dashboard errors where there is existing data](https://trello.com/c/sELANss0/378-edit-workplace-data-changing-capacities-from-the-dashboard-errors-where-there-is-existing-data)

PR #433, opened 2019-04-02
Restructuring the output on service users reference endpoint.

wozzer72:
The reference output looks equivalent to that already consumed by frontend for `.../api/services/byCategory`.


wozzer72:
[Service Users Question](https://trello.com/c/6j4x0gui/342-service-users-question)

PR #440, opened 2019-04-05
Correcting the isEqual logic to properly handle change on value based Vacancies, Starters and Leavers

wozzer72:
The Change Property `isEqual` method is what controls change audit history. The fact it (`isEqual`) was only checking for arrays not values, meant the changes to values were being tracked as change everytime.


wozzer72:
[Establishment Audit: unchanged starters / leavers / vacancies are flagged as changed](https://trello.com/c/bC0HtL7e/417-establishment-audit-unchanged-starters-leavers-vacancies-are-flagged-as-changed)

PR #441, opened 2019-04-05
Reducing login token expiry (config driven), and introducing a login …

wozzer72:
…refresh API endpoint to renew token (only if current token is valid).


wozzer72:
[Expiry Session - Log Out](https://trello.com/c/JLzv3aIF/421-expiry-session-log-out)

PR #442, opened 2019-04-05
Session Timeout

ChazUK:
Adding an Idle Service that will log the user out after `x` amount of seconds in inactivity

The service also ping's every `x` amount of seconds which can be used to refresh tokens.

### DIFF
--- a/src/app/features/login/login.component.ts
+++ b/src/app/features/login/login.component.ts
@@ -7,6 +7,9 @@ import { EstablishmentService } from '@core/services/establishment.service';
 import { IdleService } from '@core/services/idle.service';
 import { MessageService } from '@core/services/message.service';
 
+const PING_INTERVAL = 240;
+const TIMEOUT_INTERVAL = 1800;
+
 @Component({
   selector: 'app-login',
   templateUrl: './login.component.html',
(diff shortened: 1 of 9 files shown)


ChazUK:
[Expiry Session - Log Out](https://trello.com/c/JLzv3aIF/421-expiry-session-log-out)

src/app/features/login/login.component.ts
```diff
@@ -89,6 +91,15 @@ export class LoginComponent implements OnInit, OnDestroy {
 
           const token = response.headers.get('authorization');
           this.authService.authorise(token);
+
+          this.idleService.init(60, 600);
```
wozzer72: what do these hard coded numbers mean?

src/app/features/login/login.component.ts
```diff
@@ -89,6 +91,15 @@ export class LoginComponent implements OnInit, OnDestroy {
...
+          this.idleService.start();
+
+          this.idleService.ping$.subscribe(() => console.log('REFRESH BEARER TOKEN'));
```
wozzer72: I take it this is where you'll call the new API endpoint?

It needs to be able to handle 403 errors - failed refresh

wozzer72: It needs to replace the existing token.

src/app/features/login/login.component.ts
ChazUK: 60 seconds to Ping, 600 seconds to timeout. I can put them as constants in a file somewhere but they're still going to be hardcoded

src/app/features/login/login.component.ts
ChazUK: Yep, haven't implemented that yet, just a placeholder

src/app/features/login/login.component.ts
ChazUK: Updated to Constants at the top of the file

PR #452, opened 2019-04-06
Login sequelize error - and other endpoints

wozzer72:
Hi Nasir

https://trello.com/c/PLIP2Ddv

Following on from the hotfix deployment yesterday whereby it was witness sequelize errors being returned from backend on login owing to schema mismatch between code and DB, Charlie raised the defect and I've fixed it here.

I've taken the opportunity to review all existing endpoints for good error handling from sequelize (meaning I've had to add exception handling blocks) and to return basic 503 responses without disclosure of the error.

### DIFF
--- a/server/routes/login.js
+++ b/server/routes/login.js
@@ -145,7 +145,7 @@ router.post('/',async function(req, res) {
       })
       .catch((error) => {
         console.error(error);
-        return res.status(400).send(error);
+        return res.status(503).send();
       });
 });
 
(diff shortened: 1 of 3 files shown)


wozzer72:
[SequelizeDatabaseError is being returned as a response when DB Errors on an API call](https://trello.com/c/PLIP2Ddv/424-sequelizedatabaseerror-is-being-returned-as-a-response-when-db-errors-on-an-api-call)

PR #464, opened 2019-04-10
Feature | gds assessment | 427 | your account

omarCreativeDev:
- Rename `change-user-summary` route, component, spec to `your-account` 
- Created `UserAccount` interface and used in service and component
- Updated markup to confirm to GDS framework


src/app/features/your-account/your-account.component.html
```diff
@@ -0,0 +1,106 @@
...
+
+    <hr class="govuk-section-break govuk-section-break--visible">
+
+    <dl class="govuk-summary-list govuk-summary-list--no-border basic-records govuk-!-margin-bottom-0">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Fullname
+        </dt>
+        <dd class="govuk-summary-list__value">
+          {{ user?.fullname }}
```
ChazUK: very annoying that we have to check if user is populated, but I guess it'll take a lot of work to create a new BE endpoint and update the login / app init process

wozzer72: @ChazUK - not sure what your asking for. If it's to not do a username lookup during registration then that is a change of requirement rather than just an API change.

Until we can replace login using email address we are stuck with usernames, and needing to assist the user to create a unique username.

omarCreativeDev:
[Change user summary](https://trello.com/c/sVWUII8R/427-change-user-summary)

PR #465, opened 2019-04-10
patch to get daily snapshot reports API endpoint into live

wozzer72:
Hi Nasir

https://trello.com/c/G0VFyXYn

Raising this is anticipation (not hope) that we have to do a hotfix on UAT to get the daily snapshot report out.

This hotfix is:
1. A copy of the new daily snapshot endpoint itself as demonstrated in dev - all the new code is within that one single file: `server/routes/reports/dailySnapshot/index.js`
2. A dedicated route in `server.js`; the current 1.5 branch does not have a reporting route group to leverage as done in dev.


wozzer72:
Closing this PR, because it has been sensibility agreed that we will follow good dev, staging and then live deployment, rather than rushing a hotfix into live.

PR #467, opened 2019-04-11
Qualifications

ChazUK:
Add / View / Edit / Delete Qualifications

I also updated the HTTP Error Handler to throw an error on Unauthorised as it was chucking out a load of console logs.

### DIFF
--- a/src/app/core/services/worker.service.ts
+++ b/src/app/core/services/worker.service.ts
@@ -1,5 +1,12 @@
-import { HttpClient } from '@angular/common/http';
+import { HttpClient, HttpParams } from '@angular/common/http';
 import { Injectable } from '@angular/core';
+import {
+  AvailableQualificationsResponse,
+  QualificationRequest,
+  QualificationResponse,
+  QualificationsResponse,
+  QualificationType,
+} from '@core/model/qualification.model';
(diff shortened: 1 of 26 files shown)


ChazUK:
[Add/ Update worker qualifications #88](https://trello.com/c/zkIxMNm0/340-add-update-worker-qualifications-88)

src/app/core/services/worker.service.ts
```diff
@@ -111,6 +121,57 @@ export class WorkerService {
     );
   }
 
+  getAvailableQualifcations(workerId: string, type: QualificationType) {
```
wozzer72: I see you've put all the qualifications and training APIs in the same worker service. A bit late to change now, but separate qualifications and separate training services would have been more encapsulated.

ChazUK: I could pull them about but I still see these calls as directly Worker related. Happy to move it if you like.

src/assets/scss/forms.scss
```diff
@@ -91,6 +91,10 @@ form,
     .panel-border-narrow {
 
       @include rem(margin, 20px 0 0 20px);
+
+      &.no-pad {
```
denny13: no-pad - referencing margin?

You can use a class of 'ml-0' to achieve the same thing also

PR #469, opened 2019-04-11
Updating copy on Qualifications and Training Summary

ChazUK:


ChazUK:
[Qualifications and training: Reword original section as per Rich's design](https://trello.com/c/RRYlwEBd/462-qualifications-and-training-reword-original-section-as-per-richs-design)

PR #478, opened 2019-04-13
WDF Reports Rework

wozzer72:
https://trello.com/c/yBCo8FV7

Following a review with Will Fenton, it was highlighted that a workplace attains Overal WDF Eligibility at any time throughout the year, and once attained, keeps it until the effective period restarts.

Will also wanted the calculation to be based on when the user Completes their Establishment or Worker; because Establishment Complete is not yet available, this implementation is based on Establishment update.

This now means having to store the Overall WDF Eligibility. It also means we have to recalculate the Overall WDF Eligibility (which is against an Establishment and all it's Workers) whenever the Establishment or Worker is updated. To keep the code simpler, I've introduced a singleton as the `WdfCalculator` which also encapsulates the "effective from" date.

Thanks

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -798,7 +798,6 @@ class Establishment {
                 myDefaultJSON.locationRef = this.locationId;
                 myDefaultJSON.isRegulated = this.isRegulated;
                 myDefaultJSON.nmdsId = this.nmdsId;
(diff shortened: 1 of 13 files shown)


wozzer72:
[Revised WDF Reporting Requirements](https://trello.com/c/yBCo8FV7/456-revised-wdf-reporting-requirements)

PR #481, opened 2019-04-15
Feature | gds assessment | 469 | registered question page

omarCreativeDev:
- Make registered question page GDS compliant
- Plus general tidy up and refactor


omarCreativeDev:
[Registered question page](https://trello.com/c/mcZYsRGj/469-registered-question-page)

src/app/features/cqc-registered-question/cqc-registered-question.component.ts
```diff
@@ -1,35 +1,231 @@
...
+        {
+          regulatedPostcode: ['', Validators.maxLength(8)],
+          locationId: ['', Validators.maxLength(50)],
+        },
+      ),
+      nonRegulatedPostcode: ['', Validators.maxLength(8)],
+    });
+  }
+
+  public onRegulatedByCQCTruthy(): void {
```
ChazUK: Does this not need a `clearValidators()` as well?

omarCreativeDev: it doesn't because before this method is invoked the `group` has no validators once the form is setup. the `group` also has its validators cleared once `onRegulatedByCQCFalsy()` is called : )

PR #492, opened 2019-04-17
GOV.UK Framework | Staff Record | Main Job Start Date

ChazUK:
This PR touches a number of files and areas of the Staff Record Flow.

* Updated Date Validators to be more consistent and use built in `moment` functions
* Updated Date Picker Component so that it can be imported and used multiple times on a page
* Created Base Question Component that pulls out all the duplicate code from the Staff Records Questions into one class
* Removing unused SCSS that has been trailing in this branch
* Added a utils file for reusable SCSS
* Allow users to be redirected to different pages as a `return` function from Question Pages. This can either be the Staff Record page or the Check Answers Summary
* Submit Buttons Shared Component now emits events to the parent component to let them handle the logic


src/app/core/services/worker.service.ts
```diff
@@ -77,8 +77,8 @@ export class WorkerService {
     this._worker$.next(worker);
   }
 
-  setReturnToSummary(val) {
-    this.returnToSummary$.next(val);
+  setReturnTo(val: string[]) {
```
omarCreativeDev: is it possible to type the function as well as its return type?

ChazUK: Sure `void` ok?

src/app/features/workers/main-job-start-date/main-job-start-date.component.html
```diff
@@ -1,27 +1,37 @@
...
-  </form>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <h1 class="govuk-heading-xl">What date did they start their main job role?</h1>
+  </div>
 </div>
+<form (ngSubmit)="onSubmit()" [formGroup]="form">
+  <div class="govuk-form-group" [class.govuk-form-group--error]="submitted && (form.invalid || serverError)">
+    <fieldset class="govuk-fieldset" aria-describedby="mainJobStartDate-hint" role="group">
+      <!-- <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
```
omarCreativeDev: how comes this is commented out?

src/app/core/services/worker.service.ts
omarCreativeDev: sweet

src/app/features/workers/main-job-start-date/main-job-start-date.component.html
ChazUK: Good spot, wanted to discuss this with Rich

ChazUK:
[Staff Record - Main Job Start Date](https://trello.com/c/R9rjT0Td/508-staff-record-main-job-start-date)

PR #496, opened 2019-04-18
Back Link Service

ChazUK:
Adding a Back Link Service. You can set the back link by calling

```this.backService.back = { url: ['/path'] }```

It can also take `fragments` and `queryParams` as optional properties

### DIFF
--- a/src/app/core/services/back.service.ts
+++ b/src/app/core/services/back.service.ts
@@ -1,13 +1,8 @@
 import { Injectable } from '@angular/core';
-import { NavigationEnd, Params, Router } from '@angular/router';
+import { NavigationEnd, Router } from '@angular/router';
+import { URLStructure } from '@core/model/url.model';
 import { BehaviorSubject, Observable } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
-
-interface URLStructure {
-  url: string[];
-  fragment?: string;
-  queryParams?: Params;
-}
 
 @Injectable({
   providedIn: 'root',
--- a/src/app/core/services/back.service.ts
+++ b/src/app/core/services/back.service.ts
@@ -8,8 +8,8 @@ import { filter, map } from 'rxjs/operators';
   providedIn: 'root',
 })
 export class BackService {
-  private _back$: BehaviorSubject<URLStructure> = new BehaviorSubject<URLStructure>(null);
-  public back$: Observable<URLStructure> = this._back$.asObservable();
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/app/core/services/back.service.ts [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+import { NavigationEnd, Router } from '@angular/router';
+import { URLStructure } from '@core/model/url.model';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class BackService {
+  private _back$: BehaviorSubject<URLStructure> = new BehaviorSubject<URLStructure>(null);
```
omarCreativeDev: is there any benefit to prepending an underscore? the private type should be enough no?

src/app/core/services/back.service.ts
```diff
@@ -0,0 +1,29 @@
...
+  public back$: Observable<URLStructure> = this._back$.asObservable();
```
omarCreativeDev: i dont see why this is needed if we have the above BehaviourSubject?

src/app/core/services/back.service.ts
ChazUK: I used the underscore for Behaviour Subjects and then without for the actual Observable that Components can subscribe to.

src/app/core/services/back.service.ts
ChazUK: No idea tbh, I saw this pattern being used somewhere else, I think the general premise is that it stops the direct access to the BehaviourSubject

src/app/core/services/back.service.ts
omarCreativeDev: I just feel as a behaviour subject is essentially an observable there's no point in creating another

src/app/core/services/back.service.ts
omarCreativeDev: thats fair and i see your point. however i think in our application [from what i have seen so far] the state managed observables need to be accessible by different routes/components and therefore the observable stream should be accessable/public anyway : )

ChazUK:
Updated to better handle state shown here https://dev.to/avatsaev/simple-state-management-in-angular-with-only-services-and-rxjs-41p8

This will become redundant once NGRX or NGXS is implemented.

ChazUK:
[Back Link Service](https://trello.com/c/CwPBM3xI/506-back-link-service)

PR #501, opened 2019-04-18
GOV.UK Framework | Staff Record | Mental Health Professional

ChazUK:
Also fixed an issue around removing a pre-selected job role from Other Jobs. This should be handled on the Server but currently isn't


ChazUK:
[Staff Record - Mental Health Professional](https://trello.com/c/DlIgdRCn/509-staff-record-mental-health-professional)

PR #511, opened 2019-04-22
bugfix | non cqc regulated workplace lookup

omarCreativeDev:
- fix routing whether cqc regulated or not
- move workplace locations data stream


src/app/core/model/registration.model.ts
```diff
@@ -9,12 +9,5 @@ export interface RegistrationModel {
...
   locationdata: Array<WorkplaceLocation>;
-  postcodedata: [{
-    locationName: string;
-    addressLine1: string;
-    addressLine2: string;
-    townCity: string;
-    county: string;
-    postalCode: string;
-  }];
+  postcodedata: Array<WorkplaceLocation>;
```
ChazUK: Almost seems like it'd be better to change this interface to Address or LocationAddress

omarCreativeDev: done : )

PR #517, opened 2019-04-23
Making consistent the responses to frontend on failed login.

wozzer72:


wozzer72:
[Log in screen: Authentication failure message is no longer being displayed](https://trello.com/c/d8t2m7Eh/293-log-in-screen-authentication-failure-message-is-no-longer-being-displayed)

PR #543, opened 2019-04-25
Feature/wdf reporting

denny13:


src/app/features/reports/reports.component.ts
```diff
@@ -0,0 +1,91 @@
...
+        this.reportsService.updateState(this.eligibility);
+      })
+    );
+  }
+
+  ngOnDestroy() {
+    this.subscriptions.unsubscribe();
+  }
+
+  formValidator(formGroup: FormGroup): ValidationErrors {
```
ChazUK: There's no validation error being returned here

denny13: This was me quickly trying to reuse some date stuff from somewhere else, I will be cleaning that up along with adding the date stuff for Ann.

PR #547, opened 2019-04-25
Skeleton Records Bugs

ChazUK:
Adding Start Screen for Basic Records
Save and Return for Total Staff

### DIFF
--- a/src/app/core/services/worker.service.ts
+++ b/src/app/core/services/worker.service.ts
@@ -39,6 +39,7 @@ export class WorkerService {
   public worker$ = this._worker$.asObservable();
   private lastDeleted$ = new BehaviorSubject<string>(null);
   private returnToSummary$ = new BehaviorSubject<boolean>(false);
+  private totalStaffReturn$ = new BehaviorSubject<boolean>(false);
   public createStaffResponse = null;
   public trainingRecordCreated = null;
   public trainingRecordEdited = null;
@@ -65,6 +66,10 @@ export class WorkerService {
     return this.returnToSummary$.value as boolean;
   }
 
+  public get totalStaffReturn() {
+    return this.totalStaffReturn$.value as boolean;
+  }
+
   setLastDeleted(name: string) {
     this.lastDeleted$.next(name);
   }
@@ -79,6 +84,10 @@ export class WorkerService {
 
   setReturnToSummary(val) {
     this.returnToSummary$.next(val);
+  }
+
+  setTotalStaffReturn(val) {
+    this.totalStaffReturn$.next(val);
   }
 
   getWorker(workerId: string): Observable<Worker> {
--- a/src/app/features/workers/create-basic-records-start-screen/create-basic-records-start-screen.component.html
+++ b/src/app/features/workers/create-basic-records-start-screen/create-basic-records-start-screen.component.html
@@ -15,6 +15,6 @@
     adding or removing records.
   </p>
   <div class="btn-container mt-5">
-    <app-start-button link="/worker/create-staff-record"></app-start-button>
+    <app-start-button link="/worker/total-staff"></app-start-button>
   </div>
 </div>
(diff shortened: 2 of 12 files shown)


ChazUK:
[Workplace dashboard showing wrong info for New starters](https://trello.com/c/CnaJ37pP/560-workplace-dashboard-showing-wrong-info-for-new-starters)

ChazUK:
[User should be able to update Total Staff number from the dashboard without entering more staff](https://trello.com/c/lqifxtmU/562-user-should-be-able-to-update-total-staff-number-from-the-dashboard-without-entering-more-staff)

ChazUK:
[Full record flow doesn't show the start page.](https://trello.com/c/iKaJteGH/561-full-record-flow-doesnt-show-the-start-page)

src/app/features/workers/create-basic-records-start-screen/create-basic-records-start-screen.component.html
```diff
@@ -0,0 +1,20 @@
...
+  <h1 class="h1">Staff records</h1>
+  <p>To create staff records for your workplace, for each record you will need to add their:</p>
+  <ul class="list list--bullet">
+    <li>full name or ID number</li>
+    <li>main job role</li>
+    <li>type of contract</li>
+  </ul>
+  <p>Once you have added these you will be able to continue with each record.</p>
+  <p>
+    We will use this information to work out your total number of staff. You will be able to change the number later by
```
wozzer72: Isn't it a little confusing now to say "we will use this information to work out your total number of staff"? We're no longer updating total staff, so what does "work out" really mean?

src/app/features/workers/total-staff/total-staff.component.ts
```diff
@@ -11,16 +12,22 @@ import { Subscription } from 'rxjs';
...
     private messageService: MessageService,
-    private establishmentService: EstablishmentService
+    private establishmentService: EstablishmentService,
+    private workerService: WorkerService
   ) {}
 
   ngOnInit() {
+    this.returnToDash = this.workerService.totalStaffReturn;
+
+    console.log(this.returnToDash);
```
wozzer72: remove the console debug

src/app/core/services/worker.service.ts
```diff
@@ -65,6 +66,10 @@ export class WorkerService {
     return this.returnToSummary$.value as boolean;
   }
 
+  public get totalStaffReturn() {
+    return this.totalStaffReturn$.value as boolean;
```
wozzer72: This is a frontend value not an API value yes? When I first started to review this I thought it was a wrapper for Establishment number of staff.

But this is actually the count of worker records I believe having reviewed the rest of the changes below.
In mathematics, "total" is the sum of something, hence "total staff" suggest the "sum of staff".

Perhaps a name more like "currentNumberOfStaff"? Or "countOfStaff"?

Also, what is the suffix "Return" implying?

I know this is tedious reasoning, but great value comes from the name of attributes/methods, especially when there are two similar values (that declared and call upon via the API for establishment and that being the count of Worker records).

src/app/features/workers/create-basic-records-start-screen/create-basic-records-start-screen.component.html
ChazUK: Waiting on copy from Charlie

src/app/core/services/worker.service.ts
ChazUK: This is just a boolean value that dictates wether or not the totalStaff should return to the dashboard, yes it's pretty badly named, but I want to review how this kind of thing works in the updated GDS branch as we're using this in a few other places.

src/app/features/workers/total-staff/total-staff.component.ts
ChazUK: Removed

PR #549, opened 2019-04-26
bugfix | highlight form fields on unauthorized user login

omarCreativeDev:
- highlight form fields on unauthorized user login
- corrected error classes

### DIFF
--- a/src/app/features/login/login.component.ts
+++ b/src/app/features/login/login.component.ts
@@ -80,6 +80,10 @@ export class LoginComponent implements OnInit, OnDestroy {
         message: 'User unauthorised - username or password is incorrect.',
       },
       {
+        name: 404,
+        message: 'User not found.',
+      },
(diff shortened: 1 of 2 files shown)


src/app/features/login/login.component.ts
```diff
@@ -79,6 +79,10 @@ export class LoginComponent implements OnInit, OnDestroy {
         name: 401,
         message: 'User unauthorised - username or password is incorrect.',
       },
+      {
```
ChazUK: This is more of a BE thing, but I think sending 401 and 404 is a security risk as it shows that a user exists in the system or not.

omarCreativeDev: @wozzer72 @ChazUK perhaps this is better off as a seperate discussion for monday and if needed we can create a new ticket/pr for the extra work

PR #552, opened 2019-04-26
bugfix | IE11 unable to select first dropdown value, also fix compile…

omarCreativeDev:
also fix compile issues

### DIFF
--- a/src/app/features/registration/select-workplace-address/select-workplace-address.component.html
+++ b/src/app/features/registration/select-workplace-address/select-workplace-address.component.html
@@ -34,7 +34,7 @@
           name="address"
           (change)="onLocationChange($event.target.value)"
         >
-          <option [value]="" disabled>{{ locationAddresses.length }} addresses found</option>
+          <option [value]="">{{ locationAddresses.length }} addresses found</option>
           <option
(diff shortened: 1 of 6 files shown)


src/app/features/registration/select-workplace-address/select-workplace-address.component.html
```diff
@@ -34,7 +34,7 @@ <h1 class="govuk-heading-l">What is your workplace address?</h1>
           name="address"
           (change)="onLocationChange($event.target.value)"
         >
-          <option [value]="" disabled>{{ locationAddresses.length }} addresses found</option>
+          <option [value]="">{{ locationAddresses.length }} addresses found</option>
```
ChazUK: You can set value to null here which is better than an empty string.

omarCreativeDev: by using null it renders out the first option as empty

omarCreativeDev:
[Accessibility version of select address page - cannot select first address listed](https://trello.com/c/AOzr3nnx/566-accessibility-version-of-select-address-page-cannot-select-first-address-listed)

PR #554, opened 2019-04-28
Sticky Footer

ChazUK:
Adding in a flex based sticky footer.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,6 +1,6 @@
 <app-header></app-header>
 
-<div class="govuk-width-container">
+<div id="main" class="govuk-width-container">
   <app-phase-banner phase="beta">
     This is a new service – your
     <a routerLink="/feedback" target="_blank">feedback</a> will help us to improve it.
--- a/src/assets/scss/modules/_sticky-footer.scss
+++ b/src/assets/scss/modules/_sticky-footer.scss
@@ -1,8 +1,16 @@
+html, body {
+  height: 100%;
+}
+
+body {
+  margin: 0;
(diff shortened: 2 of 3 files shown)


ChazUK:
[Sticky Footer](https://trello.com/c/CcKC3B11/585-sticky-footer)

src/assets/scss/modules/_sticky-footer.scss
```diff
@@ -0,0 +1,27 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+app-root {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+
+  > * {
```
omarCreativeDev: why was this needed?

src/app/app.component.html
```diff
@@ -1,6 +1,6 @@
 <app-header></app-header>
 
-<div class="govuk-width-container">
+<div id="main" class="govuk-width-container">
```
omarCreativeDev: could we use the <main> tag instead?

src/assets/scss/modules/_sticky-footer.scss
ChazUK: For some reason the width would reduce to like 50%, I'm guessing it's something to do with flex turning child elements to inline-block

src/app/app.component.html
ChazUK: This template has been copied from https://design-system.service.gov.uk/styles/page-template/custom/index.html, and we need flex working on app header, footer and the div in between them. Annoyingly the `main` tag is too nested

PR #556, opened 2019-04-29
Summary Back logs user out

ChazUK:
Adding missing goBack function


ChazUK:
[Back navigation from Staff record summary is logging the user out](https://trello.com/c/RACUFEwS/586-back-navigation-from-staff-record-summary-is-logging-the-user-out)

PR #557, opened 2019-04-29
Updating link on Home Tab to direct users to Skeleton Records

ChazUK:


ChazUK:
[When no staff records exist 'Orange banner' should take user into the skeleton record flow.](https://trello.com/c/VuhrMMUo/592-when-no-staff-records-exist-orange-banner-should-take-user-into-the-skeleton-record-flow)

PR #565, opened 2019-04-30
feature | create account management module

omarCreativeDev:


omarCreativeDev:
[account management module](https://trello.com/c/1faic4hp/618-account-management-module)

PR #572, opened 2019-04-30
Fixing issue with Back link on Total Staff Question

ChazUK:


src/app/shared/components/total-staff-panel/total-staff-panel.component.ts
```diff
@@ -14,6 +14,7 @@ export class TotalStaffPanelComponent {
   constructor(private workerService: WorkerService) {}
 
   setReturn() {
+    console.log(true);
```
omarCreativeDev: can this be removed?

ChazUK: removed

PR #573, opened 2019-04-30
Correct handling of non-empty property for WDF Eligibility calculation

wozzer72:


wozzer72:
[Couple of issues with WDF compliance.](https://trello.com/c/QyEJqKLp/624-couple-of-issues-with-wdf-compliance)

PR #593, opened 2019-05-08
bugfix | show todays date as today string rather than date

omarCreativeDev:


omarCreativeDev:
[Dashboard UI Issues: Last updated date](https://trello.com/c/Sni5MhWW/200-dashboard-ui-issues-last-updated-date)

PR #595, opened 2019-05-08
GOV.UK Framework | Staff Record | Training and Qulifications

ChazUK:
* Updated Date Picker to only show errors when submitted is true
* Fixed an error with character count where max length value was set to null
* Updated Staff Record page to fit GDS standards
* Add and View Qualifications
* Add and View Training

I'm not 100% happy with this, there is still a lot left around error messages that needs to be fixed. These pages should definitely be revisited


ChazUK:
[Staff Record - Training and Qualifications](https://trello.com/c/oitOlLp1/714-staff-record-training-and-qualifications)

PR #598, opened 2019-05-09
feature | gds assessment | feedback page

omarCreativeDev:
- handle form errors and server errors
- fix error class on <textarea> everywhere


omarCreativeDev:
[feedback page](https://trello.com/c/9a17IkJD/715-feedback-page)

src/app/features/feedback/feedback.component.ts
```diff
@@ -1,36 +1,112 @@
...
+  private setupForm(): void {
     this.form = this.formBuilder.group({
-      tellUs: [null, [Validators.required, Validators.maxLength(500)]],
-      doingWhat: [null, [Validators.required, Validators.maxLength(500)]],
-      fullname: [null, [Validators.maxLength(120)]],
-      email: [null, [Validators.email, Validators.maxLength(120)]],
+      doingWhat: [null, [Validators.required, Validators.maxLength(this.doingWhatCharacterLimit)]],
+      email: [null, [Validators.email, Validators.maxLength(this.emailCharacterLimit)]],
+      fullname: [null, [Validators.maxLength(this.fullNameCharacterLimit)]],
+      tellUs: [null, [Validators.required, Validators.pattern(TEN_WORDS_PATTERN)]],
```
ChazUK: It might be better making a custom Validator for maxWords, the character count component already handles max characters and also can handle max words

src/app/features/feedback/feedback.component.ts
```diff
@@ -1,36 +1,112 @@
...
+        ],
+      },
+    ];
+  }
+
+  private setupServerErrorsMap(): void {
+    this.serverErrorsMap = [
+      {
+        name: 503,
+        message: 'Unable to update database.',
```
ChazUK: This error doesn't seem very helpful to the user

wozzer72:
Hey @ChazUK and @omarCreativeDev 

Alas. My first component now looking pretty.

Yes, the given feedback is stored in the database. Why? For prosperity? Showing of SQL skills? Or more so, because at the time, that was all we had.

But, since this component was written, the backend is notifying the SFC Service Team through Slack.

So in respect of the 503 error (I would have thought something like "Unable to contact SFC Service Desk. Please try again later or call the service desk on.....", something Charlie Stewart can do more with. 

But the main reason for my comment is refactoring feedback into a separate module. Feedback opens a new tab. It would be great if it didn't take three days and an eternity to simply show four fields which have no dependency on any other part of the app. I've seen how refactoring into separate modules has benefitted other parts of of the app, if it a simple thing to do here, then it would be an easy win.

src/app/features/feedback/feedback.component.ts
omarCreativeDev: done

src/app/features/feedback/feedback.component.ts
omarCreativeDev: updated copy

omarCreativeDev:
@wozzer72 cheers for the suggestion warren, I will do this in a separate pr. will include you to the pr once ready

PR #600, opened 2019-05-10
bugfix | hide server error if cqc regulated radio is toggled

omarCreativeDev:
also prettify file


omarCreativeDev:
[CQC Regulated question: previous errors should be cleared when the user swaps between regulated and not regulated](https://trello.com/c/poReWr02/205-cqc-regulated-question-previous-errors-should-be-cleared-when-the-user-swaps-between-regulated-and-not-regulated)

PR #609, opened 2019-05-14
Parents and subs "view my workplaces"

wozzer72:
https://trello.com/c/ikScd2O3

This PR includes the original structural changes to introduce a new endpoint, but significantly, it now includes support for passing Establishment UID rather than ID (the internal primary key) on every establishment, user (account) and worker (including training and qualifications) endpoints.

All regression tests ran, updated and re-ran.

This PR also includes the removing of previously pushed migration specific `localhost.yaml`.


wozzer72:
[Parent and Subsidiaries - View My Workplaces](https://trello.com/c/ikScd2O3/3-parent-and-subsidiaries-view-my-workplaces)

nasirmids:
Looks all ok; I like to the idea of early review as well otherwise it become so many files

server/routes/establishments/capacity.js
```diff
@@ -11,22 +11,10 @@ router.route('/').get(async (req, res) => {
...
-  } else {
-    // unexpected establishment id
-    return res.status(400).send();
-  }
-
   const thisEstablishment = new Establishment.Establishment(req.username);
 
   try {
-    if (await thisEstablishment.restore(byID, byUUID, showHistory)) {
+    if (await thisEstablishment.restore(establishmentId, showHistory)) {
```
nasirmids: Quetion; in the description you have mentioned all endpoint now support establishment UID rather than ID but all the functions from here below seems to be doing other way i.e. removing UID and accepting ID?

nasirmids: Quetion; in the description you have mentioned all endpoint now support establishment UID rather than ID but all the functions from here below seems to be doing other way i.e. removing UID and accepting ID?

PR #610, opened 2019-05-14
Defect | Back Link now persists when navigating fragments

ChazUK:
Added url-parse package
Back Link now checks to see if the URL portion of the Location URL has changed before removing the Back Link


ChazUK:
[Accessibility - Back Link Display](https://trello.com/c/hW20bbic/54-accessibility-back-link-display)

PR #612, opened 2019-05-15
Bulkupload api presign url

nasirmids:
Hello, 

Summary

Here is the trello card https://trello.com/c/kIUwAZTN  and has same details as below with screen shots.

This is an end point to retrieve the presign URL and it accept establishment name and filename. It is valid for 5 minutes at the moment. The url is using aws user key to upload the file and set the relevent values

API end point is http://localhost:3000/api/establishment/144/bulkupload/signedUrl
S3 Bucket Name Sfcbulkuploadfiles
Path https://s3.conssfcs3policyonlyole.aws.amazon.com/s3/buckets/sfcbulkuploadfiles/?region=eu-west-2&tab=overview
AWS IAM User = BulkUploadUser (policy name is sfcs3policyonly and can only access s3)
Meta data= I am storing few information here regarding the file
versioning = no

This is so far best of knowledge/scenario  i have and it might change if requirement changes but not expecting massively.


server/config/config.js
```diff
@@ -207,7 +207,33 @@ const config = convict({
           env: 'SLACK_LEVEL',
           default: 0
       }
-  }
+  },
+  bulkuploaduser: {
+    accessKeyId: {
+      doc: 'The BU user',
```
wozzer72: What does BU mean? Don't use acronyms.

server/config/config.js
```diff
@@ -207,7 +207,33 @@ const config = convict({
...
+      format: '*',
+      default: '',
+      env: 'Access_KEY_ID',
```
wozzer72: Why is Access lowercase?

server/config/config.js
```diff
@@ -207,7 +207,33 @@ const config = convict({
           env: 'SLACK_LEVEL',
           default: 0
       }
-  }
+  },
+  bulkuploaduser: {
+    accessKeyId: {
```
wozzer72: I presume this is the AWS Access Key. Why is there an access key/secret for just bulk upload?

server/config/config.js
```diff
@@ -207,7 +207,33 @@ const config = convict({
...
+      doc: 'The BU user',
+      format: '*',
+      default: '',
+      env: 'Access_KEY_ID',
+    },
+    secretAccessKey: {
+      doc: 'The BU user',
```
wozzer72: Description needs to be informative. So the fact that you have the same description on all means it is not informative.

server/config/config.js
```diff
@@ -207,7 +207,33 @@ const config = convict({
...
+      format: '*',
+      default: '',
+      env: 'S3_BUCKET',
+    },
+    region: {
+      doc: 'The BU user',
+      format: '*',
+      default: '',
+      env: 'S3_REGION',
```
wozzer72: Why does this need to be an environment variable?

server/routes/establishments/bulkUpload.js
```diff
@@ -0,0 +1,48 @@
...
+      // ACL: 'public-read',
+      ContentType: req.query.type,
+      Metadata: {
+        dateuploaded: new Date().toString(),
+        username: '',
+        establishmentid: establishmentId.toString(),
+        filestatus: FileStatusEnum.New,
+        validationstatus: FileValidationStatusEnum.Pending
+      },
+      Expires: 300, // 5 minutes
```
wozzer72: should be a constant, rather than hard coded

wozzer72: Even better, should be a config variable.

server/routes/establishments/bulkUpload.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  try {
+    var uploadPreSignedUrl = s3.getSignedUrl('putObject', {
+      Bucket: appConfig.get('bulkuploaduser.bucketname').toString(),
+      Key: establishmentId + '/new/' + req.query.filename,
+      // ACL: 'public-read',
+      ContentType: req.query.type,
+      Metadata: {
+        dateuploaded: new Date().toString(),
```
wozzer72: Why is dateuploaded metadata? Surely, an S3 object has a created date already.

server/routes/establishments/bulkUpload.js
```diff
@@ -0,0 +1,48 @@
...
+    accessKeyId: appConfig.get('bulkuploaduser.accessKeyId').toString(),
+    secretAccessKey: appConfig.get('bulkuploaduser.secretAccessKey').toString(),
+    region: appConfig.get('bulkuploaduser.region').toString(),
+    useAccelerateEndpoint: true,
+  });
+
+  try {
+    var uploadPreSignedUrl = s3.getSignedUrl('putObject', {
+      Bucket: appConfig.get('bulkuploaduser.bucketname').toString(),
+      Key: establishmentId + '/new/' + req.query.filename,
```
wozzer72: Shouldn't `new` here be taken from the enum?

server/routes/establishments/bulkUpload.js
```diff
@@ -0,0 +1,48 @@
...
+      // ACL: 'public-read',
+      ContentType: req.query.type,
+      Metadata: {
+        dateuploaded: new Date().toString(),
+        username: '',
+        establishmentid: establishmentId.toString(),
+        filestatus: FileStatusEnum.New,
```
wozzer72: You have status as meatdata and in object path. This is duplication that needs to be managed. Why dupkicate?

server/routes/establishments/bulkUpload.js
```diff
@@ -0,0 +1,48 @@
...
+        validationstatus: FileValidationStatusEnum.Pending
+      },
+      Expires: 300, // 5 minutes
+    });
+    res.json({ success: true, message: 'Pre-Signed', urls: uploadPreSignedUrl });
```
wozzer72: Why are you retuning "status" of a fixed value of true? It adds nothing to the response.

server/routes/establishments/bulkUpload.js
```diff
@@ -0,0 +1,48 @@
...
+    res.end();
+  }
+  catch (err) {
+    console.error('establishment::bulkupload GET/:PreSigned - failed', err.message);
+    return res.status(503).send('Failed to restrieved Presigned URL');
```
wozzer72: Angular prefers JSON as I found out yesterday when wanting to raise defect on login failure as a result of GDS refactoring.
Any reason why you need to return the message? What value is it adding? This endpoint does not more other than returning a presigned URL. Returning 503 is enough.

server/config/config.js
nasirmids: ok to that, no need this to be env variable

server/routes/establishments/bulkUpload.js
nasirmids: removed

server/routes/establishments/bulkUpload.js
nasirmids: indeed a good point; i was lazy in thinking to read from metadata than parsing the path but everytime status change; path will also changed accordingly. I remove it now

server/routes/establishments/bulkUpload.js
nasirmids: i have notice and Just done before you commit :)

server/routes/establishments/bulkUpload.js
nasirmids: ok it make no sense for Front end to have this message; i removed

server/routes/establishments/bulkUpload.js
nasirmids: initially i thought the front end want to check status true but url itself is success response; do removing and keeping only "urls"

nasirmids:
All changes done; may be few things change when front end guys actually pick this.

server/config/config.js
nasirmids: This is bulk upload specific user access which is limited to s3 bucket.

nasirmids:
Two key points after discussion with Warren.

- Establishment ID need to replaced with UUID
- AWS access key need to be application level and managed more properly otherwise we are end up managing multiple access keys for different task. I am going to raise this trello to have this managed as key manager

PR #615, opened 2019-05-15
Parents and subs access establishments and workers

wozzer72:
Hi Nasir

https://trello.com/c/ikScd2O3

Well, this is the last of the big changes for this "parent and subs - view subsidiaries".

This one changes the very rigid authorisation that restricts establishments to users of the establishments only.

The main crux of the changes are in: `server/utils/security/isAuthenticated.js`. It might serve best for me to walk you through over appear.in.

I'm not overly happy with the code; because its middleware, I'm having to use promise callbacks which makes conditional paths hard! Whilst there is some duplication, there is a much bigger piece of work pending around permissions in general, where I expect such authentication logic to be completed encapsulated.

Thanks

### DIFF
--- a/resources/establishment_types.schema.json
+++ b/resources/establishment_types.schema.json
@@ -89,10 +89,10 @@
             "type" : "string",
             "enum" : ["Workplace", "Parent"]
         },
-        "DataOwnerPermissions" : {
-            "description" : "Associated with an Establishment, this represents permissions granted to the owner (typically the parent). If workplace, then only workplace no worker access. If staff, then owner has access to both workplace and staff.",
+        "ParentPermissions" : {
+            "description" : "Associated with an Establishment, this represents permissions granted to the parent. If Workplace, then only workplace no worker access. If Workplace and Staff, then parent has access to both workplace and staff. If null/not present, parent has no access.",
             "type" : "string",
-            "enum" : ["Workplace", "Staff"]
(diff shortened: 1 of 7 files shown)


wozzer72:
[Parent and Subsidiaries - View My Workplaces](https://trello.com/c/ikScd2O3/3-parent-and-subsidiaries-view-my-workplaces)

ChazUK:
After taking a quick look at the code I see that this includes the update to use the UID instead of the integer ID for Establishments in the API Calls. But I noticed that you've got a switch in there. Would it not be worth it to just rely on the UID and remove the code for ID's to avoid redundant code bloat?

wozzer72:
@ChazUK - I some respects it would have been easier to simply drop the establishment id in the request, but, timing is everything.

To make the API available to frontend, my only choice is to deploy to dev. This would then break the existing frontend, and would put pressure on the frontend to implement their fix.

Alternatively, the frontend works off a separate branch. But, then my fear would be that work stops on parents and sub, this work is then never merged, and subsequent delays would then make merging the changes more difficult.

So if I support both, I can deploy now. By supporting both, I've done the hard work. My integration tests have a few test for integer based id, but most default now to UID. And it will be easy enough to come back later and remove the old id style parameter.

wozzer72:
Approved by Nasir and agreed with Charlie my feedback on his comment regarding maintain both legacy id and new uid. We agreed to deploy.

PR #618, opened 2019-05-16
Disabling Registration

ChazUK:


ChazUK:
[Stop registrations in ASC-WDS](https://trello.com/c/d2NzIrSg/56-stop-registrations-in-asc-wds)

PR #621, opened 2019-05-21
fix alignment issues

omarCreativeDev:
- display block is needed on the angular component tags otherwise firefox and internet explorer show a weird space between the list number and the component


omarCreativeDev:
[Look of Bulk upload page not as per design](https://trello.com/c/YIvQOsET/64-look-of-bulk-upload-page-not-as-per-design)

PR #623, opened 2019-05-22
bugfix | bulk upload | alignment issues

omarCreativeDev:
- second attempt at fixing this but this time a cross browser solution that works in IE, firefox and chrome and no more alignment issues.
- gds classes such as `govuk-grid-row` and column width classes was the root cause for the `<ol>` list to not line up its inner content properly. so have simplified the markup


omarCreativeDev:
[Look of Bulk upload page not as per design](https://trello.com/c/YIvQOsET/64-look-of-bulk-upload-page-not-as-per-design)

PR #624, opened 2019-05-22
Bugfix/bulk upload file types

omarCreativeDev:
- created `FileUpload` interface that extends `File` interface and additionally stores file extension
- fixed bug where file types were not correctly validated
- some updates to markup to match design
- only show error summary if form has been submitted
- prettify files


omarCreativeDev:
[Bulk upload : Choose files - Gives error if .zip or .csv files are selected](https://trello.com/c/uGgc3I0s/70-bulk-upload-choose-files-gives-error-if-zip-or-csv-files-are-selected)

PR #625, opened 2019-05-22
make IE11 show validation errors

omarCreativeDev:
- clear sync validators for fileUpload when files are cleared
- use a timeout when setting sync validators for file upload to fix validation for IE11

### DIFF
--- a/src/app/shared/validators/custom-form-validators.ts
+++ b/src/app/shared/validators/custom-form-validators.ts
@@ -60,7 +60,8 @@ export class CustomValidators extends Validators {
     });
 
     if (Object.keys(errors).length) {
-      return fileUpload.setErrors(errors);
+      // timeout needed for IE11
+      setTimeout(() => fileUpload.setErrors(errors));
     }
 
     return null;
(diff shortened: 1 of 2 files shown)


omarCreativeDev:
[Bulk upload : Choose files - No error messages shown in IE11](https://trello.com/c/rwVOdUfH/69-bulk-upload-choose-files-no-error-messages-shown-in-ie11)

src/app/shared/validators/custom-form-validators.ts
```diff
@@ -60,7 +60,8 @@ export class CustomValidators extends Validators {
     });
 
     if (Object.keys(errors).length) {
-      return fileUpload.setErrors(errors);
+      // timeout needed for IE11
+      setTimeout(() => fileUpload.setErrors(errors));
```
wozzer72: What is the timeout value if no timeout specified?

Also, if this is an IE11 workaround, isn't there an Angular helper that supports browser detection? I'm doubting this is the first, and it won't be the last workaround for IE11.

https://stackoverflow.com/questions/48182912/how-to-detect-browser-with-angular

omarCreativeDev: the timeout duration is optional and if not specified is 0.

I dont feel we need to detect the browser type, as the `setTimeout` solution works on modern browsers too. Generally speaking modern javascript features that are not supported would be fixed by adding polyfills but setTimeout is pretty old skool

wozzer72: Happy with that then.

PR #626, opened 2019-05-22
bugfix | remove local authorities extra row

omarCreativeDev:
- removed `this.addLocalAuthority();` which was ultimately trying to create an authority without a custodianCode and therefore displaying a select dropdown with no authority selected
- remove empty sass file and its reference
- alphabetised imports

### DIFF
--- a/src/app/features/workplace/data-sharing-with-local-authorities/data-sharing-with-local-authorities.component.ts
+++ b/src/app/features/workplace/data-sharing-with-local-authorities/data-sharing-with-local-authorities.component.ts
@@ -1,14 +1,13 @@
-import { Component } from '@angular/core';
-import { FormArray, FormBuilder, FormGroup } from '@angular/forms';
-import { Router } from '@angular/router';
-import { DataSharingOptions } from '@core/model/data-sharing.model';
 import { BackService } from '@core/services/back.service';
+import { Component } from '@angular/core';
+import { DataSharingOptions } from '@core/model/data-sharing.model';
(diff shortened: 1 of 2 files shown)


omarCreativeDev:
[Sharing with LAs: Superfluous entry row is display when page is accessed](https://trello.com/c/byJmTRxA/66-sharing-with-las-superfluous-entry-row-is-display-when-page-is-accessed)

src/app/features/workplace/data-sharing-with-local-authorities/data-sharing-with-local-authorities.component.ts
```diff
@@ -69,8 +68,6 @@ export class DataSharingWithLocalAuthoritiesComponent extends Question {
       }
     });
 
-    this.addLocalAuthority();
```
wozzer72: Is this not creating an initial empty value in the list, so when the list is displayed it automatically says "Select Local Authority" sort of thing?

omarCreativeDev: this is responsible for an extra dropdown selectbox which never gets populated. There is a dedicated `Add authority` button which works as expected in order to create a select box

wozzer72: image

Confirmation received from Rich Rees that the dropdown is to be removed (was previously intentional within UI design.

PR #627, opened 2019-05-22
bugfix | local authorities not deleting properly

omarCreativeDev:
- fix to correct `POST` payload effectively ensuring local authorities can be deleted


omarCreativeDev:
[Sharing with LAs: Cannot remove 'Other' LAs](https://trello.com/c/XKIVxy98/65-sharing-with-las-cannot-remove-other-las)

PR #629, opened 2019-05-23
bugfix | ie11 details tag polyfill

omarCreativeDev:
this polyfill is required as IE11 doesn't support the collapse/expand behavior thats comes out of the box with modern browsers.


omarCreativeDev:
[Sharing your data not working in IE](https://trello.com/c/pgRRUmVQ/73-sharing-your-data-not-working-in-ie)

PR #631, opened 2019-05-23
bugfix | bulk upload files still upload after removal

omarCreativeDev:
- ensure required validator is set whilst clearing other validators when files are removed
- ensure error summary form errors event is only dispatched if submitted flag is true when files are removed


omarCreativeDev:
[Bulk upload : Choose files - Upload appears to go ahead after files are removed](https://trello.com/c/l6azofLD/71-bulk-upload-choose-files-upload-appears-to-go-ahead-after-files-are-removed)

PR #633, opened 2019-05-24
Feature | parents and subs | view my workplaces

omarCreativeDev:
- new view my workplaces screen conditionally accessed if user `isParent`
- created relevant components, new `ParentGuard`, interfaces
- updated workplace routing and module
- created colors sass file and created some variables
- created generic `govuk-util__list-record` class which can be used on multiple places in new upcoming screens


omarCreativeDev:
[(4) Parent and Subsidiaries - View My Workplaces](https://trello.com/c/ikScd2O3/3-4-parent-and-subsidiaries-view-my-workplaces)

src/app/core/model/my-workplaces.model.ts
```diff
@@ -0,0 +1,30 @@
+export interface MyWorkplacesResponse {
+  primary: {
```
wozzer72: primary is also a `MyWorkplace`

omarCreativeDev: thanks for spotting, just fixed

PR #638, opened 2019-05-28
LA is equal use "CssrId" in comparison

wozzer72:
https://trello.com/c/XKIVxy98 - whilst delivering on the GDS changes, testing identified an update fault on Local Authorities, whereby it was actually updating the data store.

Many weeks ago, we refactored Local Authorities within the backend and database, but we agreed to keep the API exposed interface the same (Local Authorities and custodian codes).

Unfortunately, the `isEqual` didn't keep up and still referenced the old `custodianCode` assumed from the database.


wozzer72:
[Sharing with LAs: Cannot add/remove 'Other' LAs in some cases](https://trello.com/c/XKIVxy98/65-sharing-with-las-cannot-add-remove-other-las-in-some-cases)

PR #639, opened 2019-05-30
Bug Fix | Establishment id not being set on routing

ChazUK:


ChazUK:
[Cannot create new workplace details - user is immediately signed out](https://trello.com/c/XPA15MkW/98-cannot-create-new-workplace-details-user-is-immediately-signed-out)

PR #640, opened 2019-05-31
Bug Fix | Worker State and Staff Details Question

ChazUK:
Updating Staff Details Component to work with an unset Worker
Updating Question Component to properly keep the Worker state up to date
Updating surrounding Worker Questions to use the updated Worker state rather than the form values
Fixed an issue with navigation between Mental Health Professional and Main Job Start Dtate Questions


ChazUK:
[Cannot create worker using the 'Add full record' flow](https://trello.com/c/SpcMzQ6g/99-cannot-create-worker-using-the-add-full-record-flow)

PR #646, opened 2019-05-31
80 service users other input field

dbjsdev:


server/models/classes/establishment.js
```diff
@@ -572,9 +574,21 @@ class Establishment {
                     })
                 ]);
 
+                // For serviceUsers merge any other data into resultset 
+                fetchResults.serviceUsers = establishmentServiceUserResults.map((suResult)=>{
+                    const serviceUser = serviceUsers.find(element => { return suResult.serviceUserId === element.id});
+                    if(suResult.other) {
```
wozzer72: Another way to achieve this is simply:

```
fetchResults.serviceUsers = establishmentServiceUserResults.map((suResult)=>{
   const serviceUser = serviceUsers.find(element => { return suResult.serviceUserId === element.id});
   serviceUser.other = suResult.other;     // whether null or with value
   return serviceUser;
}
```

wozzer72: But the result is the same, so I'll approved anyway.

wozzer72:
[(6) Change Request - Service Users "Other" Input field](https://trello.com/c/PgdL36yn/55-6-change-request-service-users-other-input-field)

PR #649, opened 2019-06-03
Feature | change request | Job Roles 'Other' Input field (#77)

alleypa:
- WIP
- Update Job model with `other` property
- Add JobRole model for save job role text
- update template to dynamically show other job text


alleypa:
Couple of unrelated commits. Hence I close and push fresh commits

PR #650, opened 2019-06-03
bugfix | registration flow | navigating back to select main service s…

omarCreativeDev:
…hould pre populate previous entries


omarCreativeDev:
[(1) Registration flow | Navigating back to select main service should pre populate previous entries](https://trello.com/c/uJwTQwqn/101-1-registration-flow-navigating-back-to-select-main-service-should-pre-populate-previous-entries)

PR #652, opened 2019-06-03
Feature/adding job roles

alleypa:
Feature | change request | Job Roles 'Other' Input field (#77)
    - Update Job model with `other` property
    - Add JobRole model for save job role text
    - update full record template to dynamically show other job text
    - update basic staff record page to dynamically show other job text
    - Reverting change to use worker response, now uses form values to dictate navigtation path
    - Update the job role page to show other text input for the selected job
    - Add accessibility hidden option to for select option

### DIFF
--- a/proxy.js
+++ b/proxy.js
@@ -10,6 +10,23 @@ var app = express();
 
 const apiEndpoint = process.env.API_PROXY || 'https://sfcdev.cloudapps.digital';
 
+app.get('/api/jobs', function(req, res) {
+  res.send({
+    jobs: [
+      { "id": 15, "title": "Middle Management" },
+      { "id": 16, "title": "Nursing Assistant" },
+      { "id": 17, "title": "Nursing Associate" },
+      { "id": 18, "title": "Occupational Therapist" },
+      { "id": 19, "title": "Occupational Therapist Assistant", other: true },
+      {
+        "id": 20,
+        "title": "Other job roles directly involved in providing care",
+        other: true,
+      },
+    ]
+  })
+});
+
 app.get('/api/mock', function(req, res) {
     console.log("Hitting endpoint /api/mock")
     res.send({
--- a/proxy.js
+++ b/proxy.js
@@ -10,23 +10,6 @@ var app = express();
 
 const apiEndpoint = process.env.API_PROXY || 'https://sfcdev.cloudapps.digital';
 
-app.get('/api/jobs', function(req, res) {
-  res.send({
-    jobs: [
-      { "id": 15, "title": "Middle Management" },
-      { "id": 16, "title": "Nursing Assistant" },
-      { "id": 17, "title": "Nursing Associate" },
-      { "id": 18, "title": "Occupational Therapist" },
-      { "id": 19, "title": "Occupational Therapist Assistant", other: true },
-      {
-        "id": 20,
-        "title": "Other job roles directly involved in providing care",
-        other: true,
-      },
-    ]
-  })
-});
-
 app.get('/api/mock', function(req, res) {
     console.log("Hitting endpoint /api/mock")
     res.send({
--- a/src/app/core/model/job.model.ts
+++ b/src/app/core/model/job.model.ts
@@ -2,4 +2,11 @@ export interface Job {
   jobId: number;
   id?: number;
   title?: string;
+  other?: boolean;
+}
+
+export interface JobRole {
+  jobId: number;
+  title?: string;
+  other?: string;
 }
--- a/src/app/core/model/job.model.ts
+++ b/src/app/core/model/job.model.ts
@@ -1,6 +1,5 @@
 export interface Job {
-  jobId: number;
-  id?: number;
+  id: number;
   title?: string;
   other?: boolean;
 }
--- a/src/app/features/workers/create-basic-records/create-basic-records.component.ts
+++ b/src/app/features/workers/create-basic-records/create-basic-records.component.ts
@@ -123,7 +123,9 @@ export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
     this.closeStaffRecords();
     this.staffRecordsControl.controls[index].patchValue({ active: true });
     setTimeout(() => {
-      this.elementRef.nativeElement.querySelector(`.staff-record:nth-of-type(${index + 1})`).scrollIntoView(true);
+      this.elementRef.nativeElement
+        .querySelector(`.govuk-summary-list__row:nth-of-type(${index + 1})`)
+        .scrollIntoView(true);
     });
   }
 
@@ -150,7 +152,9 @@ export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
     });
 
     if (this.form.valid || this.staffRecordsControl.length === 0) {
-      this.workerService.setCreateStaffResponse(this.staffRecordsControl.length);
+      this.workerService.setCreateStaffResponse(
+        this.staffRecordsControl.controls.filter(record => record.get('uid').value).length
+      );
       this.router.navigate(['/worker', 'basic-records-save-success']);
     } else {
       const unsavedIndex = this.staffRecordsControl.controls.findIndex(control => {
--- a/src/app/features/workers/create-basic-records/create-basic-records.component.ts
+++ b/src/app/features/workers/create-basic-records/create-basic-records.component.ts
@@ -16,12 +16,15 @@ import { isNull } from 'util';
   templateUrl: './create-basic-records.component.html',
 })
 export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
+  private jobRoleCharacterLimit = 120;
+
   public contracts: Array<string> = [];
   public jobs: Job[] = [];
   public totalStaff: number;
   public totalWorkers = 0;
   public form: FormGroup;
   public submitted = false;
+  public showInputTextforOtherRole: boolean;
   private subscriptions: Subscription = new Subscription();
 
   constructor(
@@ -109,6 +112,7 @@ export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
     return this.formBuilder.group({
       nameOrId: [null, Validators.required],
       mainJobRole: [null, Validators.required],
+      jobRole: [null, [Validators.maxLength(this.jobRoleCharacterLimit)]],
       contract: [null, Validators.required],
       uid: null,
       active: true,
@@ -127,6 +131,14 @@ export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
     this.staffRecordsControl.controls.forEach(control => {
       control.patchValue({ active: false });
     });
+  }
+
+  onChangeJobRole(id: number) {
+    this.showInputTextforOtherRole = false;
+    const selectedjob = this.jobs.find(job => job.id === +id);
+    if (selectedjob.other) {
+      this.showInputTextforOtherRole = true;
+    }
   }
 
   submitHandler() {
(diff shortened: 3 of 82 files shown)


src/app/core/model/job.model.ts
```diff
@@ -2,4 +2,11 @@ export interface Job {
   jobId: number;
   id?: number;
   title?: string;
+  other?: boolean;
+}
+
+export interface JobRole {
```
omarCreativeDev: both these interfaces look very similar. Do we need both? If so can one extend the other?

src/app/features/workers/staff-details/staff-details.component.html
```diff
@@ -40,13 +40,38 @@ <h1 class="govuk-heading-xl">{{ worker ? 'Edit' : 'Create' }} Staff Details</h1>
       formControlName="mainJob"
       id="mainJob"
       name="mainJob"
-    >
+      (change)="onChangeJobRole($event.target.value)">
```
omarCreativeDev: not related to this pull request but this a reference to `serverError` in this template and it doesn't seem to be referenced in the component anywhere... could you check to see if we need it?

src/app/features/workers/staff-details/staff-details.component.html
```diff
@@ -40,13 +40,38 @@ <h1 class="govuk-heading-xl">{{ worker ? 'Edit' : 'Create' }} Staff Details</h1>
...
       <option *ngFor="let job of jobsAvailable" [value]="job.id">
         {{ job.title }}
       </option>
     </select>
   </div>
+
+  <div
+    class="govuk-form-group"
+    [class.govuk-form-group--error]="submitted && form.get('jobRole').invalid"
+    *ngIf="showInputTextforOtherRole">
```
omarCreativeDev: might be better for accessibility and screen readers to hide the content with a css class rather than an ngIf

src/app/features/workers/staff-details/staff-details.component.html
omarCreativeDev: update: might be worth checking if we/should be using the conditional content gds markup/css here?

alleypa: > not related to this pull request but this a reference to `serverError` in this template and it doesn't seem to be referenced in the component anywhere... could you check to see if we need it?

It is reference and used in the component of staff-details-component. Line 132

alleypa: > update: might be worth checking if we/should be using the conditional content gds markup/css here?

yes we should, I think we might have to extend one of the conditional ones you made. But in the interest of time `*ngIf` should just work

src/app/core/model/job.model.ts
alleypa: We do need both because we are now differentiating Job from JobRole. In the `get api` scenario the `other` property is a boolean other times when saving with/to the worker model it is a string. I think better to just keep it simple given the BE model

src/app/features/workers/staff-details/staff-details.component.html
alleypa: yes I agree.

dpanov76: Agree with Omar here

dpanov76: Everything else looks good for me

src/app/core/model/job.model.ts
omarCreativeDev: ok. but seeing as the only difference is the `id` property maybe `Job` can just be 
```
export interface Job extends JobRole {
   id?: number;
}
```

src/app/features/workers/staff-details/staff-details.component.html
omarCreativeDev: > > not related to this pull request but this a reference to `serverError` in this template and it doesn't seem to be referenced in the component anywhere... could you check to see if we need it?
> 
> It is reference and used in the component of staff-details-component. Line 132

I've double checked and line 132 of this component file doesn't reference `serverError`... could you check again please. but basically `serverError` doesn't seem to be referenced in this component or the parent component either

omarCreativeDev: > > update: might be worth checking if we/should be using the conditional content gds markup/css here?
> 
> yes we should, I think we might have to extend one of the conditional ones you made. But in the interest of time `*ngIf` should just work

I'd be keen to hear what @ChazUK thinks about this but we have an external auditor reviewing the application periodically and she is very strict on accessibility etc so this could potentially be flagged up. 

i think its better to add something like this to the utils.scss file 
```
.govuk-util__block {
  display: block;
}
```

src/app/core/model/job.model.ts
omarCreativeDev: update: oops i apologise, i didnt realise that one of the `other` properties was a boolean, in that case ignore my comments about this

proxy.js
```diff
@@ -10,6 +10,23 @@ var app = express();
 
 const apiEndpoint = process.env.API_PROXY || 'https://sfcdev.cloudapps.digital';
 
+app.get('/api/jobs', function(req, res) {
```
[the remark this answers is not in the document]
alleypa: yes. I will remember to remove. Just wanted you have a look first

src/app/features/workers/staff-details/staff-details.component.html
ChazUK: I think I put a comment below about the same thing, but I can't see the specific line you're referring to.

src/app/features/workers/other-job-roles/other-job-roles.component.html
```diff
@@ -12,26 +12,62 @@ <h1 class="govuk-heading-xl">Do they have any other job roles?</h1>
...
+            [class.govuk-checkboxes__conditional--hidden]="!job.value.checked"
+            *ngIf="showforOtherJobRole"
+            id="conditional-selectedJobRole-{{ idx }}">
+            <div class="govuk-form-group">
+              <label class="govuk-label" for="other-selectedJobRole-{{ idx }}">
+                Enter your job role
+              </label>
+              <input
+                class="govuk-input govuk-!-width-one-third"
+                id="other-selectedJobRole-{{ idx }}"
```
ChazUK: There's also no form control on this input field to capture it in the Reactive Form

src/app/core/model/job.model.ts
```diff
@@ -2,4 +2,11 @@ export interface Job {
   jobId: number;
   id?: number;
```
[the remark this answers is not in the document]
alleypa: done. We removing jobId instead

alleypa:
[(7) Change Request - Job Roles "Other" Input field](https://trello.com/c/hiy5ZcqL/77-7-change-request-job-roles-other-input-field)

src/app/features/workers/other-job-roles/other-job-roles.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component } from '@angular/core';
+import { Component, OnInit, AfterViewInit, DoCheck } from '@angular/core';
```
omarCreativeDev: `OnInit, AfterViewInit, DoCheck` don't seem to be used...

src/app/features/workers/other-job-roles/other-job-roles.component.ts
```diff
@@ -30,29 +39,83 @@ export class OtherJobRolesComponent extends QuestionComponent {
...
+          this.availableJobRoles = jobRoles.filter(j => j.id !== this.worker.mainJob.jobId);
+
+          // TODO: This does not really allow fall back for non-javascript form submissions
+          this.availableJobRoles.map(job => {
+            const otherJob = this.worker.otherJobs && this.worker.otherJobs.find(o => o.jobId === job.id);
+
+            if (job.other) {
+              this.jobsWithOtherRole.push({
+                jobId: job.id,
+                other: otherJob ? otherJob.other: ''
```
omarCreativeDev: can you add a space before the colon, this sort of thing gets autofixed by prettier by the way

src/app/features/workers/other-job-roles/other-job-roles.component.ts
```diff
@@ -30,29 +39,83 @@ export class OtherJobRolesComponent extends QuestionComponent {
...
+        const isJobWithRole = this.jobsWithOtherRole.some(jbRole => jbRole.jobId === j.jobId);
+        if (isJobWithRole) {
+          const otherValue = this.form.get(`otherSelectedJobRole${j.jobId}`).value;
+          return {
+            jobId: j.jobId,
+            ...(otherValue && {  other: otherValue })
+          };
+        }
+
+        return { jobId: j.jobId }
```
omarCreativeDev: can you add a trailing semi colon, this sort of thing gets autofixed by prettier by the way

src/app/features/workers/other-job-roles/other-job-roles.component.ts
```diff
@@ -30,29 +38,81 @@ export class OtherJobRolesComponent extends QuestionComponent {
...
+    this.formErrorsMap = [];
+  }
+
+  private updateForm(): void {
+    this.jobsWithOtherRole.forEach((job: JobRole) => {
+      this.form.addControl(
+        `otherSelectedJobRole${job.jobId}`,
+        new FormControl(job.other, {
+          validators: Validators.maxLength(this.otherJobRoleCharacterLimit),
+          updateOn: 'blur'
```
ChazUK: I'd remove this and make sure that the form field error classes are set when the form is submitted

src/app/features/workers/create-basic-records/create-basic-records.component.ts
```diff
@@ -16,13 +16,16 @@ import { isNull } from 'util';
...
 })
 export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
+
   public contracts: Array<string> = [];
   public jobs: Job[] = [];
   public totalStaff: number;
   public totalWorkers = 0;
   public form: FormGroup;
   public submitted = false;
+  public showInputTextforOtherRole: boolean;
```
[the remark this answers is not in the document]
alleypa: I think this will initially be hidden for all job roles but become true or visible only for job roles with other (true). Am I right on that

src/app/features/workers/other-job-roles/other-job-roles.component.ts
alleypa: So errors are already getting triggered when we submit form. This solution was only to optimize when user types into in the input field otherwise we are utilizing more resources than necessary. I can change the option to this `new FormControl(job.other, {
          validators: Validators.maxLength(this.otherJobRoleCharacterLimit),
          updateOn: 'submitted'`

src/app/features/workers/create-basic-records/create-basic-records.component.ts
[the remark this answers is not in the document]
ChazUK: With the current design I don't think it will cause any issues. My concern is when you are opening and closing individual staff records, will the input field be shown when the currently open staff record doesn't have an 'other' main job role because this flag is set globally whenever a main job role is changed. We can leave it for Ann to test

src/app/features/workers/other-job-roles/other-job-roles.component.ts
ChazUK: This looks pretty promising and might let us get rid of a load of code surrounding error display.
Let me look into this, can't believe I didn't know this existed!

ChazUK: Currently this is only available on `@angular/forms@7.1.0`, but investigating more, we can add it as tech debt

PR #653, opened 2019-06-03
Feature/other service other input field

omarCreativeDev:
- conditionally display other text input fields 
- updated payload, interface as per BE model
- corrected component property visibility by making it private, removed duplicate css class, corrected markup id and for attribute for better accessibility


omarCreativeDev:
[(6) Change Request - Other Service "Other" Input field](https://trello.com/c/bnfFCPM1/79-6-change-request-other-service-other-input-field)

PR #654, opened 2019-06-04
"other" added to otherServices and mainService

dbjsdev:


server/routes/establishments/worker.js
```diff
@@ -136,9 +136,11 @@ router.route('/').post(async (req, res) => {
             // note - req.username is assured, vecause it is provided through the
             //  hasAuthorisedEstablishment middleware which runs on all establishment routes
             await newWorker.save(req.username);
-            return res.status(201).json({
-                uid: newWorker.uid
-            });
+
+            const exceptionalOverrideHeader = req.headers['x-override-put-return-all'];
```
wozzer72: remove this - it's not relevant on a POST as there are no other properties to return.

server/routes/establishments/worker.js
```diff
@@ -136,9 +136,11 @@ router.route('/').post(async (req, res) => {
...
+            const showModifiedOnly = exceptionalOverrideHeader ? false : true;
+            return res.status(201).json(newWorker.toJSON(false, false, false, showModifiedOnly));
```
wozzer72: replace `showModifiedOnly` with `false` - for a POST we'll only show all the properties because they'll all be modified - it's just been created.

PR #658, opened 2019-06-05
bugfix | previous data entered into the other text fields isnt redisplayed

omarCreativeDev:
- added missing property to interface
- if `user has previously entered data in to the other text field then display it


omarCreativeDev:
[Previous data entered into the 'other' text fields isn't redisplayed when the user returns to the 'Other services' page](https://trello.com/c/aBAd2SuV/105-previous-data-entered-into-the-other-text-fields-isnt-redisplayed-when-the-user-returns-to-the-other-services-page)

PR #660, opened 2019-06-05
bugfix | previous data entered into the other text fields isnt redisplayed

omarCreativeDev:
- align FE model with updated BE model as requested

### DIFF
--- a/src/app/core/model/services.model.ts
+++ b/src/app/core/model/services.model.ts
@@ -2,8 +2,7 @@ export interface Service {
   id: number;
   isMyService?: boolean;
   name: string;
-  other?: boolean;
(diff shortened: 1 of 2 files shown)


omarCreativeDev:
[Previous data entered into the 'other' text fields isn't redisplayed when the user returns to the 'Other services' page](https://trello.com/c/aBAd2SuV/105-previous-data-entered-into-the-other-text-fields-isnt-redisplayed-when-the-user-returns-to-the-other-services-page)

omarCreativeDev:
closing pr as BE have suggested the other property can be a string or a boolean, i will wait till the BE is deployed before applying the fix

PR #668, opened 2019-06-07
Bugfix/confirm workplace only show either main service or other main service label

omarCreativeDev:
- conditionally display main selected service or other main service from text field if entered
- remove extra row displaying other main service from text field entry


omarCreativeDev:
[Registration: When 'Other' text is entered for a main service it should be displayed as the main service on the 'Confirm your workplace details' page](https://trello.com/c/Opkz8plj/110-registration-when-other-text-is-entered-for-a-main-service-it-should-be-displayed-as-the-main-service-on-the-confirm-your-workpl)

PR #669, opened 2019-06-07
Workplace - Type of Employer - Other Field

ChazUK:


ChazUK:
[(5) Change Request -Type of Employer "Other" Input field](https://trello.com/c/rh1Uucfk/81-5-change-request-type-of-employer-other-input-field)

ChazUK:
[Workplace Type of employer question - wording is incorrect](https://trello.com/c/bgNpnFqM/111-workplace-type-of-employer-question-wording-is-incorrect)

src/app/core/services/establishment.service.ts
```diff
@@ -157,7 +160,7 @@ export class EstablishmentService {
   }
 
   updateTypeOfEmployer(establishmentId, data) {
-    return this.http.post<EmployerTypeResponse>(`/api/establishment/${establishmentId}/employerType`, data);
+    return this.http.post<any>(`/api/establishment/${establishmentId}/employerType`, data);
```
omarCreativeDev: how comes you changed this to `any`?

src/app/features/workplace/type-of-employer/type-of-employer.component.html
```diff
@@ -2,28 +2,57 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
-    <h1 class="govuk-heading-xl">What type of employer is {{ employerName }}</h1>
+    <h1 class="govuk-heading-xl">What type of employer is {{ establishment.name }}?</h1>
```
omarCreativeDev: how comes this has changed from employerName?

src/app/features/workplace/type-of-employer/type-of-employer.component.ts
```diff
@@ -19,6 +19,7 @@ export class TypeOfEmployerComponent extends Question {
     'Voluntary / Charity',
     'Other',
   ];
+  public maxLength = 120;
```
omarCreativeDev: this could possibly go in a global constant as all of the other field questions have the same value

src/app/features/workplace/type-of-employer/type-of-employer.component.html
ChazUK: Changed this to use the establishment name of the current establishment that is being edited so that it is future proof'd for Parents & Subs. When editing a Sub Establishment you don't want the name for your own establishment being displayed

src/app/features/workplace/type-of-employer/type-of-employer.component.ts
ChazUK: Yeah I was thinking about that, but maybe it's best to leave it separate in case something random changes.

src/app/core/services/establishment.service.ts
ChazUK: Will double check, but I think I changed it as there was a problem casting the properties to the specific response type

omarCreativeDev: would be good to have a type here but something we can pick up later

ChazUK: I've added a ticket as tech debt to fix these Interfaces, they are very messy and scattered across a lot of different files, and named pretty badly.

PR #679, opened 2019-06-10
Feature | bulk upload | validate, pass no warnings

omarCreativeDev:
- FE work encapsulate happy path of bulk upload validate process with no errors

### DIFF
--- a/src/app/core/model/bulk-upload.model.ts
+++ b/src/app/core/model/bulk-upload.model.ts
@@ -9,8 +9,12 @@ export interface PresignedUrlResponse {
 }
 
 export interface UploadFile extends File {
+  errors?: number;
   extension: string;
+  fileType?: string;
+  records?: number;
   status?: FileValidateStatus;
+  warnings?: number;
 }
 
 export interface ValidatedFilesResponse {
@@ -22,6 +26,7 @@ export interface ValidatedFilesResponse {
 export interface ValidatedFile {
   errors: number;
   filename: string;
+  fileType?: string;
   records: number;
   warnings: number;
 }
--- a/src/app/core/model/bulk-upload.model.ts
+++ b/src/app/core/model/bulk-upload.model.ts
@@ -18,7 +18,7 @@ export interface UploadFile extends File {
 }
 
 export interface ValidatedFilesResponse {
-  establishments: ValidatedFile;
+  establishment: ValidatedFile;
   training: ValidatedFile;
   workers: ValidatedFile;
 }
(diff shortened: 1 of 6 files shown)


omarCreativeDev:
[(1) Bulk Upload - Passed Validation with No Warnings](https://trello.com/c/B7rG0kit/69-1-bulk-upload-passed-validation-with-no-warnings)

src/app/core/model/bulk-upload.model.ts
```diff
@@ -9,19 +9,24 @@ export interface PresignedUrlResponse {
 }
 
 export interface UploadFile extends File {
```
ChazUK: I think we should make the distinction between files that are too be Uploaded, and files that have been uploaded.

src/app/core/model/bulk-upload.model.ts
```diff
@@ -9,19 +9,24 @@ export interface PresignedUrlResponse {
...
 }
 
 export interface ValidatedFilesResponse {
-  establishments: ValidatedFile;
+  establishment: ValidatedFile;
   training: ValidatedFile;
   workers: ValidatedFile;
 }
 
 export interface ValidatedFile {
```
ChazUK: As stated above it's probably better to have make ValidatedFiles contain the errors filetype etc that is returned from the API rather than extending File, as we won't be using anything from that interface

src/app/features/bulk-upload/uploaded-files-list/uploaded-files-list.component.html
```diff
@@ -11,8 +11,8 @@
   <tbody class="govuk-table__body">
     <tr class="govuk-table__row" *ngFor="let file of uploadedFiles">
       <td class="govuk-table__cell" scope="row">{{ file.name }}</td>
-      <td class="govuk-table__cell">{{ file.extension }}</td>
-      <td class="govuk-table__cell">-</td>
+      <td class="govuk-table__cell">{{ file.fileType ? file.fileType : file.extension }}</td>
```
ChazUK: Same sort of comment as before, we shouldn't be using the file.extension as that's part of the File interface and not UploadedFiles.

ChazUK: Looking at it again, I think this comes down to my query with displaying the file extension and the file "type" (Staff Records, Workplace etc) seems strange that it's flipping between those 2 in the same column as they have different meanings.

PR #682, opened 2019-06-11
Bug | Safari File Upload

ChazUK:
Adding ControlValueAccessor to ensure that onchange events are fired for File Inputs in Safari


ChazUK:
[Bulk Upload - File Uploading does not work in Safari](https://trello.com/c/DunzLeSL/78-bulk-upload-file-uploading-does-not-work-in-safari)

PR #688, opened 2019-06-11
GOV.UK Framework | Workplace | Service Users

ChazUK:
Updating Service Users to use the GOV.UK Design Framework
Adding Other fields to Service Users

The names of the interfaces are very confusing and I'm not 100% sure what to call them.

### DIFF
--- a/src/app/features/workplace/service-users/service-users.component.ts
+++ b/src/app/features/workplace/service-users/service-users.component.ts
@@ -63,7 +63,7 @@ export class ServiceUsersComponent extends Question {
     filter(this.allUserServices, { other: true }).forEach((service: ServiceForUser) => {
       this.form.addControl(
         `serviceUsers-${service.id}-otherService`,
-        new FormControl(null, { validators: [Validators.maxLength(this.otherMaxLength)], updateOn: 'submit' })
+        new FormControl(null, [Validators.maxLength(this.otherMaxLength)])
       );
 
       this.formErrorsMap.push({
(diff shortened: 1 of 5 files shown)


ChazUK:
[Selected service users are not being saved](https://trello.com/c/7GkAie88/73-selected-service-users-are-not-being-saved)

ChazUK:
[(3) Workplace - Service Users - Accessibility](https://trello.com/c/fz56wGVn/74-3-workplace-service-users-accessibility)

ChazUK:
[(6) Change Request - Service Users "Other" Input field](https://trello.com/c/PgdL36yn/72-6-change-request-service-users-other-input-field)

src/app/features/workplace/service-users/service-users.component.ts
```diff
@@ -59,6 +59,34 @@ export class ServiceUsersComponent extends Question {
     );
   }
 
+  private addFormControls(): void {
+    filter(this.allUserServices, { other: true }).forEach((service: ServiceForUser) => {
+      this.form.addControl(
+        `serviceUsers-${service.id}-otherService`,
+        new FormControl(null, { validators: [Validators.maxLength(this.otherMaxLength)], updateOn: 'submit' })
```
omarCreativeDev: as discussed having this could potentially mean that if user enters text longer than the limit and submits the form they are presented with errors... but then if they make the text a valid length the errors will stay remain on the screen which could be confusing for the user...

omarCreativeDev:
i've left 1 comment, and note there are some conflicts on `services.model.ts`. otherwise pr is looking good

src/app/features/workplace/service-users/service-users.component.ts
ChazUK: Sorry didn't realise I left that in!

ChazUK: removed

PR #691, opened 2019-06-11
Bug | Fixed typo on GET Available Qualifications endpoint

ChazUK:


ChazUK:
[(2) Defect - Qualifications - cannot add a qualification - Name box is truncated and only shows 'Select' - Front End](https://trello.com/c/FiBeZacl/10-2-defect-qualifications-cannot-add-a-qualification-name-box-is-truncated-and-only-shows-select-front-end)

PR #693, opened 2019-06-12
allow empty string for other job role other

AndrewKeig:


AndrewKeig:
[When amending other job roles, cannot 'save and return' until text is entered](https://trello.com/c/MVjE9e6k/89-when-amending-other-job-roles-cannot-save-and-return-until-text-is-entered)

PR #694, opened 2019-06-12
Accessibility failure #24

alleypa:
- Add aria accessibility and id generated to support assisted technology
screen reader usage


alleypa:
code changes now merged with main accessibility issues branch

PR #695, opened 2019-06-12
bugfix | only prefill form if user has previously checked other services

omarCreativeDev:
bugfix | only prefill form if user has previously checked other services as BE omits the otherServices property from get establishment call altogether rather than sending back an empty array


omarCreativeDev:
[No other services listed for selection for a new establishment](https://trello.com/c/eKpxAYub/93-no-services-listed-for-selected-for-a-new-establishment)

PR #696, opened 2019-06-12
GOV.UK Framework | Workplace | Confirm Vacancies

ChazUK:
Extended the Panel SCSS to include an interruption card https://github.com/alphagov/govuk-design-system-backlog/issues/27
Added Definition List SCSS to display block of 3 list items, using CSS Grids, making sure it's prefixed for IE.

Updating Confirm Vacancies to use the GOV.UK Design Framework

### DIFF
--- a/src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.html
+++ b/src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.html
@@ -2,14 +2,14 @@
   <h1 class="govuk-panel__title">Current staff vacancies</h1>
   <div class="govuk-panel__body">
     <p class="govuk-heading-m">You told us you have:</p>
-    <dl class="govuk-list--definition govuk-!-margin-bottom-8">
-      <dt>current staff vacancies</dt>
+    <dl class="govuk-list--definition govuk-list--definition-xl govuk-!-margin-bottom-8">
+      <dt>Current staff vacancies</dt>
       <dd>{{ establishment.TotalVacencies }}</dd>
     </dl>
     <p>Continue if this is correct or go back and make a change.</p>
     <div class="govuk-button-group">
       <a
-        class="govuk-button govuk-!-margin-right-8"
+        class="govuk-button govuk-!-margin-right-8 govuk-!-margin-bottom-0"
         role="button"
         [routerLink]="['/workplace', this.establishment.id, 'starters']"
       >
@@ -17,7 +17,7 @@
       </a>
       <span class="govuk-visually-hidden">or</span>
       <a
-        class="govuk-button govuk-button--link"
+        class="govuk-button govuk-button--link govuk-!-margin-bottom-0"
         role="button"
         [routerLink]="['/workplace', this.establishment.id, 'vacancies']"
       >
--- a/src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.html
+++ b/src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.html
@@ -1,16 +1,27 @@
-<div class="main-page-container confirmation-page col-12">
-  <h1 class="h1">Current staff vacancies</h1>
-
-  <h2>You told us you have:</h2>
-  <div class="facts">
-    <span class="amount">{{total}}</span>
-    <span class="label">current staff vacancies</span>
+<div class="govuk-panel govuk-panel--interruption">
+  <h1 class="govuk-panel__title">Current staff vacancies</h1>
+  <div class="govuk-panel__body">
+    <h2 class="govuk-heading-m">You told us you have:</h2>
+    <dl class="govuk-list--definition govuk-!-margin-bottom-8">
+      <dt>current staff vacancies</dt>
+      <dd>{{ establishment.TotalVacencies }}</dd>
+    </dl>
+    <p>Continue if this is correct or go back and make a change.</p>
+    <div class="govuk-button-group">
+      <a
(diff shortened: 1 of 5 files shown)


ChazUK:
[(3) Workplace - Confirm Vacancies](https://trello.com/c/tHAY39Ww/9-3-workplace-confirm-vacancies)

src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.html
```diff
@@ -1,16 +1,28 @@
...
-  <div class="facts">
-    <span class="amount">{{total}}</span>
-    <span class="label">current staff vacancies</span>
+<div class="govuk-panel govuk-panel--interruption">
+  <h1 class="govuk-panel__title">Current staff vacancies</h1>
+  <div class="govuk-panel__body">
+    <p class="govuk-heading-m">You told us you have:</p>
+    <dl class="govuk-list--definition govuk-list--definition-xl govuk-!-margin-bottom-8">
+      <dt>Current staff vacancies</dt>
+      <dd>{{ establishment.TotalVacencies }}</dd>
```
omarCreativeDev: I assume `TotalVacencies` property is coming back from the BE like this? weird casing and spelling...

ChazUK: yep :(

src/assets/scss/components/_panel.scss
```diff
@@ -0,0 +1,41 @@
+.govuk-panel--interruption {
+  background-color: govuk-colour('blue');
+  text-align: left;
+
+  * {
```
omarCreativeDev: is the wildcard selector needed here? if so do we need the color's applied to the anchor states?

src/assets/scss/components/_panel.scss
```diff
@@ -0,0 +1,41 @@
...
+    color: #fff;
+  }
+
+  a {
```
omarCreativeDev: could probs combine like 
```
a, a:link, a:visited { color: #fff; )
```
also would be good to use variables for colors `src/assets/scss/partials/_colors.scss`

src/assets/scss/components/_panel.scss
ChazUK: yeah, if the wildcard isn't there we'd have to set the colour for every element as it's usually the default black.

src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.ts
```diff
@@ -1,37 +1,28 @@
...
+  constructor(private establishmentService: EstablishmentService) {}
 
   ngOnInit() {
-    this.subscriptions.push(
-      this.establishmentService
-        .getJobs()
-        .pipe(map(jobs => jobs.TotalVacencies))
-        .subscribe(total => (this.total = total))
+    this.subscriptions.add(
+      this.establishmentService.establishment$.pipe(take(1)).subscribe(establishment => {
```
omarCreativeDev: do we need to subscribe to to this observable? the workplace module has the `Auth` guard set so to be able to access this route the user has to be logged in and therefore the establishment would be already set on the `EstablishmentService` anyway? we could probs just do something like `this.establishment = this.establishmentService.establishment;`....?

src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.ts
```diff
@@ -1,37 +1,28 @@
...
+        this.establishment = establishment;
+      })
     );
   }
 
   ngOnDestroy() {
-    this.subscriptions.forEach(s => s.unsubscribe());
+    this.subscriptions.unsubscribe();
```
omarCreativeDev: I dont think you need to unsubscribe if you do a `take(1)`

src/assets/scss/components/_panel.scss
ChazUK: I think this is going to be a temporary fix until they bring in the Interruption Card properly in GDS https://github.com/alphagov/govuk-design-system-backlog/issues/27

src/assets/scss/components/_panel.scss
ChazUK: We should probably extend the `govuk-colour` map to contain these new colours and then use that instead of just the colour variable, will update though

ChazUK: Colours updated and rules grouped

src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.ts
ChazUK: The reason I've put the unsubscribe here is if the person navigates away from the page we want to cancel that request, very unlikely, but just good practice I think. And per your previous comment I think this can probably be removed too.

src/app/features/workplace/confirm-vacancies/confirm-vacancies.component.ts
ChazUK: After taking another look at this we should leave it in. Comes down to the state management RXJS Pattern and best practices, you should access the state directly, and instead subscribe to the observable. This would definitely be clearer with NGXS/NGRX as there's a set way to pull data from the state. I do the same thing on the Question pages, which this is a part of.

ChazUK: Btw, there isn't any extra HTTP request that is being made for this page.

PR #697, opened 2019-06-12
Accessibility issue voice input

alleypa:
- Add view input detection to force data entry


ChazUK:
Would be good to get an overview of what this aims to fix and how, looking forward to tomorrow!

alleypa:
this was a proof of concept. More investigation to be done

PR #699, opened 2019-06-12
Bulk upload validations merge with api

wozzer72:
This is a fairly large PR for review against https://trello.com/c/FUtQEhVR.

It covers extending the bulk upload specific validation to incorporate the entity property framework validation.

To do this I have introduced the `EntityValidator` base class to all four entities: Establishment, Worker, Qualification and Training, which intercepts the existing `hasMandatoryProperties` and `isValid` entity validators.

Establishments/Workers are based on Change Properties. Training/Qualifications are based on internal properties. Consequently, the way they valid is slightly different.

I've had to reduce the severity in response for both Training and Qualification when there is a validation error;l before it raised an exception, now it returns false (aligned with Workers/Establishment).

I also had to fix an error I introduced to `Establishment::toJSON` in respect to the created/updated/updatedBy properties; this is bombing the integration tests.

With the entities now raising `ValidationMessages`, `bulkupload.js` validation functions are now checking for entity validity before caching the entity, and if not valid, merging the validation errors and warnings, with those raised during CSV parsing/validation - including mapping the line number of the associated CSV line validator.

Finally, the `[POST] validate` has been updated to filter between errors and warning and sort on line number.

If you want to gander at the resulting validation intermediary data files, have a look here: https://s3.console.aws.amazon.com/s3/buckets/sfcbulkuploadfiles/479/validation/?region=eu-west-1&tab=overview.

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -174,6 +174,8 @@ class Establishment extends EntityValidator {
     // Returns true if the resulting Establishment is valid; otherwise false
     async load(document) {
         try {
+            this.resetValidations();
+
             await this._properties.restore(document, JSON_DOCUMENT_TYPE);
 
         } catch (err) {
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -11,6 +11,8 @@ const uuid = require('uuid');
 // database models
 const models = require('../index');
 
+const EntityValidator = require('./validations/entityValidator').EntityValidator;
+
 // notifications
 
 // temp formatters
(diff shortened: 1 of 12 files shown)


wozzer72:
[(5) Bulk Upload - Validate Files - Back End](https://trello.com/c/FUtQEhVR/58-5-bulk-upload-validate-files-back-end)

wozzer72:
[(5) Bulk Upload - Validate Files - Back End](https://trello.com/c/FUtQEhVR/58-5-bulk-upload-validate-files-back-end)

wozzer72:
[(5) Bulk Upload - Validate Files - Back End](https://trello.com/c/FUtQEhVR/58-5-bulk-upload-validate-files-back-end)

server/models/classes/establishment.js
```diff
@@ -27,8 +30,10 @@ const SEQUELIZE_DOCUMENT_TYPE = require('./user/userProperties').SEQUELIZE_DOCUM
 // WDF Calculator
 const WdfCalculator = require('./wdfCalculator').WdfCalculator;
 
-class Establishment {
+class Establishment extends EntityValidator {
```
AndrewKeig: Nice, I was going to suggest this.

PR #700, opened 2019-06-12
Bug | Service Users error for new users

ChazUK:
Adding check to see if Establishment has ServiceUsers before trying to prefill form


ChazUK:
[No service Users listed for selection for a new establishment](https://trello.com/c/9HMQZ293/101-no-service-users-listed-for-selection-for-a-new-establishment)

PR #705, opened 2019-06-13
GOV.UK Framework | Workplace | Confirm Starters

ChazUK:


src/app/features/workplace/confirm-starters/confirm-starters.component.ts
```diff
@@ -1,37 +1,28 @@
...
-  }
+  constructor(private establishmentService: EstablishmentService) {}
 
   ngOnInit() {
-    this.subscriptions.push(
-      this.establishmentService
-        .getJobs()
-        .pipe(map(jobs => jobs.TotalStarters))
-        .subscribe(total => (this.total = total))
+    this.subscriptions.add(
```
tgreyuk: I am not convinced about the use case of this. If the user navigated from the page then by definition they have already entered the ngOnInit lifecycle event and the take(1) has already destroyed the subsription.

PR #712, opened 2019-06-14
Accessibility | Total Staff Panel now using Definition List

ChazUK:
Updated the Total Staff Panel on the Staff Records Tab to use the definition list rather than divs for accessibility.

Tested and works on IE10/11, Edge and all modern browsers


ChazUK:
[Accessibility - Total Staff Panel](https://trello.com/c/17Hwv7Hg/111-accessibility-total-staff-panel)

PR #715, opened 2019-06-14
Fix input other field should only show for other job role

alleypa:


alleypa:
[Other text field should not appear until appropriate Main job role has been selected](https://trello.com/c/r5Di3Nx5/87-other-text-field-should-not-appear-until-appropriate-main-job-role-has-been-selected)

PR #717, opened 2019-06-15
Aws secrets manager

wozzer72:
https://trello.com/c/OxdRoU55

Incorporates AWS Secret Manager with convict configuration, to allow overriding of sensitive information.

Introduces AppConfig emitter, to broadcast when the convict configuration had been updated following successful initialisation from AWS Secret Manager.

The express server now waits for the config ready event.

The sequelize configuration also now rebind on the config ready event, thus restoring connection pool once sensitive properties are made available, namely, database host, password and client SSL certs.

Sequelize model now broadcasts when it is ready, which is then consumed by BUDI singleton, to ensure it does try to initialise with "not yet ready" sequelize parameters.

And finally, had to refactor the way client SSL certs/keys are presented to sequelize, differentiating between files (if running on localhost) to pure data if running from AWS Secret Manager.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 # IDE - VSCode
 .vscode/*
 !.vscode/settings.json
+!.vscode/launch.json
 !.vscode/tasks.json
 !.vscode/extensions.json
 .vs/*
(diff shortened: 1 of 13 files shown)


wozzer72:
[(8) Step 1 - Environment Based Configuration to be held in AWS Secret Manager](https://trello.com/c/OxdRoU55/35-8-step-1-environment-based-configuration-to-be-held-in-aws-secret-manager)

PR #721, opened 2019-06-16
Feature | gds accessibility | workplace starters

omarCreativeDev:
- Update to use GOV.UK Framework
- Update Error messages
- Fix any bugs found on the page, validation etc.


omarCreativeDev:
[(5) Workplace - Starters - Accessibility](https://trello.com/c/twpTct6f/3-5-workplace-starters-accessibility)

src/app/core/model/establishment.model.ts
```diff
@@ -100,3 +100,27 @@ export interface Establishment {
...
+  updated: string;
+  updatedBy: string;
+  vacancies: string;
+}
+
+export interface PostStartersRequest {
+  starters: string | Starter[];
+}
+
+export enum NoRecordsReason {
```
ChazUK: I don't think we need to export this ENUM, I know it's being used across 3 different pages, but we also have other places where we have Yes, No, Don't Know and I don't think that has been pulled out.

src/app/core/services/job.service.ts
```diff
@@ -8,7 +10,7 @@ import { map } from 'rxjs/operators';
 export class JobService {
   constructor(private http: HttpClient) {}
 
-  getJobs() {
-    return this.http.get<any>('/api/jobs').pipe(map(res => res.jobs));
+  public getJobs(): Observable<Job[]> {
+    return this.http.get<GetJobsResponse>('/api/jobs').pipe(map((res: GetJobsResponse) => res.jobs));
```
ChazUK: No need to type res here as it's inherited from the HTTP get typing

src/app/core/services/establishment.service.ts
```diff
@@ -105,12 +105,14 @@ export class EstablishmentService {
     return this.http.post<any>(`/api/establishment/${this.establishmentId}/jobs`, data);
   }
 
-  getStarters() {
-    return this.http.get<any>(`/api/establishment/${this.establishmentId}/jobs`).pipe(map(res => res.starters));
+  public getStarters(): Observable<string | Starter[]> {
```
ChazUK: This doesn't need to be used as we can grab this info from the full Establishment object.

src/app/core/services/establishment.service.ts
```diff
@@ -105,12 +105,14 @@ export class EstablishmentService {
...
+    return this.http
+      .get<GetStartersResponse>(`/api/establishment/${this.establishmentId}/jobs`)
+      .pipe(map((res: GetStartersResponse) => res.starters));
   }
 
-  postStarters(data) {
-    return this.http.post<any>(`/api/establishment/${this.establishmentId}/jobs`, data);
+  public postStarters(data: PostStartersRequest): Observable<any> {
```
ChazUK: I've updated this to be `updateJobs` as it can be used for Starters, Leavers and Vacancies

src/app/core/model/establishment.model.ts
```diff
@@ -100,3 +100,27 @@ export interface Establishment {
   totalLeavers: number;
   wdf: WDF;
 }
+
+export interface GetStartersResponse {
```
ChazUK: Although we're deleting the API endpoint call this is typing, this interface could be cleaned up as it's using a Base Establishment interface, and most of these properties are optional.

src/app/shared/validators/custom-form-validators.ts
```diff
@@ -24,6 +24,26 @@ export class CustomValidators extends Validators {
     }
   }
 
+  static bothControlsHaveValues(group: AbstractControl): { [key: string]: boolean } | null {
```
ChazUK: I like the idea of this Custom Validator, but I think it's not serving it's purpose here as we discussed. We should look at other places where we want 2 required fields and how that needs to be handled, otherwise each field can be handled individually with `Validators.required` on each

src/app/shared/components/error-summary/error-summary.component.ts
```diff
@@ -50,6 +51,12 @@ export class ErrorSummaryComponent implements OnInit, OnDestroy {
 
         const formGroupControls: AbstractControl = formGroup['controls'];
         Object.keys(formGroupControls).forEach(i => this.collectError(formGroupControls[i], `${key}.${i}`));
+      } else if (isFormArray) {
```
ChazUK: I also made a change similar to this, my section is omitting any FormGroup validations, and yours is omitting any FormControl validations within the FormGroup so I'm going to merge them together.

PR #730, opened 2019-06-17
GOV.UK Framework | Workplace | Vacancies

ChazUK:
* Updating Establishment Model to properly describe Starters, Leavers and Vacancies property to be a String or an Array (not ideal)
* Workplace Question Class now stays subscribed to the Establishment Observable to ensure it is updated after a POST is made
* Error Summary Service now handles FormArrays, although we will probably want to look at doing this in another way so that we can reference each individual array entry.

There's a few other tweaks here which I included because they were so small.

* Country of Birth and Nationality Staff Record Questions were not able to progress if the user didn't enter a value
* Other Qualifications Staff Record Question was sending `false` instead of `null`


src/app/core/services/establishment.service.ts
```diff
@@ -157,4 +153,8 @@ export class EstablishmentService {
   updateLocalAuthorities(establishmentId, data) {
     return this.http.post<any>(`/api/establishment/${establishmentId}/localAuthorities`, data);
   }
+
+  updateVacancies(establishmentId, data) {
```
omarCreativeDev: is it possible to type this method, its return type and http post type?

src/app/features/workers/country-of-birth/country-of-birth.component.ts
```diff
@@ -97,7 +97,7 @@ export class CountryOfBirthComponent extends QuestionComponent {
 
   onSuccess() {
     this.next =
-      this.worker.countryOfBirth.value === 'United Kingdom'
+      this.worker.countryOfBirth && this.worker.countryOfBirth.value === 'United Kingdom'
```
omarCreativeDev: Can we put `United Kingdom' in to an Enum?

src/app/features/workers/nationality/nationality.component.ts
```diff
@@ -99,7 +99,7 @@ export class NationalityComponent extends QuestionComponent {
 
   onSuccess() {
     this.next =
-      this.worker.nationality.value === 'British'
+      this.worker.nationality && this.worker.nationality.value === 'British'
```
omarCreativeDev: Can we put `nationality' in to an Enum?

src/app/features/workplace/question/question.component.ts
```diff
@@ -87,6 +89,11 @@ export class Question implements OnInit, OnDestroy {
     this.updateEstablishment(props);
   }
 
+  protected _init(): void {
```
omarCreativeDev: can i ask what was the purpose of this method? Surely the OnInit() hook in the Question component is a good indication of it be initiated?

src/app/features/workplace/vacancies/vacancies.component.ts
```diff
@@ -1,199 +1,223 @@
...
+  public vacanciesKnownOptions = [
     {
       label: 'There are no current staff vacancies.',
-      value: 'no-staff',
+      value: vacancyOptions.NONE,
     },
     {
       label: `I don't know how many current staff vacancies there are.`,
-      value: 'dont-know',
+      value: vacancyOptions.DONTKNOW,
```
omarCreativeDev: can we put an underscore to seperate the words? other enums across the app seem to follow this pattern

src/app/features/workplace/vacancies/vacancies.component.ts
```diff
@@ -1,199 +1,223 @@
...
 
-    vacancyControl.push(this.createVacancyControlItem());
-  }
+    this.subscriptions.add(
+      this.form.get('vacanciesKnown').valueChanges.subscribe(value => {
+        while (this.vacanciesControl.length > 1) {
+          this.vacanciesControl.removeAt(1);
+        }
+        this.vacanciesControl.controls[0].get('jobRole').clearValidators();
+        this.vacanciesControl.controls[0].get('total').clearValidators();
```
omarCreativeDev: could these two lines be simplified from;
```
 this.vacanciesControl.controls[0].get('jobRole').clearValidators();
 this.vacanciesControl.controls[0].get('total').clearValidators();
```

to
```
 this.vacanciesControl.controls[0].clearValidators();
```
?

src/app/features/workplace/vacancies/vacancies.component.ts
```diff
@@ -1,199 +1,223 @@
...
-          vacancyControl.push(this.createVacancyControlItem());
-        }
-      })
+  private createVacancyControl(jobId = null, total = null): FormGroup {
+    return this.formBuilder.group(
+      {
+        jobRole: [jobId, [Validators.required]],
+        total: [total, [Validators.min(this.minVacancies), Validators.max(this.maxVacancies)]],
+      },
+      { validators: Validators.required }
```
omarCreativeDev: does having a out of the box required validator set on a group with multiple child controls ensure the child controls both cant be empty?

src/app/features/workers/country-of-birth/country-of-birth.component.ts
ChazUK: The Country of Birth data comes from an API, so not sure what benefit it'll be adding this 1 property to an enum.

src/app/features/workers/nationality/nationality.component.ts
ChazUK: The Nationality data comes from an API, so not sure what benefit it'll be adding this 1 property to an enum.

src/app/features/workplace/question/question.component.ts
ChazUK: `init()` is fired whenever the establishment observable is updated, we want to keep the establishment variable updated with the observable value, but we don't want to call `init()` multiple times, I call a (sort of) private method so that I can set `initiated = true` so next time establishment observable fires `init()` isn't called again

src/app/features/workplace/vacancies/vacancies.component.ts
ChazUK: updated

src/app/features/workplace/vacancies/vacancies.component.ts
ChazUK: Removed as it was left after trying to test if that would work (it doesn't 😒)

src/app/features/workplace/vacancies/vacancies.component.ts
ChazUK: That will set clear the Validators on the FormGroup and not the individual controls inside that form group. But I did spot that I missed a `Validators.required` on total 👍

omarCreativeDev: ah yes you are correct, thanks for explaining

src/app/core/services/establishment.service.ts
ChazUK: Updated to include typing

ChazUK:
[(5) Workplace - Vacancies - Accessibility](https://trello.com/c/qsXFk8tt/5-5-workplace-vacancies-accessibility)

PR #733, opened 2019-06-18
Feature/bulk upload failed validation ui

tgreyuk:
* This PR implements some complex UI for highlighting error in a table row
* I could not find any pattern like this on GDS guidelines
* Therefore i created a new tables sass to customise for this pattern

### DIFF
--- a/src/assets/scss/components/_tables.scss
+++ b/src/assets/scss/components/_tables.scss
@@ -3,20 +3,20 @@ $error_row_indent: 12px;
 $error_row_offset: 50px;
 $error_row_padding: $govuk-gutter-half;
 
-.govuk-table__error {
+.govuk-error-table {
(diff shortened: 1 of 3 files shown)


tgreyuk:
[(3) Bulk Upload - Failed Validation](https://trello.com/c/ZvB9OwD3/61-3-bulk-upload-failed-validation)

src/assets/scss/components/_tables.scss
```diff
@@ -0,0 +1,62 @@
+$error_border_size: 5px;
```
ChazUK: You might be able to pull these in from the gov.uk framework. Check the form-group error's and see if there's any global variables set.

tgreyuk: Will have a look - thanks

PR #737, opened 2019-06-18
Hotfix aws secrets in staging

wozzer72:
This is to deploy the test.yaml hotfix change in to `develop` too, so subsequent deployments to staging include it.


wozzer72:
Fixing the test.yaml changes under a separate PR

PR #738, opened 2019-06-18
Feature/username validation

AndrewKeig:


server/routes/registration.js
```diff
@@ -283,18 +287,26 @@ router.route('/')
 			});
     }
 
+    req.body[0].user.username
```
dbjsdev: Is this a stray?

AndrewKeig: that was a test :)

PR #739, opened 2019-06-18
Bulk upload validation report config hotfix

wozzer72:
Hi Nasir

With the introduction of AWS Secrets Manager, I rejigged the config.js schema which include renaming the `bulkupload` root name.

At the time, I didn't have your validation report endpoint.

This PR fixes the report endpoint to use the new config name.

Also included in this PR are changes to `test.yaml` we had to do to get the deployment to staging working.

Thanks

### DIFF
--- a/server/config/test.yaml
+++ b/server/config/test.yaml
@@ -8,7 +8,6 @@ db:
     pool: 5
     ssl: true
     username: sfcapp
-    database: sfctstdb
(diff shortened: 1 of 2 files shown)


wozzer72:
This is also to test the sfcdev deployment now we're removing the non-SSL entry in postgres `pg_hba.conf`

PR #740, opened 2019-06-18
bugfix | signout signin redirect issue

omarCreativeDev:
- check if logged in user is the same as previously stored establishmentId
- then conditionally set redirect if logged in user hasn't changed

### DIFF
--- a/src/app/core/services/establishment.service.ts
+++ b/src/app/core/services/establishment.service.ts
@@ -47,16 +47,12 @@ export class EstablishmentService {
   private _establishmentId: number = null;
 
   public checkIfSameLoggedInUser(establishmentId: number): void {
-    if (!this.previousEstablishmentId) {
+    if (!this.previousEstablishmentId || this.previousEstablishmentId !== establishmentId) {
       this.previousEstablishmentId = establishmentId;
       this.isSameLoggedInUser = false;
       return;
-    }
-
-    if (this.previousEstablishmentId === establishmentId) {
-      this.isSameLoggedInUser = true;
     } else {
-      this.isSameLoggedInUser = false;
+      this.isSameLoggedInUser = true;
     }
   }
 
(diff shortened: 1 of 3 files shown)


omarCreativeDev:
[(2) Defect - Sign out/in issue - Front End](https://trello.com/c/sja7Acds/38-2-defect-sign-out-in-issue-front-end)

src/app/core/services/establishment.service.ts
```diff
@@ -39,11 +39,27 @@ interface EmployerTypeRequest {
...
   public establishment$: Observable<Establishment> = this._establishment$.asObservable();
+  public previousEstablishmentId: number;
+  public isSameLoggedInUser: boolean;
 
   constructor(private http: HttpClient) {}
 
   private _establishmentId: number = null;
 
+  public checkIfSameLoggedInUser(establishmentId: number): void {
+    if (!this.previousEstablishmentId) {
```
ChazUK: What happens if I login to 3 different accounts? Looks like this will only run for the first time.

omarCreativeDev: good spot @ChazUK, i have tweaked the pr and tested thoroughly and the fix seems to be working fine

src/app/core/services/establishment.service.ts
```diff
@@ -39,11 +39,23 @@ interface EmployerTypeRequest {
...
 
   constructor(private http: HttpClient) {}
 
   private _establishmentId: number = null;
 
+  public checkIfSameLoggedInUser(establishmentId: number): void {
+    if (!this.previousEstablishmentId || this.previousEstablishmentId !== establishmentId) {
+      this.previousEstablishmentId = establishmentId;
+      this.isSameLoggedInUser = false;
+      return;
```
ChazUK: I don't think there's a need for this return anymore

PR #746, opened 2019-06-19
Bugfix | `js-enabled` not being appended to body

ChazUK:
Updating Index file to include `js-enabled` class


ChazUK:
@omarCreativeDev yep, no idea why we aren't using some kind of environment variables.

PR #747, opened 2019-06-19
Bulk Validation Validation Safari UI fix

tgreyuk:


tgreyuk:
[(3) Bulk Upload - Failed Validation](https://trello.com/c/ZvB9OwD3/61-3-bulk-upload-failed-validation)

src/assets/scss/components/_tables.scss
```diff
@@ -34,6 +21,10 @@ $error_row_padding: $govuk-gutter-half;
     top: 0;
   }
 
+  &:first-of-type:after {
+    left: -11px;
```
omarCreativeDev: -11px seems very specific, could we use something generic or flexbox perhaps? not a massive issue as i know we are using table and have to be careful about assessibility

PR #749, opened 2019-06-19
Bulk upload - Complete Upload - setting the scene

wozzer72:
This is the first of my Bulk Upload Complete PRs. It's a large change so I'd sooner drop code to you in sizeable chunks.

There is lots going on here:

1. `server/models/BulkImport/csv/establishments.js` -  Fixing the creation of Establishment entity for vacancies, starters and leavers following a recent correction to make the updates and fetches look the same.

2. `localhost.yaml` is a correction to the bucket name for bulk upload that I had managed to commit whilst testing dev deployment on Monday.

3. Allowing the restoring from both JSON and database not just the establishment, but the associative set of Workers too, and the Worker's associative set of Training and Qualifications - but only when told to do so (because it's an expensive operation otherwise). A large establishment can have 4000+ workers.

4. I've introduced the DataSource property across Establishments, Workers, Training and Qualifications, which not just impacting the sequelize models also impacts the entities's `save` method; for bulk upload creation, updates and deletes, we need to be able to override the "data source" on committing to database.

5. Whilst parsing the CSV and creating the entities, I now create the associations between Establishments, Workers, Training and Qualfiications, which allows me to upload a single "all entity" JSON document to S3 on finalising validation. 

Have a butchers at: `s3://sfcbulkuploadfiles/479/intermediary/all.entities.json` to see an example.

6. The new `[POST] complete` endpoint; it's WIP· As noted with a TODO - I do have to put in some protection to prevent premature completion. But it is in this bad puppy that I downloading and restore to kick start the completion. Doing this, means not having to repeat all the validation.

Completion is the driver for building the associative sets of data both on restore of DB and restore of JSON. Because now I've got the validation set and the current set of entities, I simply need to "smash them together" (nuclear fusion anyone) and then save. Well, that's the theory.

### DIFF
--- a/server/config/localhost.yaml
+++ b/server/config/localhost.yaml
@@ -21,4 +21,4 @@ log:
 
 bulkupload:
     region: eu-west-2
-    bucketname: sfc-bulkupload-dev
+    bucketname: sfcbulkuploadfiles
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -1613,11 +1613,9 @@ class Establishment {
           return returnThis;
         }),
       numberOfStaff: this._totalPermTemp,
-      jobs: {
-        vacancies: this._vacancies ? this._vacancies : undefined,
-        starters: this._starters ? this._starters : undefined,
-        leavers: this.leavers ? this.leavers : undefined,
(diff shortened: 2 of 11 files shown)


wozzer72:

PR #751, opened 2019-06-19
bugfix | redirect issue

omarCreativeDev:
- simplify and fix redirect to previous page if user logs in again

**Note:** This was the best solution I could come up without exposing the previous route in the browser address bar and also without touching the `window.location` object itself.

I also tried to use angular's `ActivatedRouteSnapshot` in the `AuthService` but angular throws injector errors, I then tried to use `ActivatedRoute` within `AuthService` but not all url segments were not exposed

### DIFF
--- a/src/app/core/components/header/header.component.html
+++ b/src/app/core/components/header/header.component.html
@@ -22,7 +22,7 @@
             <a routerLink="/account-management/your-account" class="govuk-header__link">Account</a>
           </li>
           <li class="govuk-header__navigation-item">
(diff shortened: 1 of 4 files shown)


omarCreativeDev:
[(1) Change Request - Username enumeration](https://trello.com/c/fbvQ0Poc/40-1-change-request-username-enumeration)

omarCreativeDev:
[(2) Defect - Sign out/in issue - Front End](https://trello.com/c/sja7Acds/38-2-defect-sign-out-in-issue-front-end)

PR #752, opened 2019-06-19
bugfix | username fixes

omarCreativeDev:
- only trigger username existance check on CTA form submit, which also fixes any password errors appearing too quickly
- fix regex pattern

### DIFF
--- a/src/app/core/constants/constants.ts
+++ b/src/app/core/constants/constants.ts
@@ -8,5 +8,5 @@ export const PASSWORD_PATTERN = /^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9]).{8,50}$/;
 export const INT_PATTERN = /^[0-9]*$/;
 export const FLOAT_PATTERN = /^([0-9]*[.])?[0-9]+$/;
 export const API_PATTERN = /^\/api\//;
-export const ALPHA_NUMERIC_WITH_HYPHENS_UNDERSCORES = /[a-z0-9_-]$/i;
+export const ALPHA_NUMERIC_WITH_HYPHENS_UNDERSCORES = /^[A-Za-z0-9_-]*$/;
 export const FILE_UPLOAD_TYPES = ['CSV', 'ZIP'];
--- a/src/app/features/registration/create-username/create-username.component.ts
+++ b/src/app/features/registration/create-username/create-username.component.ts
@@ -194,7 +194,6 @@ export class CreateUsernameComponent implements OnInit, OnDestroy {
         .getUsernameDuplicate(this.getUsername.value)
         .pipe(
           finalize(() => {
-            console.log('finally');
             this.submitted = true;
             this.onSubmit();
           })
@@ -209,7 +208,6 @@ export class CreateUsernameComponent implements OnInit, OnDestroy {
             }
           },
           (error: HttpErrorResponse) => {
-            console.log('error fired');
             this.serverError = this.errorSummaryService.getServerErrorMessage(error.status, this.serverErrorsMap);
           }
         )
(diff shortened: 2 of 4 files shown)


omarCreativeDev:
[(3) Defect - Username should not be allowed to contain spaces](https://trello.com/c/H2wJk7cE/19-3-defect-username-should-not-be-allowed-to-contain-spaces)

omarCreativeDev:
[Create username: Error messages shown too soon](https://trello.com/c/0rlXWtVt/131-create-username-error-messages-shown-too-soon)

src/app/core/constants/constants.ts
```diff
@@ -8,5 +8,5 @@ export const PASSWORD_PATTERN = /^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9]).{8,50}$/;
 export const INT_PATTERN = /^[0-9]*$/;
 export const FLOAT_PATTERN = /^([0-9]*[.])?[0-9]+$/;
 export const API_PATTERN = /^\/api\//;
-export const ALPHA_NUMERIC_WITH_HYPHENS_UNDERSCORES = /[a-z0-9_-]$/i;
```
ChazUK: I think you can also do `A-z` weird that the case insensitive flag didn't work.

src/app/features/registration/create-username/create-username.component.ts
```diff
@@ -195,27 +187,30 @@ export class CreateUsernameComponent implements OnInit, OnDestroy {
     ];
   }
 
-  private checkUsernameDoesntExist(userName: string): void {
+  public checkUsernameDoesntExist(): void {
```
ChazUK: could you not turn this into an async await function?

src/app/features/registration/create-username/create-username.component.ts
```diff
@@ -195,27 +187,30 @@ export class CreateUsernameComponent implements OnInit, OnDestroy {
...
           if (data['status'] === '1') {
-            this.submitted = true;
             this.getUsername.setErrors({ usernameExists: true });
           } else {
             this.getUsername.setErrors({ usernameExists: null });
             this.getUsername.updateValueAndValidity();
           }
         },
         (error: HttpErrorResponse) => {
+          this.submitted = true;
```
ChazUK: might as well move this to the complete section

omarCreativeDev: done

src/app/features/registration/create-username/create-username.component.ts
ChazUK: Basically the way the code works is you want to submit the form, and then check the username. If someone came to refactor this code and remove the check username call (unlikely) they would then have to move the submit call etc, rather than just removing this function.

You can use async/await and turn the Observable into a promise using `.toPromise()`

src/app/features/registration/create-username/create-username.component.ts
```diff
@@ -195,34 +187,34 @@ export class CreateUsernameComponent implements OnInit, OnDestroy {
...
-            this.submitted = true;
-            this.getUsername.setErrors({ usernameExists: true });
-          } else {
-            this.getUsername.setErrors({ usernameExists: null });
-            this.getUsername.updateValueAndValidity();
-          }
-        },
-        (error: HttpErrorResponse) => {
-          this.serverError = this.errorSummaryService.getServerErrorMessage(error.status, this.serverErrorsMap);
+  public async checkUsernameDoesntExist(): Promise<any> {
```
ChazUK: I think with this implementation you don't need async and await, as in the onSubmit you're still putting the logic inside the `then` of the Promise, which is the same as the success on the Observable.

https://medium.com/@balramchavan/using-async-await-feature-in-angular-587dd56fdc77 scroll down to Conditional Programming, close to the bottom of the page. 

`onSubmit` will become the async function, and inside that you are awaiting the `getUsernameDuplicate` function.

```
async getConditionalDataUsingAsync() {
    let data = await this.httpClient.get<Employee>(this.url).toPromise();
    if (data.id > 5) {
      let anotherUrl = 'http://dummy.restapiexample.com/api/v1/employee/23';
      this.conditionalAsyncResult = await this.httpClient.get<Employee>(anotherUrl).toPromise();
    }
    console.log('No issues, I will wait until promise is resolved..');
  }
```

PR #753, opened 2019-06-19
Feature | Workplace | Check Answers

ChazUK:
There's a few updates in this as it required a bit of refactoring to include fragments in the navigation.

I've created a check answers page for Workplace Questions and also updated the Staff Records Summary route to become Check Answers so that they are consistent.

I also updated the submit-button component as it was sending two events causing a weird bug when submitting answers


ChazUK:
[(5) Workplace - Check Answers Page](https://trello.com/c/bagKzWpe/118-5-workplace-check-answers-page)

PR #755, opened 2019-06-20
- Add fix to correctly display staff record across browsers

alleypa:
- Re-organise markup to solve specificity

### DIFF
--- a/src/app/features/workers/create-basic-records/create-basic-records.component.ts
+++ b/src/app/features/workers/create-basic-records/create-basic-records.component.ts
@@ -1,4 +1,4 @@
-import { Component, ElementRef, OnDestroy, OnInit } from '@angular/core';
+import { Component, ElementRef, OnDestroy, OnInit, Renderer2 } from '@angular/core';
 import { FormArray, FormBuilder, FormGroup, Validators } from '@angular/forms';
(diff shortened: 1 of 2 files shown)


alleypa:
[Basic record entry fields are squeezed for subsequent records](https://trello.com/c/Ct3zXmkc/116-basic-record-entry-fields-are-squeezed-for-subsequent-records)

src/app/features/workers/create-basic-records/create-basic-records.component.ts
```diff
@@ -128,7 +128,7 @@ export class CreateBasicRecordsComponent implements OnInit, OnDestroy {
     this.staffRecordsControl.controls[index].patchValue({ active: true });
     setTimeout(() => {
       this.elementRef.nativeElement
-        .querySelector(`.govuk-summary-list__row:nth-of-type(${index + 1})`)
+        .querySelector(`.staffRecord-view_${index}`)
```
ChazUK: If you're changing this, it might be worth making that an ID instead of doing this via queryselector

alleypa: Also better to use renderer for DOM access

PR #756, opened 2019-06-20
Feature | Workplace | Success

ChazUK:


ChazUK:
[(2) Workplace - Success Screen](https://trello.com/c/fAMjO7j3/120-2-workplace-success-screen)

src/app/features/workplace/success/success.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-success',
+  templateUrl: './success.component.html',
+})
+export class SuccessComponent implements OnInit {
+  constructor() {}
+
```
omarCreativeDev: do we need to set the back link?I wouldn't have thought so being the success/final screen in the flow but just wanted to double check...

ChazUK: I don't think we need to set the back link, I put the same comment on the trello card.

PR #758, opened 2019-06-20
Bugfix: Change 'tellUs' validation from words to maxlength

tgreyuk:


tgreyuk:
[(1) Feedback | No longer allowing 500 characters](https://trello.com/c/pfAFfdv1/102-1-feedback-no-longer-allowing-500-characters)

src/app/features/public/feedback/feedback.component.html
```diff
@@ -61,7 +61,7 @@ <h1 class="govuk-heading-xl">Feedback</h1>
               formControlName="tellUs"
             ></textarea>
           </div>
-          <app-character-count max="10" [words]="true" [control]="form.get('tellUs')"></app-character-count>
+          <app-character-count max="500" [control]="form.get('tellUs')"></app-character-count>
```
ChazUK: Can you pull this out into a variable in the component, I think the same thing has been done for the other textarea

tgreyuk:
@ChazUK updated... it worked on FE because angular handles maxlength by default..
`Validator that requires the length of the control's value to be less than or equal to the provided maximum length. This validator is also provided by default if you use the the HTML5 maxlength attribute.`

PR #760, opened 2019-06-20
Updated error message for job role

tgreyuk:
Most of the changes seem to be formatting because of webstorm.


tgreyuk:
[(1) Other job role too long - error message doesn't make sense](https://trello.com/c/1jxWBKFU/103-1-other-job-role-too-long-error-message-doesnt-make-sense)

PR #761, opened 2019-06-20
Bugfix | Fixing Navigation for Workplace Start Screen

ChazUK:
forgot establishmentId 😔

### DIFF
--- a/src/app/features/workplace/start/start.component.html
+++ b/src/app/features/workplace/start/start.component.html
@@ -4,6 +4,11 @@
   starters and staff who have left in the last 12 months. We would also like to know about your current vacancies.
 </p>
 <p>Once you have added it you can change this information at any time from the workplace tab.</p>
-<a routerLink="/workplace/type-of-employer" role="button" draggable="false" class="govuk-button govuk-button--start">
+<a
+  [routerLink]="['/workplace', establishment.id, 'type-of-employer']"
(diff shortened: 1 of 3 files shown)


src/app/features/workplace/start/start.component.html
```diff
@@ -4,6 +4,11 @@ <h1 class="govuk-heading-xl">Add your workplace information</h1>
   starters and staff who have left in the last 12 months. We would also like to know about your current vacancies.
 </p>
 <p>Once you have added it you can change this information at any time from the workplace tab.</p>
-<a routerLink="/workplace/type-of-employer" role="button" draggable="false" class="govuk-button govuk-button--start">
+<a
+  [routerLink]="['/workplace', establishment.id, 'type-of-employer']"
+  role="button"
```
tgreyuk: I know this was already here but role button i don't think should be used here. role button is semantic rather than presentational:
```
The button role identifies an element as a button to screen readers. A button is a widget used to perform actions such as submitting a form, opening a dialog, cancelling an action, or performing a command such as inserting a new record or displaying information.
```

ChazUK: I would normally agree, but the GOV.UK Framework has this for the start buttons https://design-system.service.gov.uk/components/button#start-buttons, not sure of exactly why

PR #768, opened 2019-06-21
Bulk upload complete validate difference report

wozzer72:
Hi Darren

This is the validation difference report, which needs to load all current known establishments and their workers, and compare against the constructed set of establishments and workers identifying new, updated and deleted.

The difference report is uploaded to S3; I will consume it on the "complete" (the last I have to do to create, update and delete establishments and workers).

The difference report is then used to count the number of deleted establishments and deleted workers, which will be used by the frontend to display a popup modal as a warning.

Note - we cannot delete the primary establishment (the establishment of the logged in user). I do need to raise an error validation message. I'll have a look at doing that whilst you review this puppy.

Additionally, thus PR includes https://trello.com/c/F9jYfkhS; a fixed for a defect that was duplicate entity validation messages.

Thanks

### DIFF
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -45,6 +45,9 @@ class Establishment {
     //console.log(`WA DEBUG - current establishment (${this._lineNumber}:`, this._currentLine);
   };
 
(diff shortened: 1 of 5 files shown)


wozzer72:

PR #782, opened 2019-06-23
Bulk upload complete training and qualifications

wozzer72:
Hey Darren

https://trello.com/c/QAzbzesV

Well this should have been easier than I had expected, but it pretty much done.

This PR includes updating associated qualifications and training records when a Worker entity is saved. As there is no change audit on training or qualifications, its a simple case of delete all such records, then create new records with the data as given on bulk upload last Worker records.

To allow for that, I had to introduce setters on the Training/Qualification entities to set the associated establishment and worker id, necessary to create the foreign keys in database. I also took the opportunity for a quick optimisation in each of `Qualification::save` and `Training::save` so if the `WorkerID` is known (it would be on bulk upload through the associations) then no need to look up using the Worker UID (as required when called upon from frontend).

There is one known defect - that associated with creating the qualification records; for some reason, the onload qualifications are missing the `qualification.id` - I've raised a defect for that whilst commenting out the code, because I want to get this card moved into test and ready for the demo.

### DIFF
--- a/server/config/development.yaml
+++ b/server/config/development.yaml
@@ -5,7 +5,7 @@ aws:
         wallet: dev/api
 
 db:
-    pool: 5
+    pool: 15
     ssl: true
     username: sfcapp
     client_ssl:
--- a/server/config/localhost.yaml
+++ b/server/config/localhost.yaml
@@ -10,7 +10,7 @@ jwt:
     ttl:
         login: 60
 db:
-    pool: 5
+    pool: 15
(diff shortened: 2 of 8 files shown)


wozzer72:

wozzer72:
[Qualifications failing to create on Bulk Upload](https://trello.com/c/U5fVujdg/183-qualifications-failing-to-create-on-bulk-upload)

PR #785, opened 2019-06-23
bugfix | remove files still shows files count next to upload btn

omarCreativeDev:
currently if you select some files and click the remove btn the files are not removed from the dom node. this pr is to address that


wozzer72:
[Qualifications failing to create on Bulk Upload](https://trello.com/c/U5fVujdg/183-qualifications-failing-to-create-on-bulk-upload)

PR #790, opened 2019-06-24
bugfix | dont call prevalidate api on fetched uploaded-files

omarCreativeDev:
- no need to invoke the prevalidate files again as this step is immediately done once files are uploaded to s3


omarCreativeDev:
[Dont prevalidate files twice](https://trello.com/c/s7qRXmOP/200-dont-prevalidate-files-twice)

PR #792, opened 2019-06-24
Feature/workplace services capacity property validation

AndrewKeig:


AndrewKeig:
[(5) Workplace Services/Capacity Property Validation](https://trello.com/c/00kcFsia/92-5-workplace-services-capacity-property-validation)

PR #797, opened 2019-06-25
Bugfix/changes out of demo

alleypa:


alleypa:
[UI Changes out of Demo to Design](https://trello.com/c/DADeQizz/149-ui-changes-out-of-demo-to-design)

src/app/core/services/job.service.ts
```diff
@@ -11,6 +11,21 @@ export class JobService {
   constructor(private http: HttpClient) {}
 
   public getJobs(): Observable<Job[]> {
-    return this.http.get<GetJobsResponse>('/api/jobs').pipe(map(res => res.jobs));
+    return this.http.get<GetJobsResponse>('/api/jobs').pipe(map(this.sortJobs));
```
ChazUK: Could we not use lodash?

```
return this.http.get<GetJobsResponse>('/api/jobs').pipe(
  map(jobs => sortBy(jobs, ['other', 'title'])
);
```

Not sure if it's possible to do a `!` to reverse the order of 'other' so false is listed first.

src/app/features/workplace/data-sharing-with-local-authorities/data-sharing-with-local-authorities.component.ts
```diff
@@ -103,6 +105,17 @@ export class DataSharingWithLocalAuthoritiesComponent extends Question {
     };
   }
 
+  public selectableAuthorities(index): LocalAuthorityModel[] {
```
ChazUK: Could probably combine these two as the second function isn't going to be used anywhere else, and you don't need the middle return, can write that shorthand

```
public selectableAuthorities(index): LocalAuthorityModel[] {
  return this.authorities.filter(
    authority =>
      !this.localAuthoritiesArray.controls.some(
        localAuth =>
          localAuth !== this.localAuthoritiesArray.controls[index] &&
          parseInt(localAuth.get('custodianCode').value, 10) === authority.custodianCode
      )
  );
}
```

src/app/features/workplace/start/start.component.html
```diff
@@ -1,4 +1,25 @@
...
+      draggable="false"
+      class="govuk-button govuk-button--start"
+    >
+      Start now
+    </a>
+  </div>
+</div>
+
+
+<!-- <h1 class="govuk-heading-xl">Add your workplace information</h1>
```
ChazUK: This can probably be removed

src/app/features/workplace/data-sharing/data-sharing.component.html
```diff
@@ -77,7 +80,10 @@ <h2 class="govuk-fieldset__heading">Do you agree to us sharing data with:</h2>
           </div>
         </fieldset>
       </div>
-      <app-submit-button [return]="!!return" (clicked)="onSubmit($event)"></app-submit-button>
-    </form>
-  </div>
+    </div>
+
+    <div class="govuk-grid-column-full">
```
ChazUK: You have `govuk-grid-column-full` and `govuk-grid-column-full-from-desktop` on these two files, Should they be the same?

src/app/core/services/job.service.ts
alleypa: sorting this way does not execute as intended, Hence the use of `reduce` to guarantee we always have jobs with other at the bottom

src/app/features/workplace/data-sharing-with-local-authorities/data-sharing-with-local-authorities.component.ts
alleypa: done

src/app/features/workplace/data-sharing/data-sharing.component.html
alleypa: govuk-grid-column-full used for consistency

PR #804, opened 2019-06-25
bugfix | bulk upload | api call cancelled on files selection

omarCreativeDev:
- on successful file upload to s3, fire prevalidate observable event in order for the uploaded files component to go ahead and prevalidate the files. This fixes the bug where on files selection the prevalidate api was being called and cancelled immediately


omarCreativeDev:
[Api call cancelled on files selection](https://trello.com/c/hQb8yIH7/201-api-call-cancelled-on-files-selection)

PR #805, opened 2019-06-25
bugfix | unable to choose files after validate step

omarCreativeDev:
- ensure that upon successful files upload to s3 the file input dom element is cleared. Otherwise once files are validated, and in the scenario where the exact same files with the same files names are chosen the observable doesnt fire as technically the data stream hasn't changed


omarCreativeDev:
[Unable to choose 3 files](https://trello.com/c/RcqYl7bJ/202-unable-to-choose-3-files)

PR #816, opened 2019-06-26
Bug Fix | Starters, Leavers, Vacancies submission issue

ChazUK:


ChazUK:
[Cannot return from starters/leavers/vacancies without reselecting the same option](https://trello.com/c/B3rVTIoy/141-cannot-return-from-starters-leavers-vacancies-without-reselecting-the-same-option)

PR #818, opened 2019-06-27
Removing suppression of establishment capacity and services propertie…

wozzer72:
…s validation now the properties are successsfully validating against Cache.


wozzer72:
[(5) Workplace Services/Capacity Property Validation](https://trello.com/c/00kcFsia/92-5-workplace-services-capacity-property-validation)

PR #821, opened 2019-06-27
Feature/validation training

AndrewKeig:
https://trello.com/c/IAs74P7y/118-3-validation-training

- This ticket does not cover validation, file structure, so column heading validation needs to be done in another story.
- This ticket does not validate deletions
- The codes we already using, are to be used.


AndrewKeig:
[(3) Validation - Training](https://trello.com/c/IAs74P7y/118-3-validation-training)

PR #830, opened 2019-06-28
Refactoring to make better use of aysnc/await to better handling of c…

wozzer72:
…onditional flow logic. Also updating the query to find subsidariry of a parent to allow for passing the establishment id rather than the UID.

https://trello.com/c/PgoCDh3C


wozzer72:
[Subsidairy endpoints only accept establishment UID](https://trello.com/c/PgoCDh3C/218-subsidairy-endpoints-only-accept-establishment-uid)

PR #832, opened 2019-06-28
Introducing "public/download" proxy to S3 buckets.

wozzer72:
https://trello.com/c/LyVCfwYI

Hi Nasir

Requesting your review of this backend change, because the change is more about the application hosting environment than it is about code.

As per discussion yesterday, I'm introducing "npm proxy" to the `server.js` allowing me to proxy all requests to `public/download` to an S3 bucket. Using a new config property `public.download.baseurl` I am able to redirect to URL, but most significantly, control which of three S3 buckets to redirect to - config by environment specific URL.

I've introduced three new buckets allowing public access; one for each of dev, staging and production.

I've uploaded the current set of guidance doc PDFs and I've made them public.

Note, every time the PDFs are uploaded, they have to be make public again.

### DIFF
--- a/server.js
+++ b/server.js
@@ -61,7 +61,7 @@ app.use('/public/download', proxy(
     {
         proxyReqPathResolver: function (req) {
           const updatedPath = publicDownloadBaseUrl + req.url;
-          console.log("public/download proxy API request to: ", `${updatedPath}`)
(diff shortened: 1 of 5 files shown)


wozzer72:
[(4) Bulk Upload Guidance](https://trello.com/c/LyVCfwYI/53-4-bulk-upload-guidance)

PR #835, opened 2019-06-28
Added pipe to display now known test for weeklyHoursAverage

tgreyuk:


tgreyuk:
[(2) Defect - Staff record summary: Transform Do you know = 'No' and 'Don't know' answers to 'Not known' - Front End](https://trello.com/c/mtV3r7r4/39-2-defect-staff-record-summary-transform-do-you-know-no-and-dont-know-answers-to-not-known-front-end)

src/app/features/workers/staff-record-summary/employment/employment.component.html
```diff
@@ -302,9 +302,9 @@ <h2 class="govuk-heading-m">Employment</h2>
     </dt>
     <dd class="govuk-summary-list__value">
       {{
-        worker.weeklyHoursAverage?.hours !== null
+        worker.weeklyHoursAverage?.hours
```
ChazUK: I think there was a reason for this checking null, I believe 0 is a valid answer, but JS picks that up as false.

ChazUK: Or is that now handled in your pipe?

tgreyuk: Thanks will check it.

tgreyuk:
@ChazUK As result of your comments I found a bug when saving a 0 value. Also the null check did not work properly, as by definition the else was always null and didn't display anything.

PR #836, opened 2019-06-28
validating staff records

AndrewKeig:
https://trello.com/c/XHpJc6bC/125-2-validation-staff-records-bulk-upload-column-headings-mandatory


server/models/BulkImport/csv/workers.js
```diff
@@ -383,7 +396,7 @@ class Worker {
         lineNumber: this._lineNumber,
         errCode: Worker.STATUS_ERROR,
         errType: `STATUS_ERROR`,
-        error: `Status (STATUS) must be one of: ${statusValues}`,
+        error: `The status you have supplied is incorrect`,
```
wozzer72: How does this differ from `You have not supplied a correct status` added above?

AndrewKeig: Ive rolled back this change, should not have gone in, as this relates to status, which has its own ticket.

PR #845, opened 2019-06-28
Accessibility | Warning for anchors opening in new window

ChazUK:
Added a CSS class for all anchors with the attribute `target="_blank"`
Also added a visually hidden context to each anchor that opens in a new window so screen readers will be warned.


ChazUK:
[Accessibility - Warn before opening a new window](https://trello.com/c/876r4CAG/180-accessibility-warn-before-opening-a-new-window)

PR #846, opened 2019-06-28
Accessibility | Removing Global Style Classes

ChazUK:


ChazUK:
@omarCreativeDev All `a` elements and `p` elements have a global style of `govuk-link` and `govuk-body` respectively

PR #849, opened 2019-06-28
Accessibility | Removing incorrect use of title attribute

ChazUK:
Binding custom directive `title` so HTML doesn't pick it up


ChazUK:
[Accessibility - Staff Record - Redundant Title Text](https://trello.com/c/smPYsRF4/189-accessibility-staff-record-redundant-title-text)

PR #857, opened 2019-06-30
Accessibility | Removing bad ARIA reference

ChazUK:


ChazUK:
[Accessibility - Broken ARIA Reference](https://trello.com/c/IexZYLCP/172-accessibility-broken-aria-reference)

PR #861, opened 2019-07-01
feature | bulk upload | references | get workers

omarCreativeDev:
- invoke `getAllWorkers` api call and update ui
- display subsidiary name and enter temporary workplaces remaining count copy
- conditionally display name or nameOrId property on label
- relocated `WorkersResponse` and `WorkerEditResponse` to model file so they can be reused
- moved common 503 error code to base class
- removed/relocated userService from base class as only one child class needed it
- adjust cols width as per design

### DIFF
--- a/src/app/core/model/worker.model.ts
+++ b/src/app/core/model/worker.model.ts
@@ -85,3 +85,11 @@ export interface Worker {
   updated?: string;
   completed?: boolean;
 }
+
+export interface WorkersResponse {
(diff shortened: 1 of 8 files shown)


omarCreativeDev:
[(18) Bulk Upload - Local Identifiers Workers](https://trello.com/c/CzEWY2WN/48-18-bulk-upload-local-identifiers-workers)

PR #862, opened 2019-07-01
Display error panel for postcode validation with a server error

tgreyuk:


tgreyuk:
Looking at it I think they both work because i think server error will always be submitted but should be consistent..

tgreyuk:
[(2) Find your workplace address doesn't display error message when postcode not found](https://trello.com/c/OQzfRwGa/226-2-find-your-workplace-address-doesnt-display-error-message-when-postcode-not-found)

PR #868, opened 2019-07-01
worker local identifier

dbjsdev:


server/models/classes/worker.js
```diff
@@ -847,6 +861,7 @@ class Worker extends EntityValidator {
                 fetchResults.forEach(thisWorker => {
                     allWorkers.push({
                         uid: thisWorker.uid,
+                        localIdentifier: thisWorker.LocalIdentifierValue ? thisWorker.LocalIdentifierValue : undefined ,
```
wozzer72: Replace `undefined` with `null` - undefined will mean when the JSON is returned to the frontend, if attribute will be missing. But it will be more consistent for Omar if the attribute was present but null.

dbjsdev: resolved

PR #869, opened 2019-07-01
Linked error message to error summary for country of birth

tgreyuk:


tgreyuk:
[Error summary not linking and messages not shown against the field - Nationality and Country of birth](https://trello.com/c/gnqJFqsn/225-error-summary-not-linking-and-messages-not-shown-against-the-field-nationality-and-country-of-birth)

PR #870, opened 2019-07-01
Render messages with innerHTML to prevent character escaping

tgreyuk:


tgreyuk:
[Error messages showing "&pound:" instead of "£" for hourly or annual pay](https://trello.com/c/7rBztteU/227-error-messages-showing-pound-instead-of-for-hourly-or-annual-pay)

PR #872, opened 2019-07-01
Accessibility | Tab Controls IE

ChazUK:
Adding missing IE event key names


ChazUK:
[Accessibility - Tabs - Navigation and Content Access](https://trello.com/c/dUZyYSUV/190-accessibility-tabs-navigation-and-content-access)

PR #879, opened 2019-07-02
Merge Accessibility config changese into Develop

nasirmids:
Hello, 

As part of client specfic environment, i have reuse the existing accessibility config file and copied all test config in it except the database name and user which points to accessibility related database. The AWS secret has also changed and point to demo/api.

In summary
- URL is sfcaccessbility
- DB is NEW and separate from dev/staging
- Application config file is accessibility.config
- AWS Secret (user, policy etc) new and separate  dev/staging

Thanks

### DIFF
--- a/server/config/accessibility.yaml
+++ b/server/config/accessibility.yaml
@@ -1,6 +1,35 @@
+aws:
+    region: eu-west-1
+    secrets:
+        use: true
+        wallet: demo/api
+db:
+    pool: 15
+    ssl: true
+    username: sfcdemoadmin
+    database: sfcdemodb
+    client_ssl:
+        status: true
+        usingFiles: false
+
 slack:
     level : 3     # enables Slack notifications
+
 jwt:
     iss: 'sfcaccessibility.cloudapps.digital'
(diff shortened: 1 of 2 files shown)


server/config/accessibility.yaml
```diff
@@ -1,6 +1,35 @@
...
+
+notify:
+    replyTo: '73711295-a4da-4303-9127-65be2a409681'
+    templates:
+        resetPassword: 'f2c43610-9ec9-4271-b9b1-dd55569b191c'
+        addUser: 'e581eea2-4206-4c90-8594-4357960e3e74'
+
+bulkupload:
+    region: eu-west-2
+    bucketname: sfc-bulkupload-staging
```
wozzer72: Needs it's own bucket. The establishment id (primary key) may conflict with staging.

nasirmids: ok i will change it

nasirmids: New S3 bucked called sfcbulkupload-demo and new policy sfc-demo-s3 created.

PR #888, opened 2019-07-03
Feature/bulk upload references convert api call to put

omarCreativeDev:
- add missing primary establishment to view my workplaces and bulk upload references screen
- convert x number of POST requests to a single PUT request for a mass update
- added new LocalIdentifiersResponse interface

### DIFF
--- a/src/app/core/services/establishment.service.ts
+++ b/src/app/core/services/establishment.service.ts
@@ -1,10 +1,10 @@
+import { AllServicesResponse, ServiceGroup } from '@core/model/services.model';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { Establishment, LocalIdentifiersRequest, LocalIdentifiersResponse, UpdateJobsRequest } from '@core/model/establishment.model';
 import { HttpClient, HttpParams } from '@angular/common/http';
 import { Injectable, isDevMode } from '@angular/core';
-import { Establishment, LocalIdentifierRequest, UpdateJobsRequest } from '@core/model/establishment.model';
-import { AllServicesResponse, ServiceGroup } from '@core/model/services.model';
-import { URLStructure } from '@core/model/url.model';
-import { BehaviorSubject, Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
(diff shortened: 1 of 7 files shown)


omarCreativeDev:
[(18) Bulk Upload - Local Identifiers Establishment](https://trello.com/c/AscBN35F/47-18-bulk-upload-local-identifiers-establishment)

src/app/features/workplace/workplace-info-panel/workplace-info-panel.component.ts
```diff
@@ -11,6 +11,7 @@ export class WorkplaceInfoPanelComponent {
   get canAccessWorkplace() {
     return (
       this.workplace.dataOwner === WorkplaceDataOwner.Parent ||
+      this.workplace.dataOwner === WorkplaceDataOwner.Workplace ||
```
tgreyuk: This is used to set the availability of the parent and subs links. Are you sure this won't break that. Unless something has changed the condition the condition was explicity for parent or those with parent permissions.

src/app/core/services/establishment.service.ts
```diff
@@ -158,7 +158,10 @@ export class EstablishmentService {
     return this.http.post<Establishment>(`/api/establishment/${establishmentId}/jobs`, data);
   }
 
-  public updateLocalIdentifier(establishmentUid: string, localIdentifier: LocalIdentifierRequest): Observable<any> {
-    return this.http.post<any>(`/api/establishment/${establishmentUid}/localIdentifier`, localIdentifier);
+  public updateLocalIdentifiers(request: LocalIdentifiersRequest): Observable<LocalIdentifiersResponse> {
```
ChazUK: still think we should be passing the establishmentID into the function, but should be ok

src/app/features/workplace/workplace-info-panel/workplace-info-panel.component.ts
omarCreativeDev: yea i spoke to Warren about this before making the change and he said that all establishments should be editable, the above change was needed otherwise the hyperlink was missing for the parent establishment

PR #891, opened 2019-07-04
Localident ext

dbjsdev:
Hi Warren

These will probably need altering but thought I would put up for your review. I have made the error return the property 'duplicateValue'. Currently the worker mass update returns a similar response to the establishment mass update, but may need something different

### DIFF
--- a/server/routes/establishments/localIdentifier.js
+++ b/server/routes/establishments/localIdentifier.js
@@ -76,6 +76,7 @@ router.route('/').post(async (req, res) => {
       return res.status(503).send(err.safe);
     } else {
       console.error("Unexpected exception: ", err);
+      return res.status(503).send(err.safe);
     }
   }
 });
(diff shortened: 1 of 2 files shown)


server/routes/establishments/localIdentifier.js
```diff
@@ -202,20 +202,13 @@ router.route('/').put(async (req, res) => {
...
-      console.error("Establishment::localidentifier PUT: ", err.message);
-      return res.status(400).send(err.safe);
-    } else if (err instanceof Establishment.EstablishmentExceptions.EstablishmentSaveException) {
-      console.error("Establishment::localidentifier PUT: ", err.message);
-      return res.status(503).send(err.safe);
-    } else {
-      console.error("Unexpected exception: ", err);
+    if (err.name && err.name === 'SequelizeUniqueConstraintError') {
+      if(err.parent.constraint && ( err.parent.constraint === 'establishment_LocalIdentifier_unq')){
+          console.error("Establishment::localidentifier PUT: ", err.message);
```
wozzer72: can we move this error log to the more general 503 section

PR #898, opened 2019-07-04
For all array type properties in Establishments, initialising with an…

wozzer72:
… empty array if data is not given in CSV

https://trello.com/c/bkElbq7z


wozzer72:
[Complete upload is not clearing existing services and capacities - backend](https://trello.com/c/bkElbq7z/238-complete-upload-is-not-clearing-existing-services-and-capacities-backend)

PR #901, opened 2019-07-05
164 bug valid html square brackets

bwilson-ux:
Give this a check over and see if it now fits easily into the new dev branch - if not I can revisit on Monday. Cheers!


ChazUK:
It's not just formControlName that needs to be wrapped in square brackets, it's any Angular Custom Directive

ChazUK:
Gonna close this as it's not the full fix required to make the HTML valid.

PR #915, opened 2019-07-08
CSS responsive fixes for 800px breakpoints not covered by standard GDS

bwilson-ux:
Update available for review.

### DIFF
--- a/src/assets/scss/modules/_responsive.scss
+++ b/src/assets/scss/modules/_responsive.scss
@@ -5,6 +5,20 @@
   outline: 0 !important;
 }
 
+
+@media(max-width: 800px){
+  .govuk-grid-column-two-thirds-from-desktop {
+      width: 100% !important;
+  }
+  #staff-records {
+    .govuk-list--definition {
+      .govuk-util__float-right {
+        float: none !important;
+      }
+    }
+  }
+}
+
 @include govuk-media-query($until: desktop) {
 
   .has-divider {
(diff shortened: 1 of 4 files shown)


src/assets/scss/modules/_responsive.scss
```diff
@@ -5,6 +5,20 @@
   outline: 0 !important;
 }
 
+
+@media(max-width: 800px){
```
ChazUK: I don't think we should be adding any custom breakpoints, as we'll then have to support them outside of the GOV.UK Framework updates.

src/assets/scss/modules/_responsive.scss
```diff
@@ -5,6 +5,20 @@
...
+  .govuk-grid-column-two-thirds-from-desktop {
+      width: 100% !important;
+  }
+  #staff-records {
+    .govuk-list--definition {
+      .govuk-util__float-right {
+        float: none !important;
```
ChazUK: Perhaps it'd be better to add some left padding to this element to make sure that it doesn't butt up against the left aligned text.

I just spoke with Ann and Rich and he said that it's probably better to move the Change link to be in the row above next to the number. Louise didn't have a problem with this as long as the hidden contextual text is still there.

Does that sounds like a better solution?

bwilson-ux: Both are viable - with regards to the 800px width view, all text will not fit on one line anways so will always drop to a 3rd row - I suppose thats the main issue and will probably be best to report to Rich this edge case.

PR #920, opened 2019-07-09
Bulk upload changes - following review with Jackie on both upload and download

wozzer72:
Too many changes to list, but essentially:

1. Updating Worker validation - correcting errors that were not correcting being reported to file, but were counting.
2. Updating Worker validation - to allow optional on year of entry, british citizenship and nationality.
3. Allowing bulk upload to update location id and whether an establishment is CQC regulated - a simple update for now, a trello backlog card has been raised to make them change properties.
4. Establishment service users can be empty list; and correcting the set of service users that can have other data.
5. Establishment service others qualifier not mandatory.
6. On Establishment services and capacity properties, making the "load" set of all services/all capacities override those on fetch, because when changing "CQC regulated", the set of services associated when making that change are needed for validation.
7. Worker AVG/CONT hours, allow input of 999 (was hitting the 75 max validation constraint).
8. Worker -CHGUNIQUEWRKID column is optional on upload, and not present on download.
9. Worker Non Social Care Qualiication level is optional and must map to "Don't know".


wozzer72:
[Bulk Upload - Issues](https://trello.com/c/gwuJenXy/269-bulk-upload-issues)

PR #924, opened 2019-07-09
Feature | Breadcrumbs

ChazUK:
We're going to have to fiddle around with the routes and titles for pages to make sure we aren't receiving any double pages, and the nesting is working correctly


ChazUK:
[(3) Breadcrumb Component](https://trello.com/c/VtvbKp3X/160-3-breadcrumb-component)

src/app/shared/components/breadcrumbs/breadcrumbs.component.html
```diff
@@ -0,0 +1,22 @@
+<ng-container *ngIf="display">
+  <div class="govuk-breadcrumbs">
+    Go back to:
+    <ol class="govuk-breadcrumbs__list">
```
omarCreativeDev: how comes this is an `<ol>`? wouldn't this render numbers on the list items or are those being hidden for assessibility?

src/app/core/services/nested-routes.service.ts
```diff
@@ -0,0 +1,64 @@
...
+        map(() => this.activatedRoute)
+      )
+      .subscribe(route => {
+        const routes = this.getNestedRoutes(route);
+        this._routes$.next(routes);
+      });
+  }
+
+  private getNestedRoutes(route: ActivatedRoute, url: string = '', routes: IRoute[] = []): IRoute[] {
+    const ROUTE_DATA_BREADCRUMB = 'title';
```
omarCreativeDev: why is this variable not camelCased?

src/app/core/components/header/header.component.html
```diff
@@ -28,7 +28,7 @@
         <ng-container *ngIf="isLoggedIn()">
           <li class="govuk-header__navigation-item">{{ fullname }}</li>
           <li class="govuk-header__navigation-item">
-            <a routerLink="/account-management/your-account" class="govuk-header__link">Account</a>
+            <a routerLink="/account-management" class="govuk-header__link">Account</a>
```
omarCreativeDev: this change makes sense, keep things consistent with the other modules

src/app/shared/components/breadcrumbs/breadcrumbs.component.html
ChazUK: I took the start of this markup from the GOV.UK Framework https://design-system.service.gov.uk/components/breadcrumbs/

I think the logic why it's an `ol` and not a `ul` is that there is a hierarchy, level 1, level 2, level 3 and so on.

src/app/core/services/nested-routes.service.ts
ChazUK: yep, I can update this or even remove it.

src/app/core/components/header/header.component.html
ChazUK: There might be a few other places we need to check this too.

src/app/core/services/nested-routes.service.ts
omarCreativeDev: if you can rename it to camelCase that would be super

src/app/core/components/header/header.component.html
omarCreativeDev: I will keep an eye out for this once i jump on to the user accounts ticket

src/app/core/services/nested-routes.service.ts
ChazUK: done

PR #925, opened 2019-07-09
fix validations

AndrewKeig:


server/models/BulkImport/csv/workers.js
```diff
@@ -811,25 +811,20 @@ class Worker {
...
-          errType: 'RECSOURCE_ERROR',
-          error: "The code you have entered for RECSOURCE is incorrect",
-          source: this._currentLine.RECSOURCE,
-        });
-        return false;
-      } else {
-        this._recSource = myRecSource;
-        return true;
-      }
+    if (this._currentLine.RECSOURCE && (isNaN(myRecSource) || myRecSource === 0 )) {
```
wozzer72: The rec source of 0 being invalid should be picked up by the transform recruitment source function - in that 0 should not be a BUDI mapped option. Likewise, any minus number is invalid.

server/models/BulkImport/csv/workers.js
```diff
@@ -811,25 +811,20 @@ class Worker {
...
+        name: this._currentLine.LOCALESTID,
+        lineNumber: this._lineNumber,
+        errCode: Worker.RESOURCE_ERROR,
+        errType: 'RECSOURCE_ERROR',
+        error: "The code you have entered for RECSOURCE is incorrect",
+        source: this._currentLine.RECSOURCE,
+      });
+      return false;
     } else {
+      this._recSource = myRecSource;
```
wozzer72: This will set the internal recruitment source to an empty value (from it's original null value) if the recruitment source column is empty. Is this the expected behaviour?

server/models/BulkImport/csv/workers.js
```diff
@@ -920,7 +929,7 @@ class Worker {
         source: this._currentLine.STARTINSECT,
       });
       return false;
-    } else if (this._startDate && parseInt(myStartInsect) > this._startDate.year()) {
+    } else if (myRealStartDate && parseInt(myStartInsect) > myRealStartDate.year()) {
```
wozzer72: Please remind me, what is `myRealStartDate` if `myStartDate` is not a valid date?

server/models/BulkImport/csv/workers.js
```diff
@@ -1003,79 +1012,78 @@ class Worker {
...
+        name: this._currentLine.LOCALESTID,
+        lineNumber: this._lineNumber,
+        warnCode: Worker.ZERO_HRCONT_WARNING,
+        warnType: 'ZERO_HRCONT_WARNING',
+        warning: "You have entered contracted hours but have not said this worker is not on a zero hours contract",
+        source: this._currentLine.ZEROHRCONT,
+      });
+      return false;
+    }
+    else if (isNaN(myZeroHourContract) || !zeroHourContractValues.includes(myZeroHourContract)) {
```
wozzer72: This validation should be covered by `_validateZeroContractHours` - and not in this `_validateContractedHours` function.

wozzer72: And ensure that the call to validate zero contact hours in `validate` occurs before contracted hours and average hours validation.

server/models/BulkImport/csv/workers.js
```diff
@@ -1003,79 +1012,78 @@ class Worker {
...
+        name: this._currentLine.LOCALESTID,
+        lineNumber: this._lineNumber,
+        warnCode: Worker.ZERO_HRCONT_WARNING,
+        warnType: 'ZERO_HRCONT_WARNING',
+        warning: "You have entered “0” in CONTHOURS but not entered “Yes” to the ZEROHRCONT question",
+        source: this._currentLine.ZEROHRCONT,
+      });
+      return false;
     } else {
+      switch (myZeroHourContract) {
```
wozzer72: OK. I can see the logic, that zero contract hours is mostly related to contracted hours, but by validating the values here, the individual column validation is being lost.

server/models/BulkImport/csv/workers.js
```diff
@@ -1956,7 +1977,7 @@ class Worker {
             lineNumber: this._lineNumber,
             errCode: Worker.RECSOURCE_ERROR,
             errType: `RECSOURCE_ERROR`,
-            error: `Recruitement Source (RECSOURCE): ${this._recSource} is unknown`,
+            error: `The code you have entered for RECSOURCE is incorrect`,
```
wozzer72: This relates to my comment above about validating recSource being 0. That 0 check should be caught here.

server/models/BulkImport/csv/workers.js
AndrewKeig: Fixed.

server/models/BulkImport/csv/workers.js
AndrewKeig: Fixed. Set to null

server/models/BulkImport/csv/workers.js
AndrewKeig: removed this.

server/models/BulkImport/csv/workers.js
AndrewKeig: Fixed

server/models/BulkImport/csv/workers.js
AndrewKeig: @wozzer72 , The validation rules document has these assigned to ZEROHRCONT., not sure what you want here?

server/models/BulkImport/csv/workers.js
AndrewKeig: Dont understand?

server/models/BulkImport/csv/workers.js
wozzer72: Why remove it? It was a question!

server/models/BulkImport/csv/workers.js
wozzer72: Howabout you call me?

server/models/BulkImport/csv/workers.js
wozzer72: Again then - call me as I suggested in Slack when I announced I had made comments!

PR #928, opened 2019-07-10
Removing CQC help text when Establishment is not CQC Regulated

ChazUK:


ChazUK:
[Sharing your data - CQC text needs to be removed for non-cqc user](https://trello.com/c/keqtFFNN/306-sharing-your-data-cqc-text-needs-to-be-removed-for-non-cqc-user)

PR #929, opened 2019-07-10
Created account module, account details component so its reuseable ac…

omarCreativeDev:
Created account module, account details component so code is re usable across registration, account creation and changing account details flows


omarCreativeDev:

PR #934, opened 2019-07-11
Feature/change your details extends account details component

omarCreativeDev:
- made `AccountDetails` a regular class so that angular components can properly extend and overwrite any methods
- as the relevant base markup template cannot therefore be shared [due to the above] I have refactored and leveraged a for loop to generate the markup so its a bit cleaner
- used show first form error on the form [as its a better approach]


omarCreativeDev:

src/app/features/account/account-details/account-details.ts
```diff
@@ -75,10 +69,10 @@ export class AccountDetailsComponent implements OnInit, OnDestroy {
 
   protected setUserDetails(): UserDetails {
     return (this.userDetails = {
-      emailAddress: this.getEmail.value,
-      fullname: this.getFullName.value,
-      jobTitle: this.getJobTitle.value,
-      contactNumber: this.getPhone.value,
+      emailAddress: this.form.get(this.formControlsMap[2].name).value,
```
ChazUK: What's the benefit of using this controls map than just a `.get('controlName')`?

omarCreativeDev: I just didnt want to have to copy the controlName string

PR #935, opened 2019-07-11
Feature/extend account details component and reuse onaccount management change your details flow

omarCreativeDev:
- extend account details component and reuse onaccount management change your details flow
- align FE with BE json response as it seems the JSON response has changed


omarCreativeDev:

PR #958, opened 2019-07-15
feature | create user account | remove conditionally revealing primar…

omarCreativeDev:
- remove conditionally revealing primary account checkbox
- created array of radio control values, new model and then looped over them in html

### DIFF
--- a/src/app/core/model/form-controls.model.ts
+++ b/src/app/core/model/form-controls.model.ts
@@ -1,0 +1,4 @@
+export interface RadioField {
+  value: string;
+  label: string;
+}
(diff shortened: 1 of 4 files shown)


omarCreativeDev:

src/app/core/model/form-controls.model.ts
```diff
@@ -0,0 +1,4 @@
+export interface RadioField {
```
ChazUK: If you're going to create an interface for Radio's then can you add it across the site, or double check to see if Radio's all follow the same pattern.

This would be something that could benefit from being in Design Guide

omarCreativeDev: Just renamed this interface to `RadioFieldData`. Have added a task to go ahead and make this change globally

PR #961, opened 2019-07-15
Minor refactor to checkFiles validator to work in IE11

tgreyuk:


tgreyuk:
[Select/Upload files is giving an error that 'between 2 and 3 files' must be uploaded - IE11](https://trello.com/c/WRYqqR1n/245-select-upload-files-is-giving-an-error-that-between-2-and-3-files-must-be-uploaded-ie11)

omarCreativeDev:
awesome and clean fix especially as the dodgy timeout I originally put in has been removed too!

PR #969, opened 2019-07-16
Bulk upload use local identifiers

wozzer72:
Predominantly, this is all about effecting use upon local identifiers replacing previous dependency on Establishment::name and Worker::nameOrId for uniqueness.

I didn't realise how many occurrences would be impacted.

But I have done full testing here on upload and downloads, on new, updated and deleted establishments, or new, updated and deleted workers.

I've changed Local Identifiers and checked download and impact on upload. I've changed Local Idenitifier in the bulk upload CSV for worker and establishment, and confirmed the results of the upload.

All is looking good and repeatable.

Note - although I tried to include CHANGEWORKID in the scope of these changes, alas, rather than updating the worker, the worker is deleted. I've moved the associated card back to backend development.

Other fixed also included because they were stopping the import:

1. I've downgraded PRVID and ADDRESS1 errors to warnings; there are cards in the backlog to rectify these.
2. I've corrected OTHERJOBS validation errors and also allowed for it to be optional.
3. I've fixed type mismatch comparison on sick days.

### DIFF
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -295,15 +295,17 @@ class Establishment {
         errCode: Establishment.ADDRESS_ERROR,
         errType: `ADDRESS_ERROR`,
         error: 'First line of address (ADDRESS1) must be defined',
-        source: myAddress1,
         name: this._currentLine.LOCALESTID,
       });
     } else if (myAddress1.length > MAX_LENGTH) {
       localValidationErrors.push({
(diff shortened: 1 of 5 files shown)


wozzer72:
[Bulk Upload onload/current entity association to use new Local Identifiers - backend only](https://trello.com/c/9Uk1Lbr2/209-bulk-upload-onload-current-entity-association-to-use-new-local-identifiers-backend-only)

PR #979, opened 2019-07-17
Changing the Worker's local identifier only on saving the worker to m…

wozzer72:
…inimise the impact of what is effectivekly a primary key in bulk upload (immitable).


wozzer72:
[Bulk upload - prevalidate doesn't recognise a workers file without the CHGUNIQUEWORKERID column](https://trello.com/c/HzvYKsJW/316-bulk-upload-prevalidate-doesnt-recognise-a-workers-file-without-the-chguniqueworkerid-column)

PR #982, opened 2019-07-17
Implement delete user account dialog and associated methods

tgreyuk:


tgreyuk:
[(6) User Accounts - Delete User Accounts](https://trello.com/c/7cUmofEk/184-6-user-accounts-delete-user-accounts)

src/app/core/services/user.service.ts
```diff
@@ -30,8 +30,11 @@ export class UserService {
   /*
    * GET /api/user/establishment/:establishmentId/:username
    */
-  public getUserDetails(username): Observable<UserDetails> {
-    return this.http.get<any>(`/api/user/establishment/${this.establishmentService.establishmentId}/${username}`);
+  public getUserDetails(
+    username: string,
+    establishmentuid = this.establishmentService.establishmentId.toString()
```
omarCreativeDev: should we be using the establishment uid instead?

src/app/features/workplace/view-user-account/view-user-account.component.html
```diff
@@ -1,3 +1,13 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <span class="govuk-caption-xl">User account</span>
+      {{ user.fullname }}
+    </h1>
+  </div>
+  <div class="govuk-grid-column-one-third">
+    <button *ngIf="canDeleteUser" class="govuk-button govuk-button--link" (click)="onDeleteUser()">
+      <img src="/assets/images/bin.svg" />Delete user account
```
omarCreativeDev: does the svg need a lil space between the copy?

src/app/core/services/user.service.ts
omarCreativeDev: update, just read further down that uid is not exposed so fair enough ignore my last comment : )

tgreyuk: Yes but this method is used elsewhere and i didn't want to change that with this. Basically i have set it up to take a paramter and also preserve current behavior.

src/app/features/workplace/view-user-account/view-user-account.component.html
tgreyuk: Yeah good point and i'm looking at this. I might even make this into a component as it also looks wrong on delete workspace.

ChazUK: I've got a task to update the Button here, so you can leave it and hopefully I'll get round to it tonight

PR #994, opened 2019-07-17
changes

dbjsdev:


server/routes/accounts/user.js
```diff
@@ -65,12 +73,18 @@ router.route('/establishment/:id/:userId').get(async (req, res) => {
 
     try {
         if (await thisUser.restore(byUUID, byUsername, showHistory && req.query.history !== 'property')) {
-            return res.status(200).json(thisUser.toJSON(showHistory, showPropertyHistoryOnly, showHistoryTime, false));
+            let userData = thisUser.toJSON(showHistory, showPropertyHistoryOnly, showHistoryTime, false);
+            if(thisUser._isPrimary) userData.isPrimary = true;
```
wozzer72: Ideally, you should update the `User::toJSON` to return isPrimary as a property. Don't modify the JSON after the fact here, because it's not then reusable elsewhere.

server/routes/accounts/user.js
```diff
@@ -33,8 +33,16 @@ router.route('/establishment/:id').get(async (req, res) => {
 
     try {
         const allTheseUsers = await User.User.fetch(establishmentId);
+
+        let data = allTheseUsers.map((user) => {
+            return Object.assign(user, { status: user.username == null ? 'Pending' : 'Active'});
```
wozzer72: Add `status` as a property inside the `User::fetch`; no need to map the results after the fact.

server/routes/accounts/user.js
```diff
@@ -33,8 +33,16 @@ router.route('/establishment/:id').get(async (req, res) => {
...
+        });
+        
         return res.status(200).json({
-            users: allTheseUsers
+            users: data.sort((a, b) => {
```
wozzer72: Also. Move the sort inside of `User::fetch` too. If you sort, then map, you'll have access to the `created` as an original Date object, thus not having map from JSON back to date.

server/routes/accounts/user.js
```diff
@@ -33,8 +33,16 @@ router.route('/establishment/:id').get(async (req, res) => {
...
+
+        let data = allTheseUsers.map((user) => {
+            return Object.assign(user, { status: user.username == null ? 'Pending' : 'Active'});
+        });
+        
         return res.status(200).json({
-            users: allTheseUsers
+            users: data.sort((a, b) => { 
+                if((a.status > b.status)) return -1; 
+                return (new Date(b.created) - new Date(a.created))
```
wozzer72: it would be better to sort on `updated` not `created`

server/models/classes/user.js
```diff
@@ -701,9 +700,19 @@ class User {
                     username: thisUser.login && thisUser.login.username ? thisUser.login.username : null,
                     created:  thisUser.created.toJSON(),
                     updated: thisUser.updated.toJSON(),
-                    updatedBy: thisUser.updatedBy
+                    updatedBy: thisUser.updatedBy,
+                    isPrimary: thisUser.isPrimary ? true : false
                 })
             });
+
+            allUsers = allUsers.map((user) => {
```
wozzer72: has this been tested? the right hand side `allUsers` has not yet been set.

PR #998, opened 2019-07-18
Implemented primary workplace resolver

tgreyuk:


src/app/core/resolvers/primary-workplace.resolver.ts
```diff
@@ -0,0 +1,19 @@
...
+import { catchError } from 'rxjs/operators';
+
+@Injectable()
+export class PrimaryWorkplaceResolver implements Resolve<any> {
+  constructor(private router: Router, private establishmentService: EstablishmentService) {}
+
+  resolve() {
+    return this.establishmentService.primaryEstablishment$.pipe(
+      catchError(() => {
+        this.router.navigate(['/dashboard'], { fragment: 'workplace' });
```
ChazUK: Let's just throw people back to the dashboard with no preference of which tab

src/app/core/services/establishment.service.ts
```diff
@@ -61,27 +61,27 @@ export class EstablishmentService {
...
 
-  public get establishment() {
-    return this._establishment$.value as Establishment;
+  public get primaryEstablishment() {
+    return this._primaryEstablishment$.value as Establishment;
   }
 
-  setState(establishment) {
-    this._establishment$.next(establishment);
+  setPrimaryEstablishment(establishment: Establishment) {
```
ChazUK: I think we can remove. We shouldn't be able to set the primary establishment anywhere except in this service.

and in your `get primaryEstablishment$()` function just call `this._primaryEstablishment$.next(establishment);`

src/app/core/services/establishment.service.ts
```diff
@@ -61,27 +61,27 @@ export class EstablishmentService {
...
+    this._primaryEstablishment$.next(establishment);
   }
 
   public resetState() {
```
ChazUK: When would we want to reset the Primary Establishment?

ChazUK: Ahh I see, on the Auth Service when logging out, ignore this comment

src/app/features/dashboard/dashboard.component.ts
```diff
@@ -1,28 +1,21 @@
...
-    private authService: AuthService,
-    private userService: UserService
-  ) {}
+  constructor(private authService: AuthService, private userService: UserService, private route: ActivatedRoute) {}
 
   ngOnInit() {
-    this.establishmentService.establishment$.pipe(take(1)).subscribe(establishment => {
-      this.establishment = establishment;
-    });
+    this.establishment = this.route.snapshot.data.establishment;
```
ChazUK: On the Dashboard we'll always be pulling the Primary Establishment, so shouldn't this be `this.establishment = this.establishmentService.primaryEstablishment`

It's almost like we need to bootstrap pulling the Primary Establishment on login

src/app/features/workplace/question/question.component.ts
```diff
@@ -133,7 +133,7 @@ export class Question implements OnInit, OnDestroy {
   protected onSuccess(): void {}
 
   protected _onSuccess(data) {
-    this.establishmentService.setState({ ...this.establishment, ...data });
+    this.establishmentService.setPrimaryEstablishment({ ...this.establishment, ...data });
```
ChazUK: This is the main issue with Resolvers that I can see, we have to pull the data from the route which can't then be updated :(

PR #999, opened 2019-07-18
Introducting AWS secret configs for preprod.

nasirmids:
Hi Warren,

Here is trello card for this request https://trello.com/c/EJxfSl0p

I have introduced preproduction.yml and have following changes

- point to preprod wallet (has its own api users and other policy) - Screen shot is in Trello ticket
- point to preprod bucket (only this has auto-delete policy)
- i have turn status = false for client_ssl (there are no ssl cert for preprod need)
- I have also remove the database name and user because it comes from gov paas 

Only i dont know sns and kenisis values postfix with .prod? do you have prepro for those?

Thanks
Nasir

### DIFF
--- a/server/config/preproduction.yaml
+++ b/server/config/preproduction.yaml
@@ -1,0 +1,43 @@
+aws:
+    region: eu-west-1
+    secrets:
+        use: true
+        wallet: preprod/api
+    kinesis:
+      enabled: true
+      establishments: establishments.prod
+      workers: workers.prod
+      users: users.prod
+    sns:
+      enabled: true
+      registrations: arn:aws:sns:eu-west-1:364648107127:registrations-prod
+      feedback: arn:aws:sns:eu-west-1:364648107127:feedback-prod
+
+slack:
+    level : 3     # enables Slack notifications
+
+jwt:
+    iss: 'asc-wds.skillsforcare.org.uk'
+    ttl:
+        login: 5
+
+db:
+    pool: 15
+    ssl: true
+    client_ssl:
+        status: false
+        usingFiles: false
+
+notify:
+    replyTo: '44e98371-2e44-4c6b-ad76-235136be0f8a'
+    templates:
+        resetPassword: 'fd430f0d-6609-400a-b171-cf124453ec1c'
+        addUser: 'f20e9cbe-920d-4b1f-b05d-bac1ca8d5b54'
+
+bulkupload:
+    region: eu-west-2
+    bucketname: sfc-bulkupload-preprod
+
+public:
+    download:
+        baseurl: https://sfc-public-prod.s3.eu-west-2.amazonaws.com/public/download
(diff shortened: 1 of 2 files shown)


server/config/preproduction.yaml
```diff
@@ -0,0 +1,43 @@
+aws:
+    region: eu-west-1
+    secrets:
+        use: true
+        wallet: preprod/api
+    kinesis:
+      enabled: true
+      establishments: establishments.prod
```
wozzer72: We don't want these going to prod - this is preprod

server/config/preproduction.yaml
```diff
@@ -0,0 +1,43 @@
...
+      workers: workers.prod
+      users: users.prod
+    sns:
+      enabled: true
+      registrations: arn:aws:sns:eu-west-1:364648107127:registrations-prod
```
wozzer72: again. this is not prod

server/config/preproduction.yaml
```diff
@@ -0,0 +1,43 @@
...
+db:
+    pool: 15
+    ssl: true
+    client_ssl:
+        status: false
+        usingFiles: false
+
+notify:
+    replyTo: '44e98371-2e44-4c6b-ad76-235136be0f8a'
+    templates:
```
wozzer72: Is this production notify?

server/config/preproduction.yaml
```diff
@@ -0,0 +1,43 @@
...
+        resetPassword: 'fd430f0d-6609-400a-b171-cf124453ec1c'
+        addUser: 'f20e9cbe-920d-4b1f-b05d-bac1ca8d5b54'
+
+bulkupload:
+    region: eu-west-2
+    bucketname: sfc-bulkupload-preprod
+
+public:
+    download:
+        baseurl: https://sfc-public-prod.s3.eu-west-2.amazonaws.com/public/download
```
wozzer72: I've happy for this to be prod - it's just guidance docs

server/config/preproduction.yaml
nasirmids: Yes this is copy from production.yaml.  do we have a pre prod one?  
besides the in AWS secret notify key is correct and point to test.

server/config/preproduction.yaml
nasirmids: Just to confirm Warren this is preprod config file. 

I copy it from production yaml; should i delete from there?

server/config/preproduction.yaml
nasirmids: ok

server/config/preproduction.yaml
nasirmids: i have replced them with preprod after looking at aws kinesis and sns

server/config/preproduction.yaml
nasirmids: I have replaced this config with Test config values

PR #1001, opened 2019-07-18
Bulk upload reasons for leaving

wozzer72:
https://trello.com/c/vNHbfdms - Adding an interim solution for reasons for leaving for bulk upload - what goes in, comes out


wozzer72:
[(2) Bulk Upload - Staff Records - Reasons for leaving - back end](https://trello.com/c/vNHbfdms/95-2-bulk-upload-staff-records-reasons-for-leaving-back-end)

PR #1003, opened 2019-07-18
Feature/fix create user guard

omarCreativeDev:
fix create user and role guards

### DIFF
--- a/src/app/core/guards/create-user/create-user.guard.ts
+++ b/src/app/core/guards/create-user/create-user.guard.ts
@@ -1,12 +1,9 @@
-import { Injectable } from '@angular/core';
 import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot } from '@angular/router';
-import {
-  ValidateAccountActivationTokenRequest,
-  ValidateAccountActivationTokenResponse,
-} from '@core/model/account.model';
-import { CreateAccountService } from '@core/services/create-account/create-account.service';
-import { Observable, of } from 'rxjs';
 import { catchError, map } from 'rxjs/operators';
+import { CreateAccountService } from '@core/services/create-account/create-account.service';
+import { Injectable } from '@angular/core';
+import { Observable, of } from 'rxjs';
+import { ValidateAccountActivationTokenRequest } from '@core/model/account.model';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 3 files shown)


src/app/core/guards/create-user/create-user.guard.ts
```diff
@@ -1,37 +1,33 @@
...
     const requestPayload: ValidateAccountActivationTokenRequest = {
-      uuid: route.params.activationToken
+      uuid: route.params.activationToken,
     };
 
-    this.createAccountService.validateAccountActivationToken(requestPayload)
-      .pipe(take(1))
-      .subscribe((response: ValidateAccountActivationTokenResponse) => {
+    return this.createAccountService.validateAccountActivationToken(requestPayload).pipe(
+      map((response: ValidateAccountActivationTokenResponse) => {
```
ChazUK: I think this is automatically typed due to the return typing of `validateAccountActivationToken`

src/app/core/guards/create-user/create-user.guard.ts
```diff
@@ -1,37 +1,33 @@
...
-        this.validToken = true;
-      });
-
-    if (this.validToken) {
-      return true;
-    }
-
-    this.router.navigate(['/dashboard']);
-    return false;
+        return true;
```
ChazUK: I guess this makes sense as you are just mapping the value that will be used as an Observable 👍 

Sorry I missed that.

src/app/core/guards/role/role.guard.ts
```diff
@@ -1,33 +1,29 @@
...
-
-    if (allowedRoles.includes(this.role)) {
-      return true;
-    }
-
-    this.router.navigate(['/dashboard']);
-    return false;
+    return this.authService.auth$.pipe(
+      map((response: LoggedInSession) => {
+        if (allowedRoles.includes(response.role)) {
```
ChazUK: What happens on the else? This would could be better written as 

```map((response: LoggedInSession) => allowedRoles.includes(response.role))```

src/app/core/guards/create-user/create-user.guard.ts
omarCreativeDev: removed

src/app/core/guards/role/role.guard.ts
omarCreativeDev: done

PR #1005, opened 2019-07-18
Admin role with no primary establishment

wozzer72:
https://trello.com/c/sEjGGebv
In preparation for work on the Quick 'n' Dirty Search with Bryce tomorrow, the backend needs to allow for successful login on admin roles only where there is no primary (home) establishment.

Whilst implementing this change to login, I have been able to update the login response to include UID and username.


wozzer72:
[Quick And Dirty Search](https://trello.com/c/sEjGGebv/268-quick-and-dirty-search)

PR #1013, opened 2019-07-18
Feature | Guard to stop last user editing permissions

ChazUK:
Adding Guard to stop users editing Permissions if they are the only user left.


ChazUK:
[(12) User Accounts - View user and edit - Front end](https://trello.com/c/UWYy8JTK/183-12-user-accounts-view-user-and-edit-front-end)

PR #1014, opened 2019-07-18
Bug Fix | Updating Local Identifiers navigation so it returns correctly

ChazUK:
This is getting close to spaghetti code. Lots of random ways to enter and exit certain parts of the App. Personally I think we should be trying to create an easier navigation structure rather than forcing the user into certain pathways.


ChazUK:
[Local Identifiers - Back Link Navigation through Staff References](https://trello.com/c/rHKnXtoU/326-local-identifiers-back-link-navigation-through-staff-references)

PR #1015, opened 2019-07-19
Adding an endpoint that allows for switching the establishment securi…

wozzer72:
…ty context for an Admin user only, but retaining the logged in user security context.

https://trello.com/c/sEjGGebv


wozzer72:
[Quick And Dirty Search](https://trello.com/c/sEjGGebv/268-quick-and-dirty-search)

PR #1017, opened 2019-07-19
Resend activationlink

dbjsdev:
PR for
https://trello.com/c/HlxXnamI/141-6-user-accounts-pending-user-resend-activation-link

Also altered the activation email timeout for registration to 2 days

### DIFF
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -114,7 +114,7 @@ router.use('/establishment/:id/:userId', Authorization.hasAuthorisedEstablishmen
 router.route('/establishment/:id/:userId').put(async (req, res) => {
     const userId = req.params.userId;
     const establishmentId = req.establishmentId;
-    const expiresTTLms = isLocal(req) && req.body.ttl ? parseInt(req.body.ttl)*1000 : 3*60*60*24*1000; // 3 days
+    const expiresTTLms = isLocal(req) && req.body.ttl ? parseInt(req.body.ttl)*1000 : 2*60*60*24*1000; // 2 days
 
     // validating user id - must be a V4 UUID or it's a username
     const uuidRegex = /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/;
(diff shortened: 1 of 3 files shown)


server/routes/accounts/user.js
```diff
@@ -434,6 +435,52 @@ router.route('/add/establishment/:id').post(async (req, res) => {
     }
 });
 
+// Resend activation link
+router.use('/:id/:trackingId', Authorization.hasAuthorisedEstablishment);
+router.route('/:id/:trackingId').post(async (req, res) => {
+    const expiresTTLms = isLocal(req) && req.body.ttl ? parseInt(req.body.ttl)*1000 : 2*60*60*24*1000; // 2 days
+
+    if (!req.params.trackingId) {
```
wozzer72: The tracking UUID should be necessary. A user (identified by the `id` being their uuid) should only have one tracking record outstanding.

wozzer72: But if you're already passing it, don't feel obliged to remove it.

server/routes/accounts/user.js
```diff
@@ -434,6 +435,52 @@ router.route('/add/establishment/:id').post(async (req, res) => {
...
+
+        if(passTokenResults){
+            const thisUser = new User.User();
+            if (await thisUser.restore(passTokenResults.user.uid, null, null)) {
+                await models.sequelize.transaction(async t => {
+                    await thisUser.trackNewUser(req.username, t, expiresTTLms);
+                });
+            }
+        }
+        return res.status(200).send("Success");
```
wozzer72: You're returning 200 (success) if there is no token found for the given tracking id? Why not 404 (not found)?

server/routes/accounts/user.js
```diff
@@ -434,6 +435,52 @@ router.route('/add/establishment/:id').post(async (req, res) => {
...
+                {
+                    model: models.user,
+                    attributes: ['id', 'uid', 'FullNameValue', 'EmailValue', 'JobTitleValue', 'PhoneValue'],
+                }
+            ]
+        });
+
+        if(passTokenResults){
+            const thisUser = new User.User();
+            if (await thisUser.restore(passTokenResults.user.uid, null, null)) {
```
wozzer72: I can't see anywhere where you're completing on the old tracking id?

server/routes/accounts/user.js
```diff
@@ -434,6 +435,52 @@ router.route('/add/establishment/:id').post(async (req, res) => {
...
+    
+    try {
+        const passTokenResults = await models.addUserTracking.findOne({
+            where: {
+                uuid: givenUuid
+            },
+            include: [
+                {
+                    model: models.user,
+                    attributes: ['id', 'uid', 'FullNameValue', 'EmailValue', 'JobTitleValue', 'PhoneValue'],
```
wozzer72: What if there is a tracking id, but it's already completed? We shouldn't be sending out a new tracking id on a completed user registration.

PR #1018, opened 2019-07-19
Allowing for the splitting of `address`, to `address1`, `address2`, `…

wozzer72:
…address3`, et al


wozzer72:
[(6) Quick And Dirty Search](https://trello.com/c/sEjGGebv/268-6-quick-and-dirty-search)

PR #1021, opened 2019-07-19
bugfix/bulk-upload-error-links-navigation

omarCreativeDev:
- this was being caused due to the hyperlink trying to link to an invalid DOM selector


omarCreativeDev:
[Bulk Upload page - Error summary links don't navigate](https://trello.com/c/E9I3cWpu/303-bulk-upload-page-error-summary-links-dont-navigate)

omarCreativeDev:
@ChazUK the `getFileId()` method is specific to the uploaded files list component as its basically for setting an id on the markup. Backend dont return anything unique like an id etc so hence the filename is the only thing that tells them apart.

PR #1023, opened 2019-07-19
Implemented resend activation link and delete end points

tgreyuk:


tgreyuk:
[(6) User Accounts - Pending User - Resend Activation Link](https://trello.com/c/HlxXnamI/141-6-user-accounts-pending-user-resend-activation-link)

src/app/core/services/user.service.ts
```diff
@@ -48,19 +48,17 @@ export class UserService {
...
 
-  /*
-   * TODO - Implement API calls once BE ready
-   */
-  public resendActivationLink(useruid: string, activationuid: string) {
-    return of({});
+  public resendActivationLink(useruid: string) {
+    const token = localStorage.getItem('auth-token');
+    const headers = new HttpHeaders({ Authorization: token });
+    return this.http.post(`api/user/${useruid}/resend-activation`, null, {
```
ChazUK: This needs to be /api/ and then you don’t need to worry about adding the Bearer Token as it’ll be handled by the Auth HTTP Interceptor

src/app/core/services/user.service.ts
```diff
@@ -48,19 +48,17 @@ export class UserService {
...
+      headers,
+      responseType: 'text' as 'json',
```
ChazUK: Same with this type

src/app/core/services/user.service.ts
tgreyuk: Well that's what I thought but it wasn't authorizing without it. I will look into it further...

src/app/core/services/user.service.ts
tgreyuk: WIthout the type - it fails to return a response - again must be an issue with the interceptor and POST or something,

tgreyuk: Ok the issue was the regx in the incerptor was blocking the interceptor as i didnt have '/' in front of api. Have removed the token.

tgreyuk: Ok the issue was the regx in the incerptor was blocking the interceptor as i didnt have '/' in front of api. Have removed the token. But i still fails without the response type here.

tgreyuk: It is because the reponse is empty. If you don't set response type as text it tries to parse it as json.

PR #1024, opened 2019-07-19
Invrease allowed read only users to 20 for button add button visibility

tgreyuk:


tgreyuk:
[Should be able to add 20 Read users to parent establishment](https://trello.com/c/VMhaKvIu/370-should-be-able-to-add-20-read-users-to-parent-establishment)

PR #1026, opened 2019-07-20
update primary

dbjsdev:
This is the update isPrimary functionality. 

Updates the user to primary and sets any other users who are currently primary to not primary

### DIFF
--- a/server/models/classes/user.js
+++ b/server/models/classes/user.js
@@ -440,6 +440,7 @@ class User {
 
                     const updateDocument = {
                         ...modifedUpdateDocument,
+                        isPrimary: this._isPrimary,
                         updated: updatedTimestamp,
                         updatedBy: savedBy.toLowerCase()
                     };
--- a/server/models/classes/user.js
+++ b/server/models/classes/user.js
@@ -437,7 +437,6 @@ class User {
                     const thisTransaction = externalTransaction ? externalTransaction : t;
 
                     if(this._isPrimary){
-                        console.log('inere')
                         // Set the existing primary to not primary
                         await models.user.update({
                                 isPrimary: false,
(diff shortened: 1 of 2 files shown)


server/routes/accounts/user.js
```diff
@@ -146,8 +146,32 @@ router.route('/establishment/:id/:userId').put(async (req, res) => {
 
             // this is an update to an existing User, so no mandatory properties!
             if (isValidUser) {
-                await thisUser.save(req.username, expiresTTLms);
 
+                await models.sequelize.transaction(async t => {
+
+                    if(thisUser._isPrimary){
```
wozzer72: This should be encapsulated within the User entity - and not within the endpoint

server/models/classes/user.js
```diff
@@ -427,19 +427,42 @@ class User {
...
                 // need to update the existing User record and add an
                 //  updated audit event within a single transaction
                 await models.sequelize.transaction(async t => {
+
                     // the saving of an User can be initiated within
                     //  an external transaction
                     const thisTransaction = externalTransaction ? externalTransaction : t;
 
+                    if(this._isPrimary){
+                        console.log('inere')
```
ChazUK: Remove console.log

server/routes/accounts/user.js
```diff
@@ -146,8 +146,11 @@ router.route('/establishment/:id/:userId').put(async (req, res) => {
 
             // this is an update to an existing User, so no mandatory properties!
             if (isValidUser) {
-                await thisUser.save(req.username, expiresTTLms);
 
+                await models.sequelize.transaction(async t => {
+                    await thisUser.save(req.username, expiresTTLms, t);
+                });
```
wozzer72: Why this is being wrapped in an external transaction. The `User::save` encapsulates it's own transaction. Just restore the code.

server/models/classes/user.js
dbjsdev: Stay comment removed

server/routes/accounts/user.js
dbjsdev: Sequelize transaction removed

PR #1029, opened 2019-07-20
Delete user

dbjsdev:
PR for delete user

https://trello.com/c/7cUmofEk/184-6-user-accounts-delete-user-accounts

### DIFF
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -512,6 +512,64 @@ router.route('/validateAddUser').post(async (req, res) => {
     }
 });
 
+router.use('/:username', Authorization.isAuthorisedAddUser);
+router.route('/:username').delete(async (req, res) => {
+   //401
+   //403
+   //200
+   //404     
+
+   try {
+        // all checks pass, so find the user using facts from the token (now on the req)
+        const login = await models.login.findOne({
+            where: {
+                username: {
+                    [models.Sequelize.Op.iLike] : req.username
+                },
+                isActive: true
+            },
+            include: [
+                {
+                    model: models.user,
+                    attributes: ['id', 'FullNameValue'],
+                }
+            ]
+        });
+
+        if (login && login.username === req.username && login.user.id) {
+            await models.sequelize.transaction(async t => {
+
+                // Set the login to not active
+                login.update({
+                    isActive: false
+                },
+                {transaction: t});
+
+                // Create audit log entry
+                const auditEvent = {
+                    userFk: login.user.id,
+                    username: req.username,
+                    type: 'deletedUser',
+                    property: 'isActive',
+                    event: {}
+                };
+                await models.userAudit.create(auditEvent, {transaction: t});
+
+                
+            });
+
+            return res.status(200).send(`User deleted ${login.user.FullNameValue}`);
+        } else {
+            throw new Error(`Failed to find user: ${req.username}`);
+        }
+
+  } catch (err) {
+    console.error('User delete failed', err);
+    return res.status(503).send();
+  }
+
+});
+
 // registers (full add) a new user - authentication middleware is specific to add user token
 router.use('/add', Authorization.isAuthorisedAddUser);
 router.route('/add').post(async (req, res) => {
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -12,6 +12,7 @@ const usernameCheck = require('../../utils/security/usernameValidation').isUsern
 
 const config = require('../../config/config');
 const loginResponse = require('../../utils/login/response');
+const uuid = require('uuid');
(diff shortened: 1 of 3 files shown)


server/routes/accounts/user.js
```diff
@@ -512,6 +513,94 @@ router.route('/validateAddUser').post(async (req, res) => {
...
+                    property: 'isActive',
+                    event: {}
+                };
+                await models.userAudit.create(auditEvent, {transaction: t});
+
+                let randomNewUsername = uuid.v4();
+
+                login.user.update({
+                    Archived: true,
+                    FullNameValue: false,
```
wozzer72: Set to null - false may work here, but it is more formal to use null to remove the value

server/routes/accounts/user.js
```diff
@@ -512,6 +513,94 @@ router.route('/validateAddUser').post(async (req, res) => {
...
+                    Username: randomNewUsername,
+                    EmailValue: '',
+                    PhoneValue: '',
+                    JobTitle: '',
+                    SecurityQuestionValue: '',
+                    SecurityQuestionAnswerValue: ''
+                },
+                {transaction: t});
+
+                await models.sequelize.query('UPDATE  cqc."EstablishmentAudit" SET "Username" = :usernameNew WHERE "Username" = :username', { replacements: { username: login.username, usernameNew: randomNewUsername },type: models.sequelize.QueryTypes.UPDATE, transaction: t });
```
wozzer72: These need to be inside the transaction, to allow for rollback.

server/routes/accounts/user.js
```diff
@@ -512,6 +513,94 @@ router.route('/validateAddUser').post(async (req, res) => {
     }
 });
 
+router.use('/:username', Authorization.isAuthorised);
+router.route('/:username').delete(async (req, res) => {
+   try {
+        const login = await models.login.findOne({
```
wozzer72: Should be restoring the user entity (new/load) - which then reuses all the encapsulated security, and filtering.

server/routes/accounts/user.js
```diff
@@ -512,6 +513,94 @@ router.route('/validateAddUser').post(async (req, res) => {
...
+                    }
+                }
+            ]
+        });
+
+        if (login && login.user.id) {
+            await models.sequelize.transaction(async t => {
+
+                // If the deleted user is the primary, make us the primary
+                if(login.user.isPrimary){
```
wozzer72: This is assumed logic. Just because I am deleting the primary I become the primary,

There should be constraint that prevents deleting a primary user; ergo - should return 400 (input validation failed).

server/routes/establishments/localIdentifier.js
```diff
@@ -179,7 +179,7 @@ router.route('/').put(async (req, res) => {
           if (thisGivenEstablishment && thisGivenEstablishment.uid && myEstablishmentUIDs.includes(thisGivenEstablishment.uid)) {
             const updateThisEstablishment = updateLocalIdOnEstablishment(thisGivenEstablishment, t, updatedTimestamp, username, allAuditEvents);
             dbUpdatePromises.push(updateThisEstablishment);
-            updatedUids.push(thisGivenEstablishment.uid);
+            updatedUids.push(thisGivenEstablishment);
```
wozzer72: What is this???? Please provide the trello card where the change to this API has been through refinement!

There is no mention about changing the local identifier response within the delete user trello card, and no other trello card has been attached to this PR.

And please confirm you have updated the API Usage document before raising this PR.

server/routes/accounts/user.js
dbjsdev: This is one to bring up Maria as they requested this on friday

server/routes/establishments/localIdentifier.js
dbjsdev: Theres no trello card for this one. This was something that was raised on Friday afternoon in a meeting

server/routes/accounts/user.js
dbjsdev: These will occur within the transaction

server/routes/accounts/user.js
wozzer72: Fine. How was I know?

server/routes/accounts/user.js
wozzer72: Ok. Too true.

server/routes/establishments/localIdentifier.js
wozzer72: No changes without trello please! How do we track these changes? Now I'm am being put in an awkward position to accept a change with no way of tracking the change. And, do I now have to review all changes against scope of cards?

wozzer72:
Approving this to get it into develop and ready for frontend

PR #1030, opened 2019-07-20
268 quick and dirty search

bwilson-ux:


src/app/features/dashboard/dashboard.component.ts
```diff
@@ -12,18 +14,45 @@ import { take } from 'rxjs/operators';
...
       this.establishment = establishment;
-    });
+    })
+
+    this.authService.auth$.pipe(take(1)).subscribe((loggedInSession: LoggedInSession) => {
+
+
+      if (loggedInSession && loggedInSession.role === 'Admin') {
+        if (!this.establishment) {
+          this.router.navigate(['/search-users']);
```
tgreyuk: Do you think the functionality is better located in a route guard?

src/app/features/dashboard/dashboard.component.ts
```diff
@@ -12,18 +14,45 @@ import { take } from 'rxjs/operators';
...
+          return false;
+        } else {
+
+          const workplaceId = localStorage.getItem('establishmentId');
+
+          this.establishmentService
```
tgreyuk: The subscription above this code block is already assigning the estabnlisment from localstorage 
```
this.establishmentService.establishment$.pipe(take(1)).subscribe(establishment => {
      this.establishment = establishment;
    })
```

src/app/features/search/search.component.ts
```diff
@@ -0,0 +1,125 @@
...
+		e.preventDefault();
+		this.getNewEstablishmentId(id).subscribe((data) => this.onSwapSuccess(data), error => this.onError(error));
+	}
+
+	public onSubmit(): void {
+		this.form.errors = [];
+		this.form.submitted = true;
+		// this.errorSummaryService.syncFormErrorsEvent.next(true);
+
+		if (this.form.username.length === 0 && this.form.name.length === 0) {
```
tgreyuk: I think implementation with Reactive Forms and built in Validators offer a more dynamic way of building form groups and controls?

src/app/core/services/establishment.service.ts
```diff
@@ -48,7 +48,7 @@ export class EstablishmentService {
   public previousEstablishmentId: number;
   public isSameLoggedInUser: boolean;
 
-  constructor(private http: HttpClient) {}
+  constructor(private http: HttpClient) { }
```
tgreyuk: It would be nice to eliminate these subtle formatting differences.

src/app/core/services/establishment.service.ts
```diff
@@ -120,6 +120,11 @@ export class EstablishmentService {
     this.returnTo$.next(returnTo);
   }
 
+  public swapEstablishmentId(value: number) {
```
tgreyuk: a setter already exists that does the same thing:
```
public set establishmentId(value: number) {
    this._establishmentId = value;
    localStorage.setItem('establishmentId', value.toString());
  }
```

src/app/core/services/auth.service.ts
```diff
@@ -60,7 +60,7 @@ export class AuthService {
   }
   public get isFirstBulkUpload(): boolean {
     if (this._session) {
-      return this._session.establishment.isFirstBulkUpload;
+      return (this._session.establishment && this._session.establishment.isFirstBulkUpload ? this._session.establishment.isFirstBulkUpload : true);
```
tgreyuk: This logic doesn't look right. `isFirstBulkLoad` will always return `true`

src/app/core/services/auth.service.ts
```diff
@@ -107,6 +107,16 @@ export class AuthService {
     this.token = token;
   }
 
+  swapToken(token: any) {
```
tgreyuk: A setter already exists that does the same thing:
```
  set token(token: string) {
    this._token = token;

    if (token) {
      localStorage.setItem('auth-token', token);
    } else {
      localStorage.removeItem('auth-token');
    }
  }
```

wozzer72:
[(6) Quick And Dirty Search](https://trello.com/c/sEjGGebv/268-6-quick-and-dirty-search)

src/app/core/services/establishment.service.ts
bwilson-ux: My IDE is being an SOB - cant yet find a way to remove this stipulation so will put this on the back burner for now.

src/app/core/services/establishment.service.ts
bwilson-ux: Using existing function now

src/app/core/services/auth.service.ts
bwilson-ux: Using existing function now.

server/routes/admin/search/users/index.js
```diff
@@ -109,7 +109,7 @@ router.route('/').post(async function (req, res) {
             nmdsId: thisUser.establishment.nmdsId,
             postcode: thisUser.establishment.postcode,
             isRegulated: thisUser.establishment.isRegulated,
-            address: thisUser.establishment.address,
+            address: thisUser.establishment.address1,
```
ChazUK: Does this effect any other places where a users address is displayed?

src/app/core/services/auth.service.ts
ChazUK: Yeah I think this will always return true, so not sure what the point of this is.

src/app/features/dashboard/dashboard.component.html
```diff
@@ -17,31 +17,32 @@ <h2 class="govuk-heading-m">Contact Skills for Care</h2>
...
   </div>
-</div>
 
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
-    <app-tabs>
-      <app-tab [title]="'Home'">
-        <app-home-tab></app-home-tab>
-      </app-tab>
+  <div class="govuk-grid-row">
```
ChazUK: From looking at this file, I don't think any changes have been made to this file, so can you remove the changes and just use the one currently in dev?

src/app/features/dashboard/dashboard.component.ts
ChazUK: Is the purpose of this to redirect an admin user directly to the Search Pages?

src/app/features/dashboard/dashboard.component.ts
```diff
@@ -3,23 +3,41 @@ import { LoggedInEstablishment } from '@core/model/logged-in.model';
...
-    private userService: UserService
-  ) {}
+    private userService: UserService,
+    private router: Router
+  ) { }
 
   ngOnInit() {
-    this.establishmentService.establishment$.subscribe(establishment => (this.establishment = establishment));
+
+    this.subscriptions.add(
```
ChazUK: Again I think this subscription has already been covered by the changes in develop at the moment.

src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -17,9 +17,11 @@ <h3 class="govuk-heading-s">Notifications</h3>
     <p>You have no new notifications</p>
     <h3 class="govuk-heading-s govuk-!-margin-top-6">Other things you can do</h3>
     <ul class="govuk-list">
-      <li *ngIf="user?.role === editRole"><a [routerLink]="['/bulk-upload']" (click)="setReturn()">Bulk upload</a></li>
-      <li *ngIf="isParent"><a [routerLink]="['/workplace', 'view-my-workplaces']">View my workplaces</a></li>
+      <li *ngIf="user?.role === editRole || user?.role === adminRole"><a [routerLink]="['/bulk-upload']" (click)="setReturn()">Bulk upload</a></li>
```
ChazUK: `[editRole, adminRole].includes(user?.role)` might be a better way to write this

src/app/features/dashboard/home-tab/home-tab.component.ts
```diff
@@ -15,6 +15,8 @@ import { take } from 'rxjs/operators';
 })
 export class HomeTabComponent implements OnInit {
   public editRole: Roles = Roles.Edit;
+  public adminRole: Roles = Roles.Admin;
+  public role: Roles;
```
ChazUK: I think this can be removed

src/app/features/login/login.component.ts
```diff
@@ -121,10 +121,13 @@ export class LoginComponent implements OnInit, OnDestroy {
       this.authService.postLogin(this.getLoginCredentials()).subscribe(
         response => {
           this.authService.updateState(response.body);
-          this.establishmentService.checkIfSameLoggedInUser(response.body.establishment.id);
 
-          // update the establishment service state with the given establishment id
-          this.establishmentService.establishmentId = response.body.establishment.id;
+          if (response.body.establishment && response.body.establishment.id) {
+            this.establishmentService.checkIfSameLoggedInUser(response.body.establishment.id);
```
ChazUK: We need this to run, so it's probably a good idea to change this to be the UserID rather than Establishment ID @omarCreativeDev

src/app/features/search/search-routing.module.ts
```diff
@@ -0,0 +1,20 @@
+import { NgModule } from '@angular/core';
+import { RouterModule, Routes } from '@angular/router';
+import { SearchComponent } from './search.component';
+
+const routes: Routes = [
+  {
```
ChazUK: I think this route needs to be updated as there's no component on the second route

```
{
  path: '',
  component: SearchComponent,
  data: { title: 'Search' },
}
```

src/app/features/search/search.component.ts
ChazUK: We'd normally use Reactive Forms for this, but happy to go with what's currently in there.

server/routes/admin/search/users/index.js
bwilson-ux: @wozzer72 would be best placed to comment?

src/app/features/dashboard/dashboard.component.html
bwilson-ux: 'workplace' is now being passed into the 'workplace-tab' as it is defined in dashboard now - saves it being called again.

src/app/features/dashboard/dashboard.component.ts
bwilson-ux: Perfect - I have pulled again from develop and seems to fit better.

src/app/features/dashboard/home-tab/home-tab.component.html
bwilson-ux: Good shout - updated :)

src/app/features/dashboard/home-tab/home-tab.component.ts
bwilson-ux: I've removed ' public role: Roles;' as adminRole gets used in the template.

src/app/features/search/search-routing.module.ts
bwilson-ux: Updated 👍

src/app/core/services/auth.service.ts
bwilson-ux: Updated to reflect original - new code with user / establishment fixes the issue I was seeing with swapping establishments. 👍

src/app/features/search/search.component.ts
bwilson-ux: Awesome - happy to review once time permits 👍

ChazUK:
Happy with the changes, went through the code with Bryce so I could understand the ins and outs of it. We'll make another ticket to clean up the Dashboard code.

PR #1031, opened 2019-07-21
Bulk upload parent subs permissions

wozzer72:
https://trello.com/c/LtgGtQho

Introduces authorisation to bulk upload (BU), as defined by the parent/sub owner/parent permissions.

BU can only work on subs it owns.

On BU Download, simply filters out those establishments it is does not owner.

But on BU upload, a validation error is required if trying to upload against a sub not owned.

Whilst implementing this I noticed some dodgy handling of null data on establishments - atypical of a newly created establishment with no properties - `entity.employerType` and `entity.serviceUsers` - fixed.

Also, whilst adding this new error, I've used "unshift" to put it to the top of all errors, and have updated accordingly missing primary and expect just one establishment validation errors too.

And whilst updating the Worker authentication, I noticed that the `workerId` was not coming through - doh. Need to add the middleware explicitly with workerId `router.use('/:workerId', validateWorker);` before adding it on "/".

### DIFF
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -51,6 +51,7 @@ class Establishment {
   static get EXPECT_JUST_ONE_ERROR() { return 950; }
   static get MISSING_PRIMARY_ERROR() { return 955; }
 
+  static get NOT_OWNER_ERROR() { return 997; }
   static get DUPLICATE_ERROR() { return 998; }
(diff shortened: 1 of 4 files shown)


wozzer72:
[(3) Bulk Upload Parent/Sub Permissions](https://trello.com/c/LtgGtQho/302-3-bulk-upload-parent-sub-permissions)

PR #1041, opened 2019-07-22
feature/add-missing-screen

omarCreativeDev:
- add missing change your details screen to activate account flow

### DIFF
--- a/src/app/core/constants/constants.ts
+++ b/src/app/core/constants/constants.ts
@@ -10,5 +10,5 @@ export const FLOAT_PATTERN = /^([0-9]*[.])?[0-9]+$/;
 export const API_PATTERN = /^\/api\//;
 export const ALPHA_NUMERIC_WITH_HYPHENS_UNDERSCORES = /^[A-Za-z0-9_-]*$/;
 export const FILE_UPLOAD_TYPES = ['CSV', 'ZIP'];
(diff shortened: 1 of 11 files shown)


omarCreativeDev:
[(10) User Accounts - Activate Account](https://trello.com/c/CjBwDiN1/142-10-user-accounts-activate-account)

src/app/core/constants/constants.ts
```diff
@@ -10,3 +10,5 @@ export const FLOAT_PATTERN = /^([0-9]*[.])?[0-9]+$/;
 export const API_PATTERN = /^\/api\//;
 export const ALPHA_NUMERIC_WITH_HYPHENS_UNDERSCORES = /^[A-Za-z0-9_-]*$/;
 export const FILE_UPLOAD_TYPES = ['CSV', 'ZIP'];
+export const EMAIL_PATTERN = /^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+.[A-Za-z]{2,4}$/;
```
ChazUK: Where did you get this email pattern from? domain TLD's can be longer than 4 characters like `.healthcare`

omarCreativeDev: this is what we always had in the code but i thought it would be a good idea to extract to a constant. any suggestions on a sensible length for domain TLD?

omarCreativeDev: @ChazUK just made it 63 chars long which i believe it the max for a TLD

PR #1042, opened 2019-07-22
Bugfix/parent add user limits

tgreyuk:


src/app/features/dashboard/dashboard.component.html
```diff
@@ -31,7 +31,7 @@ <h2 class="govuk-heading-m">Contact Skills for Care</h2>
         </app-tab>
 
         <app-tab [title]="'User accounts'">
-          <app-user-accounts-summary *ngIf="workplace" [workplace]="workplace"></app-user-accounts-summary>
```
ChazUK: There are some users that may not have a workplace (admin users) So maybe `User accounts`, `Workplace` and `Staff Records` tabs could have an `*ngIf="workplace"` on them

tgreyuk: have added to wrapping div?

PR #1046, opened 2019-07-22
Double check on firstBulkUpload

bwilson-ux:
This places a check on the session and the establishment before returning a possible null value.


src/app/features/dashboard/dashboard.component.ts
```diff
@@ -29,6 +29,7 @@ export class DashboardComponent implements OnInit {
 
     this.subscriptions.add(
       this.userService.loggedInUser$.subscribe(user => {
+        console.log('WP', this.workplace);
```
wozzer72: Can you remove the console log?

bwilson-ux: already done :D

PR #1059, opened 2019-07-24
resend tracking return

dbjsdev:
https://trello.com/c/74MWmm6O/155-user-accounts-resend-activation-link-isnt-responding-with-the-trackinguuid-back-end

resend activation email now altered so that it returns back the trackingUUID as per the card above

### DIFF
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -482,8 +482,12 @@ router.route('/:uid/resend-activation').post(async (req, res) => {
                 await models.sequelize.transaction(async t => {
                     trackingUUID = await thisUser.trackNewUser(req.username, t, expiresTTLms);
                 });
+                let response = {};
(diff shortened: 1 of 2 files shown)


server/routes/accounts/user.js
```diff
@@ -478,15 +478,17 @@ router.route('/:uid/resend-activation').post(async (req, res) => {
         if(passTokenResults){
             const thisUser = new User.User();
             if (await thisUser.restore(passTokenResults.user.uid, null, null)) {
+                let trackingUUID = "";
                 await models.sequelize.transaction(async t => {
-                    await thisUser.trackNewUser(req.username, t, expiresTTLms);
+                    trackingUUID = await thisUser.trackNewUser(req.username, t, expiresTTLms);
```
wozzer72: Can't have this being returned on an API response, unless is it in dev only!

PR #1061, opened 2019-07-24
Redirecting the user on error of activating account

ChazUK:


ChazUK:
[(2) User Accounts - Return error - back end](https://trello.com/c/f3ZTNRq9/138-2-user-accounts-return-error-back-end)

src/app/features/activate-user-account/confirm-account-details/confirm-account-details.component.ts
```diff
@@ -119,11 +119,19 @@ export class ConfirmAccountDetailsComponent extends ConfirmAccountDetails {
         .activateAccount(this.generatePayload())
         .subscribe(
           () => this.router.navigate(['/activate-account', this.activationToken, 'complete']),
-          (error: HttpErrorResponse) => this.onError(error)
+          (error: HttpErrorResponse) => this._onError(error)
         )
     );
   }
 
+  private _onError(error) {
```
omarCreativeDev: why prefix with an _?

omarCreativeDev: oh actually ignore me, just read the _onError block properly and you are still calling the onError() method in the base class! cool

ChazUK: Because the class it extends already has an onError, this is just to extend that function.

PR #1063, opened 2019-07-24
update permissions check

dbjsdev:
Read/Edit permissions which exists on add, now also implemented on Update user

As per card

https://trello.com/c/AVl2iOR4/139-view-edit-accounts-issues

### DIFF
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -129,6 +129,29 @@ router.route('/establishment/:id/:userId').put(async (req, res) => {
         byUsername = escape(userId.toLowerCase());
     }
 
+    // ensure only a user having the role of Edit can update a user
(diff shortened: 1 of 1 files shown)


server/routes/accounts/user.js
```diff
@@ -129,6 +129,29 @@ router.route('/establishment/:id/:userId').put(async (req, res) => {
         byUsername = escape(userId.toLowerCase());
     }
 
+    // ensure only a user having the role of Edit can update a user
+    if (!(req.role && req.role === 'Edit')) {
```
ChazUK: I think this needs to be Edit and Admin users

ChazUK: @wozzer72 Or shall we leave it as Edit for the time being?

wozzer72:
How does this change relates to the trello card?

PR #1070, opened 2019-07-24
Bug Fix | Delete Read Users

ChazUK:


ChazUK:
[Delete link not shown for read only account after trying to delete last edit user](https://trello.com/c/14pTsY2g/208-delete-link-not-shown-for-read-only-account-after-trying-to-delete-last-edit-user)

PR #1071, opened 2019-07-24
Bug Fix | User Accounts Issues

ChazUK:
Users can now delete primary users
When a primary user is deleted the logged in user then becomes primary
Users get an error when they try to change a users permissions to a role that is already maxed out


ChazUK:
[View/Edit Accounts issues](https://trello.com/c/AVl2iOR4/139-view-edit-accounts-issues)

PR #1075, opened 2019-07-25
Adding exclusion on edit permissions for Worker auithorisation if the…

wozzer72:
… current parent owns the given establishment - and correcting the error message if indeed not write access is given.

https://trello.com/c/llLXnIsv


wozzer72:
[JWT secret password reset](https://trello.com/c/llLXnIsv/149-jwt-secret-password-reset)

PR #1077, opened 2019-07-25
Bugfix/account activation create username save and return navigation

omarCreativeDev:
- correct create-username's screens save and continue/return route
- fix broken route path
- some cleanup, remove unnessaary extra previousAndReturnRoute property


omarCreativeDev:
[Changing data from Confirm account details, the CTAs on the username and security pages should be [Save and Return]](https://trello.com/c/jK3NPr8P/207-changing-data-from-confirm-account-details-the-ctas-on-the-username-and-security-pages-should-be-save-and-return)

PR #1078, opened 2019-07-25
Refactoring old promise.then code when resolving establishment.id whe…

wozzer72:
…n req.param.id is an establishment UID to use await.


wozzer72:
[console errors in sfcdev - Warning: a promise was created in a handler at](https://trello.com/c/knhc7UAX/221-console-errors-in-sfcdev-warning-a-promise-was-created-in-a-handler-at)

PR #1081, opened 2019-07-25
Feature | Parent & Subs | Add Staff Records

ChazUK:
I still need to work on the back routing, but this is a good start


ChazUK:
[(3) Parent and Subsidiaries - Add Workers](https://trello.com/c/oGW3inOf/53-3-parent-and-subsidiaries-add-workers)

PR #1082, opened 2019-07-25
bugfix | workplace references caching

omarCreativeDev:
remove caching block as if user logs out and a different user signs in > the previous data is cached and returned


omarCreativeDev:
[Local identifiers displayed are not those just updated/saved in the database - front end](https://trello.com/c/3PzbytDg/141-local-identifiers-displayed-are-not-those-just-updated-saved-in-the-database-front-end)

PR #1083, opened 2019-07-25
Can't delete primary user

ChazUK:


ChazUK:
[Delete user accounts - primary should not be allowed to be deleted](https://trello.com/c/bkIXF7MI/214-delete-user-accounts-primary-should-not-be-allowed-to-be-deleted)

PR #1086, opened 2019-07-25
setup

dbjsdev:
https://trello.com/c/jP2FroEU/205-account-confirmation-is-creating-a-second-user-record

This is another issue that came up regarding the registration making duplicate accounts.

When the restore adding in before the load. It meant that the code in the "create new" section of the save method was no longer running and instead an update running. This mean the initial setup code in create such as making the login entry, removing token etc was not running. 

I have now added these into the update section of the code - if there is no existing login. This may not be ideal as perhaps there are circumstances where a user is updated which has no login? - however with my limited knowledge for now I have determined this to be a means of checking if it is a new account

### DIFF
--- a/server/models/classes/user.js
+++ b/server/models/classes/user.js
@@ -291,7 +291,7 @@ class User {
 
     // saves the User to DB. Returns true if saved; false if not.
     // Throws "UserSaveException" on error
-    async save(savedBy, ttl=0, externalTransaction=null) {
+    async save(savedBy, ttl=0, externalTransaction=null, firstSave=false) {
         let mustSave = this._initialise();
 
         if (!this.uid) {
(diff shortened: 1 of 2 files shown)


server/models/classes/user.js
```diff
@@ -443,6 +443,61 @@ class User {
                     //  an external transaction
                     const thisTransaction = externalTransaction ? externalTransaction : t;
 
+                    // Is this the intial update setup
+                    let loginCount = await models.login.count({ where: {
+                            registrationId: this._id
+                    }});
+                    
+                    if(loginCount == 0){
```
wozzer72: If you predicate this using a parameter that is passed into the save, then you have a more absolute measure upon which to update.

dbjsdev: changes applied

PR #1088, opened 2019-07-25
Prevent deletion of primary on BE

dbjsdev:
Added in a check to prevent the deletion of the primary as requested in:

https://trello.com/c/bkIXF7MI/214-delete-user-accounts-primary-should-not-be-allowed-to-be-deleted

### DIFF
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -615,6 +615,10 @@ router.route('/establishment/:id/:userid').delete(async (req, res) => {
                 return res.status(400).send('Cannot delete own user account');
             }
 
+            if(thisUser._isPrimary){
+                return res.status(400).send('Cannot delete primary account');
+            }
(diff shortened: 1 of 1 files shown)


server/routes/accounts/user.js
```diff
@@ -615,6 +615,10 @@ router.route('/establishment/:id/:userid').delete(async (req, res) => {
                 return res.status(400).send('Cannot delete own user account');
             }
 
+            if(thisUser._isPrimary){
+                return res.status(400).send('Cannot delete primary account');
```
wozzer72: This would be better on the frontend devs if returning JSON. I know Angular is an ass when returning not JSON responses. Check with @ChazUK

wozzer72: Charlie confirmed this is OK in sopra-development channel.

PR #1095, opened 2019-07-26
404

dbjsdev:
https://trello.com/c/otMVHWYN/224-where-account-is-deleted-before-activation-completes-a-500-error-code-should-be-returned

As agreed with FE - /validateAddUser will now return a 404 if the token has been completed

### DIFF
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -552,15 +552,8 @@ router.route('/validateAddUser').post(async (req, res) => {
             ]
         });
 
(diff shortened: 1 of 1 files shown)


server/routes/accounts/user.js
```diff
@@ -561,11 +561,6 @@ router.route('/validateAddUser').post(async (req, res) => {
                 return res.status(403).send();
```
wozzer72: Shouldn't this be 404 too?

dbjsdev: amended

PR #1099, opened 2019-07-26
Bulk upload statuses - validations

wozzer72:
https://trello.com/c/1pTd8zzJ

This first cut of code covers just the validation of the statuses for both establishments and workers, cross checking for existing or missing establishments/workers of the current state.

It also includes the creation of skeletal establishment and worker entities by updating the entities' `hasMandatoryProperties` and `isValid` to short cicuit if status is "UNCHECKED".

I've doubled checked the intermediatry `all.entities.json`, all establishments and their associated workers are accurately represented including holding the original "status", which will be required for subsequent processing.

I've also double checked the validation `difference.report.json`, and this is still showing the correct set of new, updated and deleted establishments and workers.

### DIFF
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -1,9 +1,12 @@
 const BUDI = require('../BUDI').BUDI;
 
 class Establishment {
-  constructor(currentLine, lineNumber) {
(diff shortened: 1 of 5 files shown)


wozzer72:
[(5) Bulk Upload - Status](https://trello.com/c/1pTd8zzJ/217-5-bulk-upload-status)

PR #1106, opened 2019-07-26
bugfix | dont show logout or account links in header on account activ…

omarCreativeDev:
…ation flow


omarCreativeDev:
[User activation flow implies user is already logged in](https://trello.com/c/lnezDfZH/223-user-activation-flow-implies-user-is-already-logged-in)

PR #1107, opened 2019-07-26
Bugfix/logout user if they start account activation process and if they are logged in

omarCreativeDev:
- wrap common auth service code in its own method
- if user is on account activation flow clear localstorage, session and auth token 
- prettify auth-interceptor file [forgot to do this in last pr as i had some conflicts]


omarCreativeDev:
[User activation flow implies user is already logged in](https://trello.com/c/lnezDfZH/223-user-activation-flow-implies-user-is-already-logged-in)

PR #1109, opened 2019-07-28
Bulk upload statuses 2

wozzer72:
https://trello.com/c/1pTd8zzJ

This PR is the second of three and follows on from the initial validation PR.

Initial test observations from Ann and confirmed by Jackie requested short circuiting validation if status is DELETE, in addition to the existing UNCHANGED.

But the focus of this PR is on the revised difference report, which is used as input on completion to work through creating, updated and deleting establishments and their workers.

By realigning those establishments and workers declared as deleted is the first step to processing based on status.


wozzer72:
[(5) Bulk Upload - Status](https://trello.com/c/1pTd8zzJ/217-5-bulk-upload-status)

PR #1114, opened 2019-07-29
Update to fix mispell of nmsdId code

bwilson-ux:
Silly mispelling of the code needed to display.


wozzer72:
[Internal Admin - Is subsidiary flag](https://trello.com/c/7VE7j0xH/212-internal-admin-is-subsidiary-flag)

PR #1115, opened 2019-07-29
Passing correct Workplace UID to the User Service

ChazUK:


ChazUK:
[View/Edit Accounts issues](https://trello.com/c/AVl2iOR4/139-view-edit-accounts-issues)

PR #1120, opened 2019-07-29
Feature/create base class for find workplace address

omarCreativeDev:
- create base class for find workplace address
- extend and use on registration and add-workplace flows


omarCreativeDev:
[(5) Parent and subsidiaries - Add a workplace - Non - CQC Regulated](https://trello.com/c/bcbbFepb/200-5-parent-and-subsidiaries-add-a-workplace-non-cqc-regulated)

PR #1121, opened 2019-07-29
Feature/create base class for select main service

omarCreativeDev:
- create base class for select main service
- extend and use on registration flow


omarCreativeDev:
[(5) Parent and subsidiaries - Add a workplace - Non - CQC Regulated](https://trello.com/c/bcbbFepb/200-5-parent-and-subsidiaries-add-a-workplace-non-cqc-regulated)

PR #1129, opened 2019-07-30
Removing Name and Username from being displayed on Reset Password page

ChazUK:


ChazUK:
[SECURITY - Password Reset - should not show "username" on the enter new password screen](https://trello.com/c/jOcPJsrO/229-security-password-reset-should-not-show-username-on-the-enter-new-password-screen)

PR #1130, opened 2019-07-30
Correct Exit for User Accounts

ChazUK:


ChazUK:
[Exit from a user account Permissions page - on a sub - is navigating back to user accounts for the parent](https://trello.com/c/pk5Q3mFk/232-exit-from-a-user-account-permissions-page-on-a-sub-is-navigating-back-to-user-accounts-for-the-parent)

PR #1133, opened 2019-07-30
Correcting spelling mistake

wozzer72:
https://trello.com/c/uduQlVES


wozzer72:
[Add Workplace - fails validation is "IsRegulated" is true](https://trello.com/c/uduQlVES/259-add-workplace-fails-validation-is-isregulated-is-true)

PR #1135, opened 2019-07-30
Conditionally displaying number of remaining workplaces

ChazUK:


ChazUK:
[(1) Local Identifiers - Remove 0 workplaces remaining tag](https://trello.com/c/RwLmSqfk/143-1-local-identifiers-remove-0-workplaces-remaining-tag)

PR #1136, opened 2019-07-30
permissions

dbjsdev:
https://trello.com/c/4NczPkjf/136-13-site-wide-read-only-user-permissions

Here are the modifications to the permissions.

Altered workers so that it READ role is denied
Altered the isAuthenticated middleware so that there are now two possible options, one implementing role check, one not.
User accounts router altered so that the correct middleware is applied to each route

### DIFF
--- a/server/models/classes/user.js
+++ b/server/models/classes/user.js
@@ -1108,7 +1108,7 @@ class User {
             // now, if the primary establishment is a parent
             //  and if the user's role against their primary parent is Edit
             //  fetch all other establishments associated with this parent
-            if ((myRole === 'Edit' || myRole === 'Admin') && isParent) {
+            if (isParent) {
                 // get all subsidaries associated with this parent
                 allSubResults = await models.establishment.findAll({
                     attributes: ['uid', 'isParent', 'dataOwner', 'parentUid', 'LocalIdentifierValue', 'parentPermissions', 'NameValue', 'updated'],
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -31,7 +31,7 @@ const responseErrors = {
 };
 
 // returns a list of all users for the given establishment
-router.use('/establishment/:id', Authorization.hasAuthorisedEstablishment);
+router.get('/establishment/:id', Authorization.hasAuthorisedEstablishment);
 router.route('/establishment/:id').get(async (req, res) => {
     // although the establishment id is passed as a parameter, get the authenticated  establishment id from the req
     const establishmentId = req.establishmentId;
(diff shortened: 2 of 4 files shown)


server/models/classes/user.js
```diff
@@ -1108,7 +1108,7 @@ class User {
             // now, if the primary establishment is a parent
             //  and if the user's role against their primary parent is Edit
             //  fetch all other establishments associated with this parent
-            if ((myRole === 'Edit' || myRole === 'Admin') && isParent) {
+            if (isParent) {
```
wozzer72: And what if the user's role when logged in with a primary establishment of parent is read only?

server/routes/accounts/user.js
```diff
@@ -586,7 +586,7 @@ router.route('/validateAddUser').post(async (req, res) => {
     }
 });
 
-router.use('/establishment/:id/:userid', Authorization.hasAuthorisedEstablishment);
+router.delete('/establishment/:id/:userid', Authorization.hasAuthorisedEstablishment);
```
wozzer72: This needs to be "use" if wanting to attach middleware to the endpoint!!!!!

server/routes/accounts/user.js
```diff
@@ -108,13 +108,13 @@ router.route('/me').get(async (req, res) => {
...
 // optional parameter - "history" must equal 1
-router.use('/establishment/:id/:userId', Authorization.hasAuthorisedEstablishment);
+router.get('/establishment/:id/:userId', Authorization.hasAuthorisedEstablishment);
 router.route('/establishment/:id/:userId').get(async (req, res) => {
     getUser(req, res);
 });
 
 // updates a user with given uid or username
-router.use('/establishment/:id/:userId', Authorization.hasAuthorisedEstablishment);
+router.put('/establishment/:id/:userId', Authorization.hasAuthorisedEstablishmentAllowAllRoles);
```
wozzer72: I like this!

server/models/classes/user.js
dbjsdev: According to Business, Read only users of the parent should see the sub data

server/routes/accounts/user.js
dbjsdev: This is so that the specific middleware occurs on that particular method and path

server/models/classes/user.js
wozzer72: See sub data yes, but this is against a User.put endpoint - so an update.

server/routes/accounts/user.js
wozzer72: Not my preferred approach, but you have confirmed this is tested. So I'll allow this.

PR #1144, opened 2019-07-31
fix

omarCreativeDev:


omarCreativeDev:
[User with read-only permissions sees an active 'Add a new workplace' button in My workplaces](https://trello.com/c/yT2imxw3/273-user-with-read-only-permissions-sees-an-active-add-a-new-workplace-button-in-my-workplaces)

PR #1146, opened 2019-07-31
Addinga new "localIdentfiiers" (note the plural on endpoint name) tha…

wozzer72:
…t retunrs a true or false response

https://trello.com/c/h2dX27BB


wozzer72:

PR #1151, opened 2019-07-31
Feature | Read Only View WDF

ChazUK:


ChazUK:
[User with read-only permissions should see the workplace WDF data](https://trello.com/c/1mxK99E3/274-user-with-read-only-permissions-should-see-the-workplace-wdf-data)

PR #1153, opened 2019-08-01
bugfix | account activation redirect to problem with page on 404 error

omarCreativeDev:


omarCreativeDev:
[Where account is deleted before activation completes, a 500 error code should be returned](https://trello.com/c/otMVHWYN/224-where-account-is-deleted-before-activation-completes-a-500-error-code-should-be-returned)

PR #1154, opened 2019-08-01
Bug Fix | Basic Records Error Message

ChazUK:


ChazUK:
[(1) Staff basic details; Wrong message shown when contract type not selected](https://trello.com/c/TTfeRpvW/51-1-staff-basic-details-wrong-message-shown-when-contract-type-not-selected)

PR #1161, opened 2019-08-01
Updating route to add another staff record

ChazUK:


ChazUK:
[After completing a new worker, [Add another staff record] fails - Page not found - regression issue](https://trello.com/c/weO8hXAD/277-after-completing-a-new-worker-add-another-staff-record-fails-page-not-found-regression-issue)

PR #1162, opened 2019-08-01
Bug Fix | Subsidiary Workplace not Updating Correctly

ChazUK:


ChazUK:
[Subsidiary Workplace summary data not refreshed after user changes an attribute](https://trello.com/c/4k6hTit0/161-subsidiary-workplace-summary-data-not-refreshed-after-user-changes-an-attribute)

PR #1163, opened 2019-08-01
Modified "localIdentifiers" endpoint to return summary information on…

wozzer72:
… which local ids are missing from which establishment and/or worker

https://trello.com/c/h2dX27BB


wozzer72:

PR #1167, opened 2019-08-02
Fixing return of starters leavers and vacancies confirm pages

ChazUK:


ChazUK:
[(2) Can't access all pages from the workplace summary](https://trello.com/c/XQcCb6Tt/74-2-cant-access-all-pages-from-the-workplace-summary)

PR #1171, opened 2019-08-04
Bug Fix | Starters, Leavers, Vacancies Confirm CTA

ChazUK:
Now checks if return is set to switch CTA


ChazUK:
[(2) Can't access all pages from the workplace summary](https://trello.com/c/XQcCb6Tt/74-2-cant-access-all-pages-from-the-workplace-summary)

PR #1174, opened 2019-08-05
WDF initial check

dbjsdev:
https://trello.com/c/W7ZHJv93/62-5-wdf-update-worker-record

Have raised this so you can see how on track/far off track I have gone with how this was intended.

Currently everything is set to wdfTemporal = false whilst I try get the output in the right format

I can't recall some of the conversation and am now unsure whether my changes were only supposed to effect the wdf section of the returndata, as opposed to how ive done it here where the wdf section is unchanged and the rest now returns the new wdf info aswell

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -852,13 +852,11 @@ class Establishment extends EntityValidator {
                         //  This decision is done based on if the Establishment is being marked as Completed.
                         // There does not yet exist a Completed property for establishment.
                         // For now, we'll recalculate on every update!
-                        const completedProperty = this._properties.get('Completed');
-                        if (this._properties.get('Completed') && this._properties.get('Completed').modified) {
+                        */
+                        
+                        if(!bulkUploadCompleted){
                             await WdfCalculator.calculate(savedBy.toLowerCase(), this._id, this._uid, thisTransaction);
-                        } else {
-                            // TODO - include Completed logic.
-                            await WdfCalculator.calculate(savedBy.toLowerCase(), this._id, this._uid, thisTransaction);
-                        } */
+                        }
 
                         // if requested, propagate the saving of this establishment down to each of the associated entities
                         if (associatedEntities) {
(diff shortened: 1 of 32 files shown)


server/models/classes/establishment.js
```diff
@@ -1619,6 +1622,10 @@ class Establishment extends EntityValidator {
         myWdf['starters'] = this._isPropertyWdfBasicEligible(effectiveFromEpoch, this._properties.get('Starters')) ? 'Yes' : 'No';
         myWdf['leavers'] = this._isPropertyWdfBasicEligible(effectiveFromEpoch, this._properties.get('Leavers')) ? 'Yes' : 'No';
 
+        let totalWorkerCount = await this.getTotalWorkers();
```
wozzer72: This total worker count should be predicated by `wdf`

server/models/classes/establishment/properties/employerTypeProperty.js
```diff
@@ -8,6 +8,8 @@ const OTHER_MAX_LENGTH=120;
 exports.EmployerTypeProperty = class EmployerTypeProperty extends ChangePropertyPrototype {
     constructor() {
         super('EmployerType');
+
```
wozzer72: Put this in the base class; you only them need to change it to true on those properties that are temporal.

server/models/classes/establishment/properties/employerTypeProperty.js
```diff
@@ -49,7 +51,17 @@ exports.EmployerTypeProperty = class EmployerTypeProperty extends ChangeProperty
             (!currentValue.other && !newValue.other));
     }
 
-    toJSON(withHistory=false, showPropertyHistoryOnly=true) {
+    toJSON(withHistory=false, showPropertyHistoryOnly=true, wdfEffectiveDate = false) {
+
+        if (wdfEffectiveDate) {
```
wozzer72: This function can be easily encapsulated.

PR #1176, opened 2019-08-05
Tech Debt | Authentication Refactor

ChazUK:


ChazUK:
[(2) User not logged out after using browser back button to navigate back to the sign in page](https://trello.com/c/RhJ4mYoU/52-2-user-not-logged-out-after-using-browser-back-button-to-navigate-back-to-the-sign-in-page)

PR #1178, opened 2019-08-05
Bugfix/add workplace complete screen

omarCreativeDev:
- update copy on start and create user screens
- created enum for addWorkplace flow and use it on complete screen
- show different markup on complete screen depending on flow


omarCreativeDev:
[FE | Screens not as per design - Add a CQC workplace](https://trello.com/c/JZa6RDCN/284-fe-screens-not-as-per-design-add-a-cqc-workplace)

PR #1180, opened 2019-08-05
Feature | Logged Out Guard

ChazUK:
Also updated routes to remove the `/public` URL prefix, and fixed where there was no context for links opening in new windows


ChazUK:
[(2) User not logged out after using browser back button to navigate back to the sign in page](https://trello.com/c/RhJ4mYoU/52-2-user-not-logged-out-after-using-browser-back-button-to-navigate-back-to-the-sign-in-page)

PR #1186, opened 2019-08-05
Bug Fix | Workplace Updates

ChazUK:


ChazUK:
[Workplace summary data not refreshed after user changes an attribute](https://trello.com/c/4k6hTit0/161-workplace-summary-data-not-refreshed-after-user-changes-an-attribute)

PR #1197, opened 2019-08-06
Added location not found flow to registration / add workplace

tgreyuk:


tgreyuk:
[(3) Registration - CQC Registered Account that has changed their Address](https://trello.com/c/Uc1q6md3/30-3-registration-cqc-registered-account-that-has-changed-their-address)

src/app/features/workplace-find-and-select/workplace-not-found/workplace-not-found.ts
```diff
@@ -0,0 +1,110 @@
...
+      {
+        name: 404,
+        message: 'No location found.',
+      },
+      {
+        name: 400,
+        message: 'Invalid Postcode.',
+      },
+      {
+        name: 503,
```
omarCreativeDev: we probs dont need the 503's as all error codes above 500 are intercepted and the user is redirected to a problem with service page

PR #1199, opened 2019-08-06
Set submnitted flag to false when record is closed to reset form for …

tgreyuk:
…next record


tgreyuk:
[(1) Staff basic details; No error summary shown](https://trello.com/c/uxB0JMhi/16-1-staff-basic-details-no-error-summary-shown)

PR #1207, opened 2019-08-07
Feature | Redirect On Login

ChazUK:


ChazUK:
[Regression issue: On log in the user is taken to the last page accessed. Should always go to their home tab.](https://trello.com/c/Ma8Vzwfc/275-regression-issue-on-log-in-the-user-is-taken-to-the-last-page-accessed-should-always-go-to-their-home-tab)

PR #1208, opened 2019-08-07
Only display provide workplace panel for edit users

tgreyuk:


tgreyuk:
[When a read-only user logs in to a new account the orange banners shouldn't display](https://trello.com/c/IQlZLRTt/306-when-a-read-only-user-logs-in-to-a-new-account-the-orange-banners-shouldnt-display)

PR #1209, opened 2019-08-07
Removing the "temporal"  predicate…

wozzer72:
https://trello.com/c/W7ZHJv93

The WDF property's `updatedSinceEffectiveDate` was predicated on a mythical "wdf temporal" properties which is no longer relevant, as ALL WDF properties need to reflect their current value based on updated since effective date.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -596,6 +596,166 @@
         "url": "^0.11.0"
       }
     },
+    "@babel/code-frame": {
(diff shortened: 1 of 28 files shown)


wozzer72:
[(5) WDF - Update Worker Record](https://trello.com/c/W7ZHJv93/62-5-wdf-update-worker-record)

PR #1210, opened 2019-08-07
Encapsulating Redirect in Auth Service

ChazUK:


ChazUK:
[Regression issue: On log in the user is taken to the last page accessed. Should always go to their home tab.](https://trello.com/c/Ma8Vzwfc/275-regression-issue-on-log-in-the-user-is-taken-to-the-last-page-accessed-should-always-go-to-their-home-tab)

src/app/core/services/auth.service.ts
```diff
@@ -12,8 +12,8 @@ import { UserService } from './user.service';
 })
 export class AuthService {
   private jwt = new JwtHelperService();
-  public redirect: string;
   private previousUser: string;
+  private _redirect: string;
```
omarCreativeDev: do we have to prefix with an underscore? isn't that what the private keyword is for? : )

omarCreativeDev:
@ChazUK you have a conflict on the package-lock.json file, also left one comment but otherwise this looks good

PR #1213, opened 2019-08-07
WDF - workers and establishment changes

wozzer72:
https://trello.com/c/W7ZHJv93

* Worker/Establishment save - lastEligibleDate - if "WDF::isEligible" is true and if lastEligibleDate is null or is less than the effective date

* Worker/Establishment "WDF::isEligible" is calculated based on all properties being eligible (yes/not relevant) and if yes, their _updatedSince_ being true. This compares to "WDF::currentEligibility" which considers only all property's eligible value regardless of when last updated.

* Worker "other jobs" removed from WDF.

* Establishment "other services" removed from WDF.


wozzer72:
[(5) WDF - Update Worker Record](https://trello.com/c/W7ZHJv93/62-5-wdf-update-worker-record)

PR #1238, opened 2019-08-09
Correcting more errors as a result of refactoring on login; affects a…

wozzer72:
…ll migrated users!

https://trello.com/c/bckKKP6u


wozzer72:
[Migrated users 503 on login](https://trello.com/c/bckKKP6u/327-migrated-users-503-on-login)

PR #1246, opened 2019-08-12
Removing unneeded aria tag

ChazUK:


ChazUK:
[Accessibility - Missing Aria](https://trello.com/c/6uzq5FEr/103-accessibility-missing-aria)

PR #1248, opened 2019-08-12
Bug Fix | Staff Record now exits to return URL if set

ChazUK:


ChazUK:
[Exit navigation from staff question screens is not working](https://trello.com/c/UoOJGekX/101-exit-navigation-from-staff-question-screens-is-not-working)

PR #1250, opened 2019-08-12
Fix service capacity validation

tgreyuk:
This may look slightly odd but this is the rationale.

* Service capacity validation is broken in some instances (in this case Shared lives)  because the validator method relies on the order of the controls.
* The questions are sorted by question.seq, but when added to form control this makes no difference because a FormGroup is an object map and is always ordered by its key (question.id)
* My preference here was to use a FormArray instead but this caused other issues with the error validation service.
* Fix was to prepend question.seq to form control name so validator will pick up controls in correct order.
* Filtered out this prepend when sending back to server.

### DIFF
--- a/src/app/features/workplace/services-capacity/services-capacity.component.html
+++ b/src/app/features/workplace/services-capacity/services-capacity.component.html
@@ -38,28 +38,35 @@
               *ngFor="let question of capacity.questions; let idx = index"
               class="govuk-form-group"
               [class.govuk-form-group--error]="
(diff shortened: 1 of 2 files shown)


tgreyuk:
[Nothing displayed in worker and staff tabs for account with capacity data for 'Shared lives'](https://trello.com/c/N6RQcRuF/106-nothing-displayed-in-worker-and-staff-tabs-for-account-with-capacity-data-for-shared-lives)

PR #1252, opened 2019-08-12
Forgot to update the fetch of all training records to format the comp…

wozzer72:
…leted and expires date.

https://trello.com/c/grDJCvfe


wozzer72:
[Updates to training dates are not being saved accurately](https://trello.com/c/grDJCvfe/97-updates-to-training-dates-are-not-being-saved-accurately)

PR #1254, opened 2019-08-13
localidentifiers fix

dbjsdev:
https://trello.com/c/po4YrDE7/92-parent-subs-permissions-refactor

Initial PR for the above trello card

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -95,7 +95,7 @@ class Establishment extends EntityValidator {
         this._parentUid = null;
         this._parentId = null;
         this._dataOwner = null;
-        this._parentPermissions = null;
+        this._dataPermissions = null;
 
         // interim reasons for leaving - https://trello.com/c/vNHbfdms
         this._reasonsForLeaving = null;
@@ -252,8 +252,8 @@ class Establishment extends EntityValidator {
         return this._dataOwner;
     }
 
-    get parentPermissions() {
-        return this._parentPermissions;
+    get dataPermissions() {
+        return this._dataPermissions;
     }
 
     get numberOfStaff() {
@@ -301,7 +301,7 @@ class Establishment extends EntityValidator {
         this._parentUid = parentUid;
         this._parentId = parentID;
         this._dataOwner = 'Parent';
-        this._parentPermissions = null;     // if the owner is parent, then parent permissions are irrelevant
+        this._dataPermissions = null;   
     }
 
     // this method add this given worker (entity) as an association to this establishment entity - (bulk import)
@@ -604,7 +604,7 @@ class Establishment extends EntityValidator {
                     parentUid: this._parentUid,
                     parentId: this._parentId,
                     dataOwner: this._dataOwner ? this._dataOwner : 'Workplace',
-                    parentPermissions: this._parentPermissions,
+                    dataPermissions: this._dataPermissions,
                     isRegulated: this._isRegulated,
                     locationId: this._locationId,
                     provId: this._provId,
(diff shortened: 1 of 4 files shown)


server/models/establishment.js
```diff
@@ -90,14 +90,14 @@ module.exports = function(sequelize, DataTypes) {
...
+      field: '"DataOwner"',
       default: 'Workplace',
     },
-    parentPermissions: {
+    dataPermissions: {
       type: DataTypes.ENUM,
       allowNull: true,
-      values: ['Workplace','Worker'],
-      field: '"ParentAccess"',
+      values: ['Workplace','Worker', 'None'],
```
ChazUK: I think we asked for this to be `Workplace`, `Workplace and Staff`, `None` for future clarity, but we can work with these.

server/routes/establishments/worker.js
```diff
@@ -29,7 +29,7 @@ const validateWorker = async (req, res, next) => {
     if (req.establishmentId !== req.establishment.id) {
       // the requestor is both a parent and they are requesting against non-primary establishment (aka a subsidiary)
       if (!req.parentIsOwner  &&
-          (req.parentPermissions === null || req.parentPermissions !== "Workplace and Staff")) {
+          (req.dataPermissions === null || req.dataPermissions !== "Workplace and Staff")) {
```
ChazUK: And here we have `Workplace and Staff` being used

server/models/establishment.js
wozzer72: This is the sequelize schema rather than API model. Unfortunately, the sequelize models are not in their own `model/db` folder. These values are the enums we use in the database.

server/routes/establishments/worker.js
wozzer72: These are the values used within the API model, and are as previously agreed.

server/models/establishment.js
wozzer72: We map from the database enums to API enums. But I've just noticed we might be a little remiss on a couple of the mappings.

server/models/classes/establishment.js
```diff
@@ -954,7 +954,7 @@ class Establishment extends EntityValidator {
                 this._parentId = fetchResults.parentId;
                 this._parentUid = fetchResults.parentUid;
                 this._dataOwner = fetchResults.dataOwner;
-                this._parentPermissions = fetchResults.parentPermissions;
+                this._dataPermissions = fetchResults.dataPermissions;
```
wozzer72: The database enums and the API enums need mapping. e.g. "Worker" goes to "Workplace and staff".

server/models/classes/establishment.js
```diff
@@ -1758,7 +1758,7 @@ class Establishment extends EntityValidator {
                   localIdentifier: primaryEstablishmentRecord.LocalIdentifierValue ? primaryEstablishmentRecord.LocalIdentifierValue : null,
                   mainService: primaryEstablishmentRecord.mainService.name,
                   dataOwner: primaryEstablishmentRecord.dataOwner,
-                  parentPermissions: isParent ? undefined : primaryEstablishmentRecord.parentPermissions,
+                  dataPermissions: isParent ? undefined : primaryEstablishmentRecord.dataPermissions,
```
wozzer72: The database enums and the API enums need mapping. e.g. "Worker" goes to "Workplace and staff".

server/models/classes/establishment.js
```diff
@@ -1774,7 +1774,7 @@ class Establishment extends EntityValidator {
                       localIdentifier: thisSub.LocalIdentifierValue ? thisSub.LocalIdentifierValue : null,
                       mainService: thisSub.mainService.name,
                       dataOwner: thisSub.dataOwner,
-                      parentPermissions: thisSub.parentPermissions,
+                      dataPermissions: thisSub.dataPermissions,
```
wozzer72: The database enums and the API enums need mapping. e.g. "Worker" goes to "Workplace and staff".

server/models/establishment.js
wozzer72: But moreso, these are not matching the actual DB enums. Because we don't use sequelize to create the DB schema, we'd not have had a problem on this. But yes @ChazUK , you're correct and these need updating to match that of the complementary `cqc.establishment_data_access_permission` database type.

PR #1257, opened 2019-08-13
Bug Fix | Training using date format instead of ISO

ChazUK:
Updating Completed and Expired Dates to use Default Date Format instead of ISO String


ChazUK:
[Updates to training dates are not being saved accurately](https://trello.com/c/grDJCvfe/97-updates-to-training-dates-are-not-being-saved-accurately)

PR #1258, opened 2019-08-13
Cater for instances where main service exists in other services list

tgreyuk:
It can be argued that the BE should also remove from other services list when a main service is selected (and will mention that in the ticket). However this PR guards against ui errors in this scenario and does fix the bug.


tgreyuk:
[Where a Main service is chosen that already exists as an other service it should be removed as an other service](https://trello.com/c/g7bS7ddJ/119-where-a-main-service-is-chosen-that-already-exists-as-an-other-service-it-should-be-removed-as-an-other-service)

PR #1260, opened 2019-08-13
Bugfix/admin cannot edit a users details

omarCreativeDev:


omarCreativeDev:
[Change - Admin cannot edit a user's details](https://trello.com/c/wg9hu1Dm/96-change-admin-cannot-edit-a-users-details)

src/app/features/workplace/user-account-edit-details/user-account-edit-details.component.html
```diff
@@ -0,0 +1,47 @@
...
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <fieldset class="govuk-fieldset">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+          <span class="govuk-caption-l">User account</span>
+          <h1 class="govuk-fieldset__heading">
+            Your details
+          </h1>
+        </legend>
+
+        <ng-container *ngFor="let item of formControlsMap">
```
ChazUK: Not a huge fan of looping through these, what if the input boxes are different widths? what if we need to add hint text to one of them and not the others etc.

src/app/features/workplace/user-account-edit-details/user-account-edit-details.component.html
```diff
@@ -0,0 +1,47 @@
...
+</app-error-summary>
+
+<form novalidate (ngSubmit)="onSubmit()" [formGroup]="form" id="server-error">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+      <fieldset class="govuk-fieldset">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+          <span class="govuk-caption-l">User account</span>
+          <h1 class="govuk-fieldset__heading">
+            Your details
```
ChazUK: It's not `Your details` as we're editing a different user account

src/app/features/workplace/user-account-view/user-account-view.component.html
```diff
@@ -19,7 +19,7 @@ <h1 class="govuk-heading-l">{{ user.fullname }}</h1>
 </div>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <app-summary-list [topBorder]="true" [summaryList]="userInfo"></app-summary-list>
+    <app-summary-list [canNavigate]="canEdit" [topBorder]="true" [summaryList]="userInfo"></app-summary-list>
```
ChazUK: So is this for Edit users as well? Or just Admin users? The AC's only mention Admin users

src/app/features/workplace/user-account-view/user-account-view.component.ts
```diff
@@ -134,6 +134,7 @@ export class UserAccountViewComponent implements OnInit, OnDestroy {
       {
         label: 'Full name',
         data: this.user.fullname,
+        route: { url: ['edit'] },
```
ChazUK: probably better to make this route more explicit as you can "edit" multiple parts of information about a user, such as their permissions. `edit-details` perhaps?

src/app/shared/components/summary-list/summary-list.component.html
```diff
@@ -12,11 +12,13 @@
...
       {{ item.data }}
     </dd>
-    <dd *ngIf="item.route" class="govuk-summary-list__actions">
-      <a [routerLink]="item.route.url" (click)="emitSetReturn()">
-        Change
-        <span class="govuk-visually-hidden"> {{ item.label.toLowerCase() }}</span>
-      </a>
-    </dd>
+    <ng-container *ngIf="item.route">
+      <dd *ngIf="canNavigate === undefined ? true: canNavigate" class="govuk-summary-list__actions">
```
ChazUK: This seems a bit weird, normally you have the default as false, that the user can't navigate without certain permissions to get rid of any dodgy circumstances where thing will suddenly become available for all users.

src/app/features/workplace/user-account-edit-details/user-account-edit-details.component.html
omarCreativeDev: thats a very good point, at the moment none of the fields have hint text and they are all the same width. happy to remove the loop if something changes : )

src/app/shared/components/summary-list/summary-list.component.html
ChazUK: And why do you need the container? surely it'd be `item.route && canNavigate`

src/app/features/workplace/user-account-edit-details/user-account-edit-details.component.html
omarCreativeDev: good spot thank you, have fixed

src/app/features/workplace/user-account-view/user-account-view.component.ts
omarCreativeDev: done

src/app/shared/components/summary-list/summary-list.component.html
omarCreativeDev: the `canNavigate` is a new optional Input() and isn't used in all `<summary-list>` use cases so therefore the value may be null as opposed to a default false boolean value.

therefore im checking if `canNavigate` doesn't have a value first. i hope that makes sense

src/app/features/workplace/user-account-view/user-account-view.component.html
omarCreativeDev: thank you for spotting, have addressed this

src/app/shared/components/summary-list/summary-list.component.html
ChazUK: Yeah, so the problem here would be that the default is to display the option, when in reality you don't want that to be the default in case of any missing properties. We can live with it though.

PR #1262, opened 2019-08-13
Bug Fix | Tabs not displaying in MS Edge

ChazUK:
Removing the hidden attribute which effects Angular's rendering


ChazUK:
[[## 7976 ##] Unable to see staff and workplace records](https://trello.com/c/6GceDoXn/128-7976-unable-to-see-staff-and-workplace-records)

PR #1266, opened 2019-08-13
Bug Fix | Training Dates use date format from ISO

ChazUK:
Taken from hotfix on UAT


wozzer72:
Added backend changes `...api/models/classes/training.js`, copied from uat v1.10 branch.

PR #1269, opened 2019-08-14
When a parent creates a sub, it's default access permissions should b…

wozzer72:
…e "None".

https://trello.com/c/TPev9pHG


wozzer72:
[Add Workplace - Doesn't set default Data Permissions for Workplace](https://trello.com/c/TPev9pHG/140-add-workplace-doesnt-set-default-data-permissions-for-workplace)

PR #1272, opened 2019-08-14
Bugfix/change password error anchor

omarCreativeDev:
- ensure form errors scroll properly on nested form controls on activate-user-account and registration create username screen & also on account management change password screen
- also used getFirstErrorMessage for consistency


omarCreativeDev:
[Change password: Error summary does not re-focus on the erroneous field](https://trello.com/c/pf8gMjJZ/125-change-password-error-summary-does-not-re-focus-on-the-erroneous-field)

PR #1273, opened 2019-08-14
20190814 wdf be

wozzer72:
https://trello.com/c/owi0abmt

The staff and establishment eligibility and both true or false all the time the overall eligibility is false - they're always calculated not tracked.

https://trello.com/c/xZK7xOhR
Correction to logic for "one or more" worker.


wozzer72:
[WDF - Staff Eligibility not reset when more Staff Records are added](https://trello.com/c/owi0abmt/135-wdf-staff-eligibility-not-reset-when-more-staff-records-are-added)

PR #1276, opened 2019-08-14
bugfix | registration doesnt clear state

omarCreativeDev:


omarCreativeDev:
[Registration State is not cleared when navigating back to the start screen](https://trello.com/c/BKd9lMHn/127-registration-state-is-not-cleared-when-navigating-back-to-the-start-screen)

PR #1277, opened 2019-08-14
When loadin g establishment JSON data, need to handle those the givin…

wozzer72:
…g of either "main service" or "other services", but not both.

https://trello.com/c/8zTJLDbr


wozzer72:
[WDF - API Error when trying to "Confirm and Submit" on a Workplace Summary](https://trello.com/c/8zTJLDbr/132-wdf-api-error-when-trying-to-confirm-and-submit-on-a-workplace-summary)

PR #1278, opened 2019-08-14
Bug Fix | Fix for total workers error

ChazUK:


ChazUK:
[(1) WDF - 'Soft' total staff banner should not be displayed when 'Total number of staff' is zero and also no workers have been created yet.](https://trello.com/c/0GO6X7pR/139-1-wdf-soft-total-staff-banner-should-not-be-displayed-when-total-number-of-staff-is-zero-and-also-no-workers-have-been-created-y)

PR #1283, opened 2019-08-15
Permissions

dbjsdev:
https://trello.com/c/AWmbz8Vu/4-19-parent-and-subsidiaries-permissions

Early PR for review before insertion of hardcoded permissions data

### DIFF
--- a/server/models/cache/singletons/permissions.js
+++ b/server/models/cache/singletons/permissions.js
@@ -14,6 +14,132 @@ let ALL_PERMISSIONS = [
       dataOwner : 'parent',
       dataPermissions: [],
       role: ['Edit']
+    },
+    {
+      code: 'canAddWorker',
+      description: '',
+      dataOwner : 'parent',
+      dataPermissions: ['Workplace', 'Workplace and staff'],
+      role: ['Read', 'Edit']
+    },
+    {
+      code: 'canBulkUpload',
+      description: '',
+      dataOwner : 'parent',
+      dataPermissions: ['Workplace', 'Workplace and staff'],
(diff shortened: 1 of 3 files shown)


server/models/cache/singletons/permissions.js
```diff
@@ -0,0 +1,62 @@
...
+  constructor() {}
+
+  static async initialize() {
+
+  }
+
+  static myPermission(dataOwner, dataPermissions, role){
+    return ALL_PERMISSIONS
+        .filter(x => 
+            (!x.dataOwner.length || dataOwner.toLowerCase() == x.dataOwner.toLowerCase()) &&
```
wozzer72: These are an "eeither, eiither, or" not an "and".

wozzer72: Also, you're only interested in checking data permissions if dataOwner is not `parent`

server/models/cache/singletons/permissions.js
```diff
@@ -0,0 +1,62 @@
...
+];
+
+class PermissionCache {
+  constructor() {}
+
+  static async initialize() {
+
+  }
+
+  static myPermission(dataOwner, dataPermissions, role){
```
wozzer72: Should this be "myPermissions" (plural - because it's returning many permissions

dbjsdev: Corrected

server/models/cache/singletons/permissions.js
dbjsdev: Made an attempt at this

PR #1287, opened 2019-08-15
Implemented custom breadcrumb model and aligned page h1s

tgreyuk:
**UPDATED Previous comments not valid as discussed with Charlie removing dependancy on NestRoutesService.**

The usage of the breadcrumb service remains pretty much the same (import the service and call the show() method. 

Using the show method without any arguments will produce the same behaviour (breadcrumbs will be generated from the router tree).

If this logic does not fit then a new argument `journey` is introduced. A journey is essentially a custom route config that describes the breadcrumb tree.

`this.breadcrumbService.show(Journey.REPORTS)`

When this argument is provided the breadcrumbs will be generated off this config. 

At a very high level that is it.

The custom service matches the paths in much the same way as the Angular router does and then replaces parameters with the actual values from the segment tree.

I did have a look at titles but have not done anything with those. The nestedRouteService behaves in the same way. 

No doubt there will be some comments / questions

### DIFF
--- a/src/app/core/breadcrumb/breadcrumb.model.ts
+++ b/src/app/core/breadcrumb/breadcrumb.model.ts
@@ -9,7 +9,7 @@ export enum JourneyType {
 
 export interface JourneyRoute {
   title?: string;
-  url?: string;
+  path?: string;
   children?: JourneyRoute[];
   referrer?: JourneyRoute;
   fragment?: string;
--- a/src/app/core/breadcrumb/breadcrumb.model.ts
+++ b/src/app/core/breadcrumb/breadcrumb.model.ts
@@ -1,0 +1,16 @@
+export enum JourneyType {
+  PUBLIC,
+  ACCOUNT,
+  BULK_UPLOAD,
+  MY_WORKPLACE,
+  ALL_WORKPLACES,
+  REPORTS,
+}
+
+export interface JourneyRoute {
+  title?: string;
+  url?: string;
+  children?: JourneyRoute[];
+  referrer?: JourneyRoute;
+  fragment?: string;
+}
--- a/src/app/core/breadcrumb/journey.accounts.ts
+++ b/src/app/core/breadcrumb/journey.accounts.ts
@@ -11,19 +11,19 @@ export const accountJourney: JourneyRoute = {
   children: [
     {
       title: 'Account details',
-      url: Path.ACCOUNT_DETAILS,
+      path: Path.ACCOUNT_DETAILS,
       children: [
         {
           title: 'Your details',
-          url: Path.YOUR_DETAILS,
+          path: Path.YOUR_DETAILS,
         },
         {
           title: 'Your password',
-          url: Path.YOUR_PASSWORD,
+          path: Path.YOUR_PASSWORD,
         },
         {
           title: 'Your security question',
-          url: Path.YOUR_SECURITY_QUESTION,
+          path: Path.YOUR_SECURITY_QUESTION,
         },
       ],
     },
--- a/src/app/core/breadcrumb/journey.accounts.ts
+++ b/src/app/core/breadcrumb/journey.accounts.ts
@@ -1,0 +1,31 @@
+import { JourneyRoute } from './breadcrumb.model';
+
+enum Path {
+  ACCOUNT_DETAILS = '/account-management',
+  YOUR_DETAILS = '/account-management/change-your-details',
+  YOUR_PASSWORD = '/account-management/change-password',
+  YOUR_SECURITY_QUESTION = '/account-management/change-user-security',
+}
+
+export const accountJourney: JourneyRoute = {
+  children: [
+    {
+      title: 'Account details',
+      url: Path.ACCOUNT_DETAILS,
+      children: [
+        {
+          title: 'Your details',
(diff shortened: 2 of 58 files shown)


src/app/features/account-management/change-your-details/change-your-details.component.ts
```diff
@@ -40,10 +40,6 @@ export class ChangeYourDetailsComponent extends AccountDetails {
     this.primaryWorkplace = this.establishmentService.primaryWorkplace;
   }
 
-  protected setBackLink(): void {
```
ChazUK: Can probably remove the service injection now?

src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -26,7 +26,7 @@ <h3 class="govuk-heading-s govuk-!-margin-top-6">Other things you can do</h3>
           <a [routerLink]="['/bulk-upload']" (click)="setReturn()">Bulk upload</a>
         </li>
         <li *ngIf="workplace.isParent">
-          <a [routerLink]="['/workplace', 'view-my-workplaces']">View my workplaces</a>
+          <a [routerLink]="['/workplace', 'view-my-workplaces']">View all workplaces</a>
```
ChazUK: URL needs to be updated

src/app/features/workers/staff-record/staff-record.component.html
```diff
@@ -2,8 +2,8 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-xl">
-        <span class="govuk-caption-xl">Staff record</span>
-        {{ worker.nameOrId }}
+        <span class="govuk-caption-xl"> {{ worker.nameOrId }}</span>
```
ChazUK: Not that it matters but there's a space before `{{`

src/app/features/workplace/view-workplace/view-workplace.component.html
```diff
@@ -1,8 +1,8 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds-from-desktop">
     <h1 class="govuk-heading-l">
-      <span *ngIf="workplace?.parentName" class="govuk-caption-xl">{{ workplace.parentName }}</span>
-      {{ workplace.name }}
+      <span class="govuk-caption-xl">{{ workplace.name }}</span>
```
ChazUK: Are we no longer displaying the Parents Name?

src/app/features/account-management/change-your-details/change-your-details.component.ts
tgreyuk: mmmm maybe - but it will mean having to mess with parent class and not sure of the impact..

src/app/features/dashboard/home-tab/home-tab.component.html
tgreyuk: Sure - i will update all these urls.

src/app/features/workers/staff-record/staff-record.component.html
tgreyuk: good spot!!

src/app/features/workplace/view-workplace/view-workplace.component.html
tgreyuk: No that is confirmed for now. But in the process of investigation we discovered that the 'Parent' text was missing on the dashboard. Also fixed in this PR.

tgreyuk:
[(3) Breadcrumb Component](https://trello.com/c/VtvbKp3X/59-3-breadcrumb-component)

src/app/core/breadcrumb/breadcrumb.model.ts
```diff
@@ -0,0 +1,16 @@
...
+  ACCOUNT,
+  BULK_UPLOAD,
+  MY_WORKPLACE,
+  ALL_WORKPLACES,
+  REPORTS,
+}
+
+export interface JourneyRoute {
+  title?: string;
+  url?: string;
```
omarCreativeDev: maybe the type can be `URLStructure` in case fragment or QueryParams are needed?

src/app/core/breadcrumb/journey.accounts.ts
```diff
@@ -0,0 +1,31 @@
+import { JourneyRoute } from './breadcrumb.model';
+
+enum Path {
```
omarCreativeDev: is it worth exporting this as a constant somewhere so the path's can be reused? probs something that could be useful after this pr. I've been pondering on the idea of having a url service or cache some of the route names rather than hardcoded strings all over the place

tgreyuk: Yeah i did have this originally - but it got big and confusing. This makes it clear that the Paths configured sit with the config. (you can have multiple configs in same file)

src/app/core/breadcrumb/breadcrumb.model.ts
tgreyuk: I do quite like that idea. However it is a slightly different model as URLStructure has url as an array. I actually think maybe 'path' would be a better prop name than url in this instance..

PR #1294, opened 2019-08-16
A separate endpoint for controlling the effective date, so the defaul…

wozzer72:
…t report endpoint does not keep resetting it.

https://trello.com/c/j5iywlEq


wozzer72:
[local/dev/staging - WDF effectiveDate override is not consistent](https://trello.com/c/j5iywlEq/164-local-dev-staging-wdf-effectivedate-override-is-not-consistent)

PR #1297, opened 2019-08-16
feature | make permissions api call

omarCreativeDev:


omarCreativeDev:

src/app/shared/components/user-accounts-summary/user-accounts-summary.component.ts
```diff
@@ -18,20 +18,19 @@ export class UserAccountsSummaryComponent implements OnInit, OnDestroy {
...
-        .pipe(withLatestFrom(this.userService.loggedInUser$))
-        .subscribe(([users, loggedInUser]) => {
+        .subscribe(users => {
           this.users = orderBy(
             users,
             ['status', 'isPrimary', 'role', (user: UserDetails) => user.fullname.toLowerCase()],
             ['desc', 'desc', 'asc', 'asc']
           );
-          this.canAddUser = [Roles.Edit, Roles.Admin].includes(loggedInUser.role) && this.userSlotsAvailable(users);
+          this.canAddUser = this.permissionsService.permissions$.value.canAddUser && this.userSlotsAvailable(users);
```
tgreyuk: I think we could wrap the return value in permissions service as per loggedInUser
```
  public get loggedInUser(): UserDetails {
    return this._loggedInUser$.value;
  }
```

tgreyuk: I am not sure about this myself but by deleting ` [Roles.Edit, Roles.Admin].includes(loggedInUser.role)` means that admin user does not always have edit permissions in this instance?

src/app/core/resolvers/permissions.resolver.ts
```diff
@@ -0,0 +1,21 @@
...
+import { PermissionsService } from '@core/services/permissions/permissions.service';
+import { of } from 'rxjs';
+import { catchError } from 'rxjs/operators';
+
+@Injectable()
+export class PermissionsResolver implements Resolve<any> {
+  constructor(private permissionsService: PermissionsService) {}
+
+  resolve(route: ActivatedRouteSnapshot) {
+    const workplaceUid = localStorage.getItem('establishmentId');
```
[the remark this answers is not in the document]
omarCreativeDev: cheers dude, good spot. i have made the same change on parentGuard, permissions and primary workplace resolvers too all at the same time

src/app/shared/components/user-accounts-summary/user-accounts-summary.component.ts
omarCreativeDev: have wrapped the return value as suggested

omarCreativeDev: as discussed on slack BE are returning permissions based upon user role so what i removed should be fine as have double checked with Darren

PR #1298, opened 2019-08-16
Bulk upload download training dates

wozzer72:
https://trello.com/c/IwtqPbuf


wozzer72:
[Cannot download existing data for bulk upload](https://trello.com/c/IwtqPbuf/145-cannot-download-existing-data-for-bulk-upload)

PR #1300, opened 2019-08-16
Use returnUrl as set in viewWorkplace component when required

tgreyuk:


tgreyuk:
[Exit from Change main service page for subsidiary navigating incorrectly](https://trello.com/c/jSItdoOG/147-exit-from-change-main-service-page-for-subsidiary-navigating-incorrectly)

PR #1301, opened 2019-08-16
Removed feedback close button

tgreyuk:


tgreyuk:
[Feedback no longer opens in a new window so what should the [Close] button do](https://trello.com/c/vxnHTGbO/87-feedback-no-longer-opens-in-a-new-window-so-what-should-the-close-button-do)

PR #1303, opened 2019-08-16
Fix empty workplace being saved on check answers from main service

tgreyuk:


tgreyuk:
[Workplace confirmation page shows no details after Main Service is updated](https://trello.com/c/kL1iFfap/228-workplace-confirmation-page-shows-no-details-after-main-service-is-updated)

PR #1309, opened 2019-08-19
permission guard

omarCreativeDev:
new permission guard in order to restrict access to routes
use new guard on add-workplace flow

### DIFF
--- a/src/app/core/guards/permission/permission.guard.ts
+++ b/src/app/core/guards/permission/permission.guard.ts
@@ -14,7 +14,7 @@ export class PermissionGuard implements CanActivate {
     const requiredPermissions = route.data['permissions'] as Array<string>;
 
     return this.permissionsService.permissions$.pipe(
-      map(userPermissions => this.hasInvalidPermissions(requiredPermissions, userPermissions)),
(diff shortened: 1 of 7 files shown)


src/app/core/guards/permission/permission.guard.ts
```diff
@@ -0,0 +1,34 @@
...
+    return this.permissionsService.permissions$.pipe(
+      map(userPermissions => this.hasInvalidPermissions(requiredPermissions, userPermissions)),
+      catchError(() => {
+        this.router.navigate(['/dashboard']);
+        return of(false);
+      })
+    );
+  }
+
+  private hasInvalidPermissions(requiredPermissions: string[], userPermissions: { [key: string]: boolean }): boolean {
```
tgreyuk: Should permissions not be typed to the permissions model (rather than key/value)? I am just thinking it would be safer to do that but you may have a reason not to?

```
public get permissions(): { [key: string]: boolean }  > public get permissions(): Permissions
```

omarCreativeDev: cheers, have typed permissions now

omarCreativeDev:

PR #1310, opened 2019-08-19
Delete associated workers when deleting an establishment.

wozzer72:
https://trello.com/c/3vYYn4W2


wozzer72:
[BE - When deleting a workplace, associated workers are not archived.](https://trello.com/c/3vYYn4W2/40-be-when-deleting-a-workplace-associated-workers-are-not-archived)

PR #1312, opened 2019-08-20
Revised data content structure for each kinesis stream, making it eas…

wozzer72:
…y to consume.

https://trello.com/c/WzrKGR3M

### DIFF
--- a/server/aws/kinesis.js
+++ b/server/aws/kinesis.js
@@ -60,7 +60,7 @@ const workerPump = async (action, worker)  => {
 
   const params = {
     Data: JSON.stringify(pumpData),
-    PartitionKey: worker.uid,    // establishment's primary key
+    PartitionKey: worker.uid,    // workers's primary key
     StreamName: config.get('aws.kinesis.workers'),
   };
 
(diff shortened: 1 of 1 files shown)


wozzer72:
[Kinesis Streams - establishments, workers and users data presented in wrong structure, making it hard to consume.](https://trello.com/c/WzrKGR3M/233-kinesis-streams-establishments-workers-and-users-data-presented-in-wrong-structure-making-it-hard-to-consume)

server/aws/kinesis.js
```diff
@@ -56,7 +59,7 @@ const workerPump = async (action, worker)  => {
   delete worker.postcode;
 
   const params = {
-    Data: JSON.stringify(worker),
+    Data: JSON.stringify(pumpData),
     PartitionKey: worker.uid,    // establishment's primary key
```
nasirmids: Is it not pumpData.worker.uid?

nasirmids: Minor: comment changes from establishment primary key to worker primary key

wozzer72: It can be. But `worker` is a passed in parameter, and hence `worker.uid` is still valid.

wozzer72: Too true; this has been in there for a while.

wozzer72: I've corrected the comment here for Worker and the one for User below too.

nasirmids: ok

nasirmids: ok thanks

PR #1319, opened 2019-08-20
Feature | permissions

omarCreativeDev:
- created permissions guard
- created `primaryWorkplacePermissionsResolver` && `workplacePermissionsRsolver`
- typed permissions
- created permission string literal
- handle workplace permissions caching
- use `this.permissionsService.can(uid, 'canAddUser')` create user account CTA btn

### DIFF
--- a/src/app/core/guards/permission/permission.guard.ts
+++ b/src/app/core/guards/permission/permission.guard.ts
@@ -14,7 +14,7 @@ export class PermissionGuard implements CanActivate {
     const requiredPermissions = route.data['permissions'] as Array<string>;
 
     return this.permissionsService.permissions$.pipe(
-      map(userPermissions => this.hasInvalidPermissions(requiredPermissions, userPermissions)),
+      map(userPermissions => this.hasValidPermissions(requiredPermissions, userPermissions)),
       catchError(() => {
         this.router.navigate(['/dashboard']);
         return of(false);
@@ -22,13 +22,7 @@ export class PermissionGuard implements CanActivate {
     );
   }
 
-  private hasInvalidPermissions(requiredPermissions: string[], userPermissions: { [key: string]: boolean }): boolean {
-    const permissions = [];
-    Object.entries(userPermissions).forEach(item => {
-      if (item[1]) {
-        permissions.push(item[0]);
-      }
-    });
-    return requiredPermissions.every(item => permissions.includes(item));
+  private hasValidPermissions(requiredPermissions: string[], userPermissions: { [key: string]: boolean }): boolean {
+    return requiredPermissions.every(item => Object.keys(userPermissions).includes(item));
   }
 }
--- a/src/app/core/guards/permission/permission.guard.ts
+++ b/src/app/core/guards/permission/permission.guard.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
 import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot } from '@angular/router';
+import { Permissions } from '@core/model/permissions.model';
 import { PermissionsService } from '@core/services/permissions/permissions.service';
 import { Observable, of } from 'rxjs';
 import { catchError, map } from 'rxjs/operators';
@@ -22,7 +23,7 @@ export class PermissionGuard implements CanActivate {
     );
   }
 
-  private hasValidPermissions(requiredPermissions: string[], userPermissions: { [key: string]: boolean }): boolean {
+  private hasValidPermissions(requiredPermissions: string[], userPermissions: Permissions) {
     return requiredPermissions.every(item => Object.keys(userPermissions).includes(item));
   }
 }
--- a/src/app/core/guards/permission/permission.guard.ts
+++ b/src/app/core/guards/permission/permission.guard.ts
@@ -1,0 +1,34 @@
+import { Injectable } from '@angular/core';
+import { ActivatedRouteSnapshot, CanActivate, Router, RouterStateSnapshot } from '@angular/router';
+import { PermissionsService } from '@core/services/permissions/permissions.service';
+import { Observable, of } from 'rxjs';
+import { catchError, map } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class PermissionGuard implements CanActivate {
+  constructor(private permissionsService: PermissionsService, private router: Router) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
+    const requiredPermissions = route.data['permissions'] as Array<string>;
+
+    return this.permissionsService.permissions$.pipe(
+      map(userPermissions => this.hasInvalidPermissions(requiredPermissions, userPermissions)),
+      catchError(() => {
+        this.router.navigate(['/dashboard']);
+        return of(false);
+      })
+    );
+  }
+
+  private hasInvalidPermissions(requiredPermissions: string[], userPermissions: { [key: string]: boolean }): boolean {
+    const permissions = [];
+    Object.entries(userPermissions).forEach(item => {
+      if (item[1]) {
+        permissions.push(item[0]);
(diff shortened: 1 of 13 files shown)


omarCreativeDev:

src/app/core/model/permissions.model.ts
```diff
@@ -1,25 +1,49 @@
...
-    canViewLastUpdatedTime?: boolean;
-    canViewNotifications?: boolean;
-    canViewWdfReport?: boolean;
-    canViewWorker?: boolean;
-    canViewWorkerStatus?: boolean;
-  };
+  permissions: Permissions;
 }
+
+export interface Permissions {
```
ChazUK: I think you can do this

```
Permissions {
  [key: PermissionType]?: boolean
}
```

src/app/core/services/permissions/permissions.service.ts
```diff
@@ -1,28 +1,36 @@
...
 })
 export class PermissionsService {
-  private _permissions$ = new BehaviorSubject<{ [key: string]: boolean }>(null);
+  private _permissions$ = new BehaviorSubject<PermissionsResponse[]>([]);
 
   constructor(private http: HttpClient) {}
 
-  public get permissions(): { [key: string]: boolean } {
-    return this._permissions$.value;
+  public get permissions$(): Observable<PermissionsResponse[]> {
```
ChazUK: Is this ever going to be useful? I expect we wouldn't want to call this publicly

src/app/core/services/permissions/permissions.service.ts
```diff
@@ -1,28 +1,36 @@
...
 
-  public getPermissions(workplaceUid: string): Observable<PermissionsResponse> {
-    return this.http
-      .get<PermissionsResponse>(`/api/establishment/${workplaceUid}/permissions`)
-      .pipe(tap(response => (this.permissions = response.permissions)));
+  public filterPermissions(workplaceUid: string): PermissionsResponse | null {
+   return this._permissions$.value.filter((item => item.uid === workplaceUid))[0] || null;
+  }
+
+  public setPermissions(permissions: PermissionsResponse) {
```
ChazUK: I think it's better to use a collection so that we can easily select the correct workplace by uid, rather than filtering an array of object.

src/app/core/services/permissions/permissions.service.ts
```diff
@@ -1,28 +1,36 @@
...
+    const subsidiaryPermissions: PermissionsResponse[] = this._permissions$.value;
+    subsidiaryPermissions.push(permissions);
+    this._permissions$.next(subsidiaryPermissions);
+  }
+
+  public can(workplaceUid: string, permission: PermissionType): boolean {
+    const permissionsResponse: PermissionsResponse | null = this.filterPermissions(workplaceUid);
```
ChazUK: Do we need to have the ` | null`?

src/app/core/services/permissions/permissions.service.ts
ChazUK: This is referring to the filterPermissions as well.

`this.permissions[workplaceUid]` is better than a filter which could be looping over 1000's of records.

src/app/core/services/permissions/permissions.service.ts
```diff
@@ -1,28 +1,36 @@
...
+    return permissionsResponse ? permissionsResponse.permissions.hasOwnProperty(permission) : false;
```
ChazUK: We want to check if it `hasOwnProperty`, and that the property has returned `true`.

src/app/core/services/permissions/permissions.service.ts
omarCreativeDev: its used by the permissionsGuard so that the route waits till the permissions are available

src/app/core/services/permissions/permissions.service.ts
omarCreativeDev: yea initially i was also checking if the returned property is true but Darren said the api always returns each permission with a true value if applicable otherwise [if false] the permission is omitted completely

src/app/core/model/permissions.model.ts
omarCreativeDev: this throws a compiler error `An index signature parameter type cannot be a union type. Consider using a mapped object type instead.`

src/app/core/guards/permission/permission.guard.ts
```diff
@@ -0,0 +1,36 @@
...
+export class PermissionGuard implements CanActivate {
+  constructor(
+    private permissionsService: PermissionsService,
+    private establishmentService: EstablishmentService,
+    private router: Router
+  ) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<boolean> {
+    const requiredPermissions = route.data['permissions'] as Array<string>;
+    const workplaceUid = this.establishmentService.establishmentId;
```
tgreyuk: What if we want to guard a sub, like a sub staff records.Do we need PrimaryWorkplacePermissionsGuard and WorkplacePermissions guard?

src/app/core/model/permissions.model.ts
omarCreativeDev: had to do this 

```
export interface Permissions {
  [key: string]: boolean;
}
```

src/app/core/services/permissions/permissions.service.ts
ChazUK: But is anything every going to subscribe to it?

src/app/core/services/permissions/permissions.service.ts
ChazUK: It's better to check for both, just in case it changes in the future

src/app/core/services/permissions/permissions.service.ts
tgreyuk: you can argue that all you need is `const permissionsResponse = this.filterPermissions(workplaceUid);` as the type is going to be inferred anyway.

src/app/core/services/permissions/permissions.service.ts
omarCreativeDev: done

src/app/core/services/permissions/permissions.service.ts
omarCreativeDev: removed type as its going to be inferred

src/app/core/services/permissions/permissions.service.ts
omarCreativeDev: done

src/app/core/model/permissions.model.ts
ChazUK: what's the difference between PermissionsList and Permissions?

ChazUK: Ahh got it, Permissions includes the establishments UID as the key

src/app/core/services/permissions/permissions.service.ts
```diff
@@ -1,28 +1,36 @@
...
 })
 export class PermissionsService {
-  private _permissions$ = new BehaviorSubject<{ [key: string]: boolean }>(null);
+  private _permissions$ = new BehaviorSubject<Permissions>({});
 
   constructor(private http: HttpClient) {}
 
-  public get permissions(): { [key: string]: boolean } {
-    return this._permissions$.value;
+  public get permissions$(): Observable<Permissions> {
```
ChazUK: Still don't understand where this is being used, it's not being used inside this class, and doesn't need to be used outside of this class

src/app/core/guards/permission/permission.guard.ts
omarCreativeDev: as discussed on slack im going to get rid of the resolvers and move the logic in to guards. will take care of this in a seperate pr to follow

src/app/core/services/permissions/permissions.service.ts
omarCreativeDev: you are right charlie, im going to remove this in my next pr as per our slack conversation

PR #1328, opened 2019-08-21
Fix error map postcode messages

tgreyuk:


tgreyuk:
[Error message not shown on Change postcode page](https://trello.com/c/N8Rjpcqr/247-error-message-not-shown-on-change-postcode-page)

PR #1329, opened 2019-08-21
Answer pipes should interperet 0 as an answer

tgreyuk:


tgreyuk:
[When sickness days over the past year is 0, this is shown as - in the staff summary](https://trello.com/c/WCsyRseG/252-when-sickness-days-over-the-past-year-is-0-this-is-shown-as-in-the-staff-summary)

PR #1331, opened 2019-08-21
Feature | update permissions guards & resolvers

omarCreativeDev:
This is how permissions are currently working. `PrimaryWorkplacePermissionsResolver` fires after login and stores the primary establishments permissions on the service.

The `PrimaryWorkplacePermissionGuard` currently used on `add-workplace` route, then checks `canAddEstablishment` permission and grants/rejects access to the route.

In addition to this the `WorkplacePermissionGuard` currently used on `view workplace` route first checks for cachedPermissions and if not fetches them via making an api call, then checks `canViewEstablishment` permission before finally granting/rejecting access to the route.

- get rid of `public get permissions$(): Observable<Permissions> {}` as its not useful
- get rid of `WorplacePermissionsResolver` in favour of `WorkplacePermissionGuard` instead
- use `WorkplacePermissionGuard` on workplace module, on view workplace route by checking `canViewEstablishment` permission

doing a page refresh therefore results in fetching the primary permissions first and if on the view workplace route also fetches the subsidiary workplace permissions too


omarCreativeDev:

PR #1333, opened 2019-08-22
Bug Fix | Including own workplace in filter

ChazUK:


ChazUK:
[(3) Bulk Upload - Local Identifiers - Remove ability to update References for Workplaces that are not owned by Parent](https://trello.com/c/LYzUh233/41-3-bulk-upload-local-identifiers-remove-ability-to-update-references-for-workplaces-that-are-not-owned-by-parent)

PR #1334, opened 2019-08-22
Feature/update permissions guards remove resolvers

omarCreativeDev:
First thing first I have discussed permissions with Warren and he said that permissions are always returned no matter who the user is so the api is consistent for the FE to consume.

Basically there are now 2 guards `HasPermissionsGuard` and `CheckPermissionsGuard` and no resolvers. 

`HasPermissionsGuard` grabs establishmentUid from either route param [if on a workplace route] or as a fallback grabs the primary establishmentUid from the establishment service. Then it checks if that establishment has permissions on the service already cached and if not makes the relevant api call to fetch them.

`CheckPermissionsGuard` is the granular check on a route to determine whether or not access should be granted depending upon the existence of a permission e.g `canViewEstablishment` on workplace details route or `canAddEstablishment` on add a workplace route.

if a page refresh is performed on a workplace route then two api calls are made, 1 of primary and one for non primary establishment permissions retrieval.

if a page refresh is performed on a non workplace route then one api call is made to retreive the primary permissions.


omarCreativeDev:

PR #1339, opened 2019-08-23
Bug Fix | Admin Login Loop

ChazUK:


ChazUK:
[(1) [## 8070 ##] Admin users logging in to ASC-WDS](https://trello.com/c/yCIhGrzi/249-1-8070-admin-users-logging-in-to-asc-wds)

src/app/app-routing.module.ts
```diff
@@ -73,11 +75,11 @@ const routes: Routes = [
   },
   {
     path: '',
-    canActivate: [HasPermissionsGuard],
+    canActivate: [AuthGuard, HasPermissionsGuard],
```
omarCreativeDev: does the AuthGuard need to be here? its already on line 77 at a higher level as canActivateChild...

ChazUK: It's here so that it can cancel the HasPermissionsGuard API call if there is no Auth Token

ChazUK: removed

PR #1352, opened 2019-08-27
Move pattern validator inline

tgreyuk:
* Makes all the validators trigger in the same way


tgreyuk:
[Salary question - Are you paid hourly or annually? to not allow pence](https://trello.com/c/zfnp10T6/243-salary-question-are-you-paid-hourly-or-annually-to-not-allow-pence)

PR #1364, opened 2019-08-28
Bug Fix | Clear Permissions State

ChazUK:


ChazUK:
[User role is cached between logins](https://trello.com/c/zT7X04EM/279-user-role-is-cached-between-logins)

PR #1368, opened 2019-08-28
fix permission

omarCreativeDev:


omarCreativeDev:
[Where parent has Workplace and Staff permissions they should be able to view a worker's details](https://trello.com/c/KBn5eYAM/285-where-parent-has-workplace-and-staff-permissions-they-should-be-able-to-view-a-workers-details)

PR #1371, opened 2019-08-29
Local authority report user - framework to run report and download data

wozzer72:
https://trello.com/c/UsiGWFJU
Have updated the user only local authority endpoint to call upon pgsql function to generate the report data and then download the report data ready for formatting as an Excel spreadsheet.

Releasing this code early, because it will allow someone else to work on the Excel spreadsheet creation/download whilst I continue to work on the postgres SQL to generate the report data.

Thanks


wozzer72:
[LA progress report (user)](https://trello.com/c/UsiGWFJU/13-la-progress-report-user)

PR #1372, opened 2019-08-29
replace canChangePermissionsForSubsidiary permission check with dataOwner and isParent checks

omarCreativeDev:


omarCreativeDev:
[For each workplace where the data owner is the workplace, a "Change data owner" link should be shown](https://trello.com/c/zCOqdVII/284-for-each-workplace-where-the-data-owner-is-the-workplace-a-change-data-owner-link-should-be-shown)

PR #1376, opened 2019-08-29
Bug Fix | Training & Qualifications Permissions

ChazUK:


ChazUK:
[(1) Where parent has Workplace and Staff permissions they should be able to view a worker's details but not add or edit training or quals](https://trello.com/c/Aqa3CCIi/292-1-where-parent-has-workplace-and-staff-permissions-they-should-be-able-to-view-a-workers-details-but-not-add-or-edit-training-or)

PR #1377, opened 2019-08-29
bugfix | guard against workplace tab

omarCreativeDev:


omarCreativeDev:
[Where parent owns the data a sub with no view permissions (only linked) should not be able to see any data](https://trello.com/c/mIP5MsSS/294-where-parent-owns-the-data-a-sub-with-no-view-permissions-only-linked-should-not-be-able-to-see-any-data)

PR #1378, opened 2019-08-29
bugfix | cater for admin user on guards

omarCreativeDev:


omarCreativeDev:

src/app/core/services/permissions/permissions.service.ts
```diff
@@ -32,6 +32,10 @@ export class PermissionsService {
   }
 
   public can(workplaceUid: string, permissionType: PermissionType): boolean {
+    if (!workplaceUid) {
```
tgreyuk: I was thinking more along the lines of 
```
if(this.loggedInUser.role === Roles.Admin) {
return true;
}
```

tgreyuk: and the same functionality in the permission route guards. Although think we are going to run into our guards running before resolvers issue on the guards.

PR #1380, opened 2019-08-30
Updated rules for Delete / Edit user

tgreyuk:


tgreyuk:
[Cannot edit user account when only one is associated with a subsidiary](https://trello.com/c/x2LURXDp/289-cannot-edit-user-account-when-only-one-is-associated-with-a-subsidiary)

PR #1381, opened 2019-08-30
national insurance error

dbjsdev:
https://trello.com/c/6yZLnek8/31-staff-records-errors-and-warnings-national-insurance-number

This is an initial PR to ensure it is as intended.

### DIFF
--- a/server/models/BulkImport/csv/workers.js
+++ b/server/models/BulkImport/csv/workers.js
@@ -179,6 +179,8 @@ class Worker {
   static get QUAL_ACH03_WARNING() { return 5550; }
   static get QUAL_ACH03_NOTES_WARNING() { return 5560; }
 
+  static get NI_WORKER_DUPLICATE_ERROR() { return 5570 }
+
   get headers() {
     return this._headers_v1_without_chgUnique.join(",");
   }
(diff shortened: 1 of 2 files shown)


server/models/BulkImport/csv/workers.js
```diff
@@ -2427,6 +2427,20 @@ class Worker {
     };
   }
 
+  // Exceeds national insurance maximum
+  exceedsNationalInsuranceMaximum(originalLineNumber) {
+    return {
+      origin: 'Workers',
+      lineNumber: this._lineNumber,
+      errCode: Worker.DUPLICATE_ERROR,
```
wozzer72: You need to create a new constant for this new error code.

server/models/BulkImport/csv/workers.js
```diff
@@ -2427,6 +2427,20 @@ class Worker {
...
+      errType: `DUPLICATE_ERROR`,
+      error: `NINUMBER is already associated with another full time worker record`,
+      source: this._currentLine.UNIQUEWORKERID,
+      worker: this._currentLine.UNIQUEWORKERID,
+      name: this._currentLine.LOCALESTID,
```
wozzer72: this should be the `NINUMBER` field, not `LOCALESTID`.

server/models/BulkImport/csv/workers.js
dbjsdev: Ah, I was unsure if it had a different error code, would NI_HOURS_ERROR be acceptable?

server/models/BulkImport/csv/workers.js
dbjsdev: change applied

PR #1382, opened 2019-08-30
Add checks for canViewReports

tgreyuk:


tgreyuk:
[Where parent owns the data a sub with no view permissions (only linked) should not be able to see any data](https://trello.com/c/mIP5MsSS/294-where-parent-owns-the-data-a-sub-with-no-view-permissions-only-linked-should-not-be-able-to-see-any-data)

src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -19,26 +19,21 @@ <h2 class="govuk-heading-m">Home</h2>
   <div class="govuk-grid-column-one-third">
     <h3 class="govuk-heading-s">Notifications</h3>
     <p>You have no new notifications</p>
-    <h3 class="govuk-heading-s govuk-!-margin-top-6">Other things you can do</h3>
-    <ul class="govuk-list">
-      <ng-container *ngIf="workplace">
+    <ng-container *ngIf="canBulkUpload || canViewWorkplaces || canViewReports">
```
omarCreativeDev: shouldn't the `canViewReports` permission suffice for allowing access?

tgreyuk: So i put this in so the text `Other things you can do` only shows if at least one of those conditions exist. Otherwise you get the title without anything below it. ( i mean i could have out it just on the title bit it seemed neater just to wrap the whole thing in it.

omarCreativeDev: ah yes with you now, makes sense

PR #1385, opened 2019-08-30
Bugfix | registration and add-workplace back navigation

omarCreativeDev:


omarCreativeDev:
[Back navigation incorrect from 'Select your main service' page](https://trello.com/c/EEdzFbTX/244-back-navigation-incorrect-from-select-your-main-service-page)

PR #1386, opened 2019-08-30
Bugfix/multiple worker api calls

tgreyuk:


tgreyuk:
[Front end client making multiple duplicate API calls to backend on login](https://trello.com/c/OojwScFe/105-front-end-client-making-multiple-duplicate-api-calls-to-backend-on-login)

PR #1393, opened 2019-09-02
Local authority report user - revisions following review with Jackie

wozzer72:
https://trello.com/c/UsiGWFJU

Removing the wrong dependency on worker report records to establishment records, because worker records must be created before establishment, and the establishment needs to aggregate against the Worker records.

In fact, this simplifies the local authority establishment/worker model, because now they both use the primary establishment id as pass when generating the report.

Also adding ordering to the establishment and worker data, in preparation for writing out to Excel (the latter still yet not supported).

### DIFF
--- a/server/config/config.js
+++ b/server/config/config.js
@@ -366,7 +366,7 @@ const config = convict({
         fromDate: {
           doc: 'A fixed from reporting date; in the format YYYY-MM-DD',
           format: String,
-          default: '2019-06-01'
+          default: '2019-09-09'
         },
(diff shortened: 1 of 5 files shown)


wozzer72:
[LA progress report (user)](https://trello.com/c/UsiGWFJU/13-la-progress-report-user)

wozzer72:
[LA progress report (user)](https://trello.com/c/UsiGWFJU/13-la-progress-report-user)

PR #1399, opened 2019-09-03
Remove parent flag from LA report permission

tgreyuk:


tgreyuk:
[Sub who is a LA should be able to run the LA report](https://trello.com/c/t4BiOzOF/307-sub-who-is-a-la-should-be-able-to-run-the-la-report)

PR #1400, opened 2019-09-03
Preserve isPrimary flag when saving user details

tgreyuk:


tgreyuk:
[When an admin user edits the primary user for an establishment, the primary flag is also removed](https://trello.com/c/y6jtwD7b/296-when-an-admin-user-edits-the-primary-user-for-an-establishment-the-primary-flag-is-also-removed)

PR #1401, opened 2019-09-03
bugfix | conditionally display capacities row on workplace tab

omarCreativeDev:


omarCreativeDev:
[Workplace dashboard: should not list, or navigate to, Service capacity when no answer can be provided - BE](https://trello.com/c/hXmN3I4l/245-workplace-dashboard-should-not-list-or-navigate-to-service-capacity-when-no-answer-can-be-provided-be)

PR #1402, opened 2019-09-03
Returning all "establishment specific" permissions for all admin users

wozzer72:
https://trello.com/c/bT6uTvTX

* Adding `Permission::filterForAdminRole`
* Adding new login scenario (as priority order) `requestData.role === 'Admin'`  to `Permission::myPermission`.


wozzer72:

PR #1404, opened 2019-09-03
Admin override on Worker authentication middleware for "Admin" role

wozzer72:
https://trello.com/c/bT6uTvTX
In this new brave world, a user with "Admin" role can run any Worker endpoint. Override worker  authentication middleware to ignore any restructions if role is Admin.


wozzer72:

PR #1405, opened 2019-09-04
Adding admin override to the parent/sub workplace owner/permissions r…

wozzer72:
https://trello.com/c/bT6uTvTX

Now we have given admin role all permission abilities, can you believe they also want to update establishments and workers, as per their new abilities.

This PR adds an exception to `hasAuthorisedEstablishment` for Admin role, to not evaluate data owner/data access.

Thanks


wozzer72:

PR #1408, opened 2019-09-04
Fixed report route guard

tgreyuk:


tgreyuk:
[BE | Restriction on subsidiary's WDF data seems to have gone too far](https://trello.com/c/0T3CRVn0/299-be-restriction-on-subsidiarys-wdf-data-seems-to-have-gone-too-far)

PR #1409, opened 2019-09-04
Local authority report admin

wozzer72:
https://trello.com/c/6DpN53GL

Removed the returning of stub CSV data, with the actual call upon admin report pgsql function, and iterating through the return data formatting CSV data column by column.

Checked the data returned in Insomonia; column by column.

This PR includes a fix update to return "InvalidAttempts" on the search user records; I've squeezed the change in this PR because it's effectively a one liner, and because whilst sat in Leeds I observed the number of requests from the Admin team asking Sopra to check if a user was locked out. The SFC admin team should be able to see this info themselves.

### DIFF
--- a/server/config/config.js
+++ b/server/config/config.js
@@ -366,7 +366,7 @@ const config = convict({
         fromDate: {
           doc: 'A fixed from reporting date; in the format YYYY-MM-DD',
           format: String,
-          default: '2019-06-01'
+          default: '2019-09-09'
         },
         toDate: {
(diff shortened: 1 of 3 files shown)


wozzer72:
[Admin - LA report #27](https://trello.com/c/6DpN53GL/11-admin-la-report-27)

wozzer72:
[Admin - LA report #27](https://trello.com/c/6DpN53GL/11-admin-la-report-27)

PR #1418, opened 2019-09-05
Fixed broken route due to nested array

tgreyuk:


tgreyuk:
[Registration - change account details is broken](https://trello.com/c/VgrGNJU8/320-registration-change-account-details-is-broken)

PR #1423, opened 2019-09-06
Set call to action labels after init as relies on returnUrl

tgreyuk:


tgreyuk:
[Registration - change account details is broken](https://trello.com/c/VgrGNJU8/320-registration-change-account-details-is-broken)

PR #1428, opened 2019-09-07
Local Authority user report spreadsheet export

pjshumphreys:
http://trello.com/c/UsiGWFJU


server/routes/reports/localAuthorityReport/template/xl/sharedStrings.xml
```diff
@@ -0,0 +1,366 @@
...
+        <t>Local authority name</t>
+    </si>
+    <si>
+        <t>Local Authority</t>
+    </si>
+    <si>
+        <t>Parent name</t>
+    </si>
+    <si>
+        <t></t>
```
wozzer72: What does an empty element mean? There are more than one empty elements in this XML structure.

server/routes/reports/localAuthorityReport/user.js
```diff
@@ -1,207 +1,1400 @@
...
+            );
+        } break;
+
+        case 'H': {
+          redIifMissing(
+              putString,
+              cellToChange,
+              reportData.workers[row].mainJob,
+              columnText,
+
```
wozzer72: Are you saving a file here? If so, then best check with Nasir on the container file storage capability - and to estimate (or better still) calculate the expected largest file size to ensure use of local file system can be supported?

And you'll need to cope with multiple concurrent requests. How many concurrent requests, I don't know.

Are you deleting the file - to maintain storage availability?

server/routes/reports/localAuthorityReport/template/xl/sharedStrings.xml
pjshumphreys: The shared strings as used via an index, and this template was creted from a spreadsheet with sample data. In order for the Ids not to change but not share the sample data I've just blanked them out

server/routes/reports/localAuthorityReport/user.js
pjshumphreys: No files are being saved. They are purely created in memory then streamed to the client then garbage collected

wozzer72: Excellent

PR #1432, opened 2019-09-09
20190909 bu la defects backend

wozzer72:
https://trello.com/c/lhVVjVVb
Although this defect was raised against the LA Admin report, the actual defect was in the bulk upload Staff (Worker) downloaded CSV. Two of the records were effectively "Don't know" on the LA reports, but in Bulk Upload were output as empty; should output 999. Impacts both the "CONTHOURS" and "AVGHOURS" CSV column output data.

**During observations/reproducing the defect with Jackie, it is observed a spelling mistake one of the LA Admin columns. Corrected that as part of this PR.**


wozzer72:
[LA Admin Report - contracted/average hours not pulling the correct number through](https://trello.com/c/lhVVjVVb/333-la-admin-report-contracted-average-hours-not-pulling-the-correct-number-through)

PR #1436, opened 2019-09-10
V xg yg bdm

pjshumphreys:
https://trello.com/c/vXgYgBdm

This PR also includes a bugfix on the postcode lookup for the local authority as an undeclared variable was being used

### DIFF
--- a/server/routes/reports/localAuthorityReport/user.js
+++ b/server/routes/reports/localAuthorityReport/user.js
@@ -15,6 +15,10 @@ const sharedStringsName = 'xl/sharedStrings.xml';
 const schema = 'http://schemas.openxmlformats.org/spreadsheetml/2006/main';
 const isNumberRegex = /^[0-9]+(\.[0-9]+)?$/;
 
+//Is the occupation a social worker or occupational therapist?
+//TODO: find out if there's a better way to identify this
+const isSWOrOTRegex = /(?<!not )(social worker|occupational therapist)/i;
+
 //XML DOM manipulation helper functions
 const { DOMParser, XMLSerializer } = new (require('jsdom').JSDOM)().window;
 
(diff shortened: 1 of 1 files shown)


server/routes/reports/localAuthorityReport/user.js
```diff
@@ -15,6 +15,10 @@ const sharedStringsName = 'xl/sharedStrings.xml';
 const schema = 'http://schemas.openxmlformats.org/spreadsheetml/2006/main';
 const isNumberRegex = /^[0-9]+(\.[0-9]+)?$/;
 
+//Is the occupation a social worker or occupational therapist?
+//TODO: find out if there's a better way to identify this
+const isSWOrOTRegex = /(?<!not )(social worker|occupational therapist)/i;
```
wozzer72: You shouldn't doing this. SCQUAL is "No" from the database, but it will be "Must be yes" if the worker is a Social Worker or Occupational Therapist. There is another defect for that missing mapping.

server/routes/reports/localAuthorityReport/user.js
```diff
@@ -1220,8 +1224,16 @@ const updateStaffRecordsSheet = (
           const value = String(reportData.workers[row].relevantSocialCareQualification);
 
           switch (value.toLowerCase()) {
-            case 'missing':
             case 'no': {
+
+              if(!isSWOrOTRegex.test(reportData.workers[row].mainJob)) {
+                //If the worker is not a social worker of an occupational
+                //therapist then no shouldn't be highlighted in red
```
wozzer72: Again, you shouldn't need to be checking on job title.

server/routes/reports/localAuthorityReport/user.js
pjshumphreys: Very well. I'll close this PR then and just do the mapping instead

PR #1438, opened 2019-09-10
Re-enable LA reports for live release

tgreyuk:


tgreyuk:
[Disable functionality not yet ready for live - again](https://trello.com/c/uIP2Rk94/314-disable-functionality-not-yet-ready-for-live-again)

src/app/features/reports/pages/reports/reports.component.html
```diff
@@ -33,5 +33,26 @@ <h2 *ngIf="isParent" class="govuk-body-l">Workforce Development Fund</h2>
         </div>
       </div>
     </app-details>
+
+    <ng-container *ngIf="canRunLocalAuthorityReport || isAdmin">
```
omarCreativeDev: is the `isAdmin` check needed here? im guessing `canRunLocalAuthorityReport` could be true whilst user isn't an admin? ideally the permission should take care of both in my opinion

tgreyuk: Not sure, but i would rather take it out on dev than here as this is the code that has been tested against (this is going straight into live).

tgreyuk: Actually i think canRunLocalAuthorityReport has some other conditions set on it.

PR #1441, opened 2019-09-11
fix min value validation

omarCreativeDev:


omarCreativeDev:
[Capacities screen: Validation issue](https://trello.com/c/vn8RRYNN/317-capacities-screen-validation-issue)

PR #1443, opened 2019-09-11
Config changes for pre production so we can test Bulk upload

NMDSdevopsServiceAdm:
As per card (https://trello.com/c/QgCEoU60/342-scaling-for-bulk-upload) 

I have made changes for Live branch
   - Made the db pool size environment driven
   - container size 10 and 8 GB memory

I have yet to define the env variable in Gov Paas (not AWS) so everytime we change we restart the app 

Thanks

### DIFF
--- a/manifest.bluegreen.yml
+++ b/manifest.bluegreen.yml
@@ -1,5 +1,5 @@
 applications:
-- name: sfcuatgreen
+- name: sfcuatblue
   memory: 8G
   disk_quota: 1GB
   instances: 10
--- a/manifest.bluegreen.yml
+++ b/manifest.bluegreen.yml
@@ -1,7 +1,7 @@
 applications:
 - name: sfcuatblue
   memory: 8G
-  disk_quota: 1GB
+  disk_quota: 4GB
   instances: 10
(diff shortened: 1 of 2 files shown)


manifest.bluegreen.yml
```diff
@@ -1,5 +1,5 @@
 applications:
-- name: sfcuatgreen
+- name: sfcuatblue
   memory: 8G
   disk_quota: 1GB
```
wozzer72: If I remember from the daily snapshot memory issues, when increasing memory, we had to increase the disk_quota too.

NMDSdevopsServiceAdm: Yes i remembe somehting on it but recently i tried it 8 Gb without increasing disk size but i also didnt change snapshot report either. I can make it 2GB now

NMDSdevopsServiceAdm:
> Need the `preproduction.yaml` with db.pool set to 500 - actually, let's make it 450 given we've got 10 containers.

Question though; if i fix it in proprod then i wont be able to change the connection without deploying if i need to? i thought i make this as gov paas env variable so change easily and just restart everytime

PR #1444, opened 2019-09-11
optimised out unnecessary and slow dom querySelector calls

pjshumphreys:
https://trello.com/c/dtKHg6ZT

Optimised out calls to querySelector where nextSibling can be used instead and cached sst element for shared strings


pjshumphreys:
[LA User report - XLSx times out](https://trello.com/c/dtKHg6ZT/341-la-user-report-xlsx-times-out)

PR #1458, opened 2019-09-16
feature | omit establishments when logged in user has no data permiss…

omarCreativeDev:
…ions

### DIFF
--- a/src/app/core/services/user.service.ts
+++ b/src/app/core/services/user.service.ts
@@ -1,6 +1,6 @@
 import { HttpClient, HttpParams } from '@angular/common/http';
 import { Injectable } from '@angular/core';
-import { GetWorkplacesResponse } from '@core/model/my-workplaces.model';
+import { DataPermissions, GetWorkplacesResponse } from '@core/model/my-workplaces.model';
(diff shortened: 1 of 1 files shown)


omarCreativeDev:
[(4) Parent WDF - 2. Viewing other Workplaces 2c. WDF Workplaces](https://trello.com/c/51gAeqPx/19-4-parent-wdf-2-viewing-other-workplaces-2c-wdf-workplaces)

src/app/core/services/user.service.ts
```diff
@@ -96,7 +96,19 @@ export class UserService {
    */
   public getEstablishments(wdf: boolean = false): Observable<GetWorkplacesResponse> {
     const params = wdf ? new HttpParams().set('wdf', `${wdf}`) : null;
-    return this.http.get<GetWorkplacesResponse>(`/api/user/my/establishments`, { params });
+    return this.http.get<GetWorkplacesResponse>(`/api/user/my/establishments`, { params }).pipe(
+      map(response => {
```
pjshumphreys: I'm not keen on the join between establishments and permissions going on in the frontend, but this will achieve the desired aim with no backend changes needed at least

PR #1460, opened 2019-09-17
New endpoint to list notifications for a user

pjshumphreys:
https://trello.com/c/gq5jNg9d

creating an enpoint to list notifications. As disussed, there will be separate restful api endpoints to create and update ownershipChangeRequest records. These endpoint will also create potentially multiple notification records, and more notifications may be created afterwards if the state of users or establishments change.

Rich also has specified pagination will be needed later, so I've added the backend facilities for it now. 

Here's a link the parent ticket that contains the wireframes for the piece of work as a whole. https://trello.com/c/22fJuVNO

### DIFF
--- a/server/data/notifications.js
+++ b/server/data/notifications.js
@@ -19,7 +19,7 @@ exports.getListByUser = async ({ userUid, limit, offset }) =>
   db.query(getListQuery, {
     replacements: {
       userUid,
-      limit: Number.isInteger(limit) ? limit : null,
+      limit: Number.isInteger(limit) && limit > 0 ? limit : null,
       offset: Number.isInteger(offset) && offset > 0 ? offset : 0
     },
     type: db.QueryTypes.SELECT
--- a/server/data/notifications.js
+++ b/server/data/notifications.js
@@ -1,0 +1,27 @@
+'use strict';
+
+const db = rfr('server/utils/datastore');
+
+const getListQuery =
+`
+  SELECT
+    "notificationUid",
+    type,
+    created,
+    "isViewed"
+  FROM cqc."Notifications"
+  WHERE cqc."Notifications"."recipientUserUid" = :userUid
+  LIMIT :limit
+  OFFSET :offset;
+`;
+
+exports.getListByUser = async ({ userUid, limit, offset }) =>
+  db.query(getListQuery, {
+    replacements: {
+      userUid,
+      limit: Number.isInteger(limit) ? limit : null,
+      offset: Number.isInteger(offset) && offset > 0 ? offset : 0
+    },
+    type: db.QueryTypes.SELECT
+  });
+
--- a/server/routes/accounts/user.js
+++ b/server/routes/accounts/user.js
@@ -14,9 +14,9 @@ const config = require('../../config/config');
 const loginResponse = require('../../utils/login/response');
 const uuid = require('uuid');
 
-
 // all user functionality is encapsulated
 const User = require('../../models/classes/user');
+const notifications = rfr('server/data/notifications');
 
 // default route
 router.route('/').get(async (req, res) => {
@@ -727,7 +727,7 @@ router.route('/my/establishments').get(async (req, res) => {
     const isWDF = req.query.wdf ? true : false;
 
     try {
-        const thisUser = new User.User(primaryEstablishmentId);;
+        const thisUser = new User.User(primaryEstablishmentId);
         await thisUser.restore(null, theLoggedInUser, false);
 
         const myEstablishments = await thisUser.myEstablishments(isParent, isWDF, null);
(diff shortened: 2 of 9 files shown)


wozzer72:
Can you change the name of this PR, so it is more reflective of what it is raised for? `Gq5j ng9d` is not very descriptive.

server/data/notifications.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const db = rfr('server/utils/datastore');
+
+const getListQuery =
+`
+  SELECT
+    "notificationUid",
+    type,
```
wozzer72: Why are you using embedded queries into of the sequelize models? Using embedded SQL makes the backend VERY hard to maintain.

wozzer72: And at the very least, this is better encapsulated within the `User` entity, or better still, within a "Notification" entity. At least that way, it is easy to locate SQL.

server/utils/security/generateJWT.js
```diff
@@ -3,14 +3,15 @@ const jwt = require('jsonwebtoken');
...
-exports.loginJWT = (ttlMinutes, establishmentId, establishmentUid, isParent, username, role) => {
+exports.loginJWT = (ttlMinutes, establishmentId, establishmentUid, isParent, username, role, userUid) => {
   const Token_Secret = Authorization.getTokenSecret();
 
   var claims = {
     EstblishmentId: establishmentId,
     EstablishmentUID: establishmentUid,
     role,
     isParent,
+    userUid,
```
wozzer72: You also need to update the regenerate JWT method too, otherwise, after five minutes when the client refreshes, this `userUid` will be missing from the new JWT.

server/routes/accounts/user.js
```diff
@@ -739,6 +739,35 @@ router.route('/my/establishments').get(async (req, res) => {
     }
 });
 
+// Lists the notifications for the logged in (as given by JWT) user
+router.use('/my/notifications', Authorization.isAuthorised);
+router.route('/my/notifications').get(async (req, res) => {
+  try {
```
wozzer72: What happens if the uuid is not found? In all other endpoints, this would return a 404.

I can see you've updated the JWT to include the `userUid` so there is less chance of that. I can also see you're not restoring the `User` entity which I did previously recommend, which means you're not authorising the user first.

server/routes/accounts/user.js
```diff
@@ -739,6 +739,35 @@ router.route('/my/establishments').get(async (req, res) => {
...
+    if (Number.isInteger(+req.query.offset)) {
+      params.offset = +req.query.offset;
+    }
+
+    console.log('/my/notifications', params);
+
+    //return the list
+    return res.status(200).send(await notifications.getListByUser(params));
+  } catch(e) {
+    return res.status(500).send({
```
wozzer72: All other endpoints return 503 not 500.

server/routes/accounts/user.js
```diff
@@ -739,6 +739,35 @@ router.route('/my/establishments').get(async (req, res) => {
...
+    //add pagination parameters if specified on the query string
+    if (Number.isInteger(+req.query.limit)) {
+      params.limit = +req.query.limit;
+    }
+
+    if (Number.isInteger(+req.query.offset)) {
+      params.offset = +req.query.offset;
+    }
+
+    console.log('/my/notifications', params);
```
wozzer72: Wise to remove debug logging.

server/routes/accounts/user.js
```diff
@@ -739,6 +739,35 @@ router.route('/my/establishments').get(async (req, res) => {
...
+router.use('/my/notifications', Authorization.isAuthorised);
+router.route('/my/notifications').get(async (req, res) => {
+  try {
+    //pull the user's uuid out of JWT
+    const params = {
+      userUid: req.userUid
+    };
+
+    //add pagination parameters if specified on the query string
+    if (Number.isInteger(+req.query.limit)) {
```
wozzer72: There is a whole backlog card about pagination. Pagination should be done consistently across the site, not piecemeal endpoint by endpoint.

server/data/notifications.js
pjshumphreys: How so? I'm not even bothering with a sequalize model as it all it does is duplicate the database schema.

server/utils/datastore/index.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+// This file abstracts away details of how to easily and safely run a query
+// using the existing database connection pool, but avoiding most of the
+// complexity introduced by Sequelize.
+
+// Used to easily validate the parameters to the query
+const ajv = (new (require('ajv'))());
```
wozzer72: What is this for?????

server/utils/security/generateJWT.js
pjshumphreys: I have done that one also as part of this PR

server/routes/accounts/user.js
pjshumphreys: Then you'll get an empty record set. In my view, this is what should happen otherwise users will go probing for others notifications

server/routes/accounts/user.js
pjshumphreys: Why? that's not my understanding of what the code means. 503 is if an external service to the one being contacted is unavailable. i.e. the postgres server was doen for instance.

server/routes/accounts/user.js
pjshumphreys: I can do if you prefer, but this isn't particularly verbose and is still useful to see in production logs

server/routes/accounts/user.js
pjshumphreys: Has anything been done regarding pagination so far? This has been how pagination has been hangled on all the other projects I've seen or worked on up to now. This could be the beginning of the pagination approach used across this project.

server/utils/datastore/index.js
pjshumphreys: Json schema validation. I'm using this to ensure that sql parameters are passed in a way that mitigates sql injection

server/data/notifications.js
wozzer72: That's your choice I supposed. If I was staying, then I would not accept it. By embedding SQL you must accept the maintenance liability that comes with it.

server/utils/security/generateJWT.js
wozzer72: OK. It's hard to see from just the changes. As long as you're aware, I'm good.

server/routes/accounts/user.js
wozzer72: You know better of course!

server/routes/accounts/user.js
wozzer72: Whatever

PR #1466, opened 2019-09-17
bugfix | correct workplaces filtering

omarCreativeDev:
only apply filtering on wdf workplaces list page


omarCreativeDev:
[(4) Parent WDF - 2. Viewing other Workplaces 2c. WDF Workplaces](https://trello.com/c/51gAeqPx/19-4-parent-wdf-2-viewing-other-workplaces-2c-wdf-workplaces)

PR #1469, opened 2019-09-18
HOTFIX - Bulk Upload F134107 failing on Download Workplaces

wozzer72:
https://trello.com/c/m8XEJyhS
Handling better when entity.[starters|leavers|vacancies] being null when downloading the Establishment CSV content.

### DIFF
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -2312,11 +2312,11 @@ class Establishment {
     });
 
     columns.push(uniqueJobs.map(thisJob => BUDI.jobRoles(BUDI.FROM_ASC, thisJob)).join(';'));
-    if (entity.starters !== null && !Array.isArray(entity.starters)) {
+    if (entity.starters && !Array.isArray(entity.starters)) {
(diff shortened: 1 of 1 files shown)


wozzer72:
[[## 10062 ##] Downloaded workplace file does not include all workplaces](https://trello.com/c/m8XEJyhS/364-10062-downloaded-workplace-file-does-not-include-all-workplaces)

server/models/BulkImport/csv/establishments.js
```diff
@@ -2322,7 +2322,7 @@ class Establishment {
       } else {
         columns.push('');
       }
-    } else {
+    } else if (entity.starters !== null) {
```
pjshumphreys: In my view this section of code is more complex than it needs to be and would prefer this refactored. I don't understand why retesting would take to long to do for a simple mapping of an array, but ass this is for a hotfix I'll approve anyhow.

wozzer72: Changed the logic because the original was flawed

wozzer72:
@pjshumphreys  - revised logic that does two things:

1. Handles the original null property error as observed by the live P1
2. Handles another test case where the output was blank and should have been 0.

Refer to the trello card for the test cases.

PR #1471, opened 2019-09-19
HOTFIX Bulk Upload Head Office and Zero Staff

wozzer72:
Allowing "0" as valid numberOfStaff value when restoring StaffProperty from JSON. BU Establishment  validation to only allow 0 if main service head office.


wozzer72:
[Number of Staff=0 when Main Service is head office](https://trello.com/c/k6l7XwaY/365-number-of-staff0-when-main-service-is-head-office)

wozzer72:
Pete Woodford has confirmed these go into the hotfix

PR #1472, opened 2019-09-19
bugfix | correct workplaces filtering

omarCreativeDev:


omarCreativeDev:
[Parent WDF Report - Need to display all workplaces the parent has access to](https://trello.com/c/1rCqMVy1/367-parent-wdf-report-need-to-display-all-workplaces-the-parent-has-access-to)

PR #1473, opened 2019-09-19
New endpoint to request data ownership with permissions

rchitranshi:
https://trello.com/c/BmEiIfud


pjshumphreys:
[New endpoint to request data ownership with permissions](https://trello.com/c/BmEiIfud/344-new-endpoint-to-request-data-ownership-with-permissions)

pjshumphreys:
[New endpoint to request data ownership with permissions](https://trello.com/c/BmEiIfud/344-new-endpoint-to-request-data-ownership-with-permissions)

server/routes/accounts/user.js
```diff
@@ -878,4 +879,32 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
   return res.set({'Authorization': 'Bearer ' + token}).status(200).json(response);
 });
 
+// PUT request for ownership change request
+router.use('/my/:id/ownershipChange', Authorization.isAuthorised);
```
wozzer72: The middleware function `isAuthorised` only confirms the JWT is valid. It does nothing to assert that the resource being requested is valid for the given user's scope - namely, if trying to change the ownership on an establishment, then that user MUST have access to that establishment. That establishment must exist.

For that reason, you should be using `hasAuthorisedEstablishment`.

wozzer72: Regarding the namespace on this API. What is `:id` parameter here? The `/my` namespace is API is associated with `/user` - so it's the user's API namespace.

Consequently, I would expect `:id` to be a User UID, but that doesn't make sense if this request is to change ownership of an establishment, in which case, the API namespace should be `.../establishment/:id`.

wozzer72: And the API has no restrictions that bind it to the frontend. You cannot assume that there is a frontend app that is sequencing all API calls, and thus protecting the resources controlled by the API.

That is highlighted in that you can use an RESTful client, such as, postman or Insomina to invoke the APIs directly, or even within the console using a browser's dev tools (which can then be easily abused once you've logged in).

There should be no way that a user who belongs to one establishment is allowed to trigger a change of ownership on another establishment they have no association with.

server/routes/accounts/user.js
```diff
@@ -878,4 +879,32 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
...
+        return res.status(404).send({
+            message: 'Invalid request',
+          });
+    }else{
+        let resp = await ownership.lastOwnershipRequest(params);
+        return res.status(200).send(resp[0]);
+    }
+  } catch(e) {
+    return res.status(500).send({
+      message: e.message
```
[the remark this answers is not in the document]
rchitranshi: const thisError = new Establishment.EstablishmentExceptions.EstablishmentRestoreException(
      thisEstablishment.id,
      thisEstablishment.uid,
      null,
      err,
      null,
      `Failed to retrieve Establishment with id/uid: ${establishmentId}`);

    console.error('establishment::permissions GET/:eID - failed', thisError.message);
    return res.status(503).send(thisError.safe);

Can I use this for returning error

server/routes/accounts/user.js
```diff
@@ -878,4 +879,32 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
...
+  try {
+    const params = {
+        ownerRequestChangeUid: uuid.v4(),
+        userUid: req.userUid, //pull the user's uuid out of JWT
+        subEstablishmentId: req.params.id, // and the id from the url
+        permissionRequest: req.body.permissionRequest
+    };
+
+    //save records
+    let changeRequestResp = await ownership.changeOwnershipRequest(params);
```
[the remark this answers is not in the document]
rchitranshi: > What happens if there are multiple requests against the same establishment to change the ownership? Does the last request close down previous requests?

I need to make changes according to that. But I don't know if multiple requests can be sent or it will be a single request. Please confirm.

server/routes/accounts/user.js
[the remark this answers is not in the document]
rchitranshi: console.error("/establishment/:id/ownershipChange: ERR: ", err.message);
    return res.status(503).send({});

Now I have used this.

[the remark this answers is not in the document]
wozzer72: Much better. It means we are restricted to determining what errors are happening through having access to the console, but it is better than exposing unexpected content.

server/routes/accounts/user.js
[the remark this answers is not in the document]
wozzer72: Of course multiple requests can be sent. By a malicious hacker. By an error in logic on the frontend. By bad tests scripts. Or it could even be business logic, like:

On Monday, a primary user thinks they need onwership on their workspace from their parent. They initiate the request.

Then on Wednesday, they haven't heard back, so they make another request.

And on Friday, they've still not heard back, so they make another request.

server/routes/accounts/user.js
```diff
@@ -878,4 +879,44 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
...
+        userUid: req.userUid, //pull the user's uuid out of JWT
+        subEstablishmentId: req.params.id, // and the id from the url
+        permissionRequest: req.body.permissionRequest
+    };
+
+    //check post establishment ID has IsParent value true and has some ParentID
+    let checkEstablishmentResult = await ownership.checkEstablishment(params);
+    if(!checkEstablishmentResult.length){
+        return res.status(400).send({
+            message: 'Establishement is not a subsidiary',
```
wozzer72: a lil typo on `Establishment`

server/data/ownership.js
```diff
@@ -0,0 +1,78 @@
...
+
+const insertNotificationQuery =
+`
+INSERT INTO
+cqc."Notifications"
+("notificationUid", "type", "typeUid", "recipientUserUid", "isViewed")
+VALUES (:nuid, :type, :typUid, :recipientUserUid, :isViewed);
+`;
+
+exports.checkEstablishment = async(params) =>
```
wozzer72: I'm not overly comfortable seeing an express `request.params` object passed into what is effectively a DAO.
Because of threats of SQL injection, I would like to see explicit arguments here, and some level of argument validation/assertion.

server/routes/accounts/user.js
```diff
@@ -878,4 +879,44 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
...
+  try {
+    const params = {
+        ownerRequestChangeUid: uuid.v4(),
+        userUid: req.userUid, //pull the user's uuid out of JWT
+        subEstablishmentId: req.params.id, // and the id from the url
+        permissionRequest: req.body.permissionRequest
+    };
+
+    //check post establishment ID has IsParent value true and has some ParentID
+    let checkEstablishmentResult = await ownership.checkEstablishment(params);
```
wozzer72: You should validate all params!

If you expect an integer, validate for an integer and return 400 (bad request) if the given input is not as expected.

server/routes/accounts/user.js
```diff
@@ -878,4 +879,44 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
...
+    if(!checkEstablishmentResult.length){
+        return res.status(400).send({
```
wozzer72: Yes, it could be that the given establishment is not a subdiary, but you'll also get here if the establishment is not found, which should be a HTTP 404 response.

server/routes/accounts/user.js
rchitranshi: Only expecting parameter is : permissionRequest. This will be normal text.

rchitranshi: But I'll check this

server/data/ownership.js
pjshumphreys: SQL injection can't happen using this datastore object as it asserts only named parameters are used.

rchitranshi: Will discuss about it. For time being please ignore it and approve this PR, Will make changes accordingly in upcoming tasks if this is not a good approach.

wozzer72: So there is some level of protection against SQL injection, but there is still no parameter validation here.

If you're expecting a parameter being establishment ID and you're expecting that to be a UUID, then extract the parameter, called is `establishmentID` and assert that it is a UUID, else return HTTP 400 (bad request).

server/routes/accounts/user.js
wozzer72: And besides, `permissionRequest` cannot be just normal text. There are three values that can be expected. It is an enum.

server/data/ownership.js
rchitranshi: Ok sure. will add validation of params.

server/routes/accounts/user.js
```diff
@@ -878,4 +879,87 @@ router.route('/swap/establishment/:id').post(async (req, res) => {
   return res.set({'Authorization': 'Bearer ' + token}).status(200).json(response);
 });
 
+// PUT request for ownership change request
+router.use('/establishment/:id/ownershipChange', Authorization.hasAuthorisedEstablishment);
```
wozzer72: This endpoint cannot exist under the `.../routes/account/user` router, because the API endpoint namespace would be `.../api/user/establishment/...`.

This endpoint needs to be moved to it's own file under `.../routes/establishment`.

pjshumphreys: The requirement is that either sub establishments can only request their own data from their parent or a parent establishment can request ownership of a subs data back from the sub. The route used should be appropriate for those situations.

server/data/ownership.js
```diff
@@ -0,0 +1,110 @@
...
+            uid: params.ownerRequestChangeUid,
+            subEstId: params.subEstablishmentId,
+            permReq: params.permissionRequest,
+            appStatus: "REQUESTED",
+            userUid: params.userUid
+          },
+          type: db.QueryTypes.INSERT
+    })
+
+exports.insertNewNotification = async (params) =>
```
pjshumphreys: This function should be in data/notifications.js instead

server/data/ownership.js
```diff
@@ -0,0 +1,110 @@
...
+const insertNotificationQuery =
+`
+INSERT INTO
+cqc."Notifications"
+("notificationUid", "type", "typeUid", "recipientUserUid", "isViewed")
+VALUES (:nuid, :type, :typUid, :recipientUserUid, :isViewed);
+`;
+
+exports.getEstablishmentDetails = async(params) =>
+  db.query(getEstablishmentDetailsQuery, {
```
pjshumphreys: The codebase already has a model object that encapsulates the business logic for establishments. You should not duplicate that existing capability here.

server/routes/accounts/user.js
rchitranshi: Handled this validation.

server/routes/accounts/user.js
rchitranshi: Added check for permissionRequest for fixed ENUM values

server/data/ownership.js
rchitranshi: Added validaiton for required params

server/routes/accounts/user.js
rchitranshi: Changed

server/routes/accounts/user.js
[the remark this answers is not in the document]
rchitranshi: Handled for multiple request

server/routes/establishments/index.js
```diff
@@ -65,6 +66,7 @@ router.use('/:id/bulkupload', BulkUpload);
 router.use('/:id/localIdentifier', LocalIdentifier);
 router.use('/:id/localIdentifiers', LocalIdentifiers);
 router.use('/:id/permissions', Permissions);
+router.use('/:id/actions', Actions);
```
pjshumphreys: Can you change the route to /:id/ownershipChange rather than /:id/actions/ownershipChange please?

rchitranshi: @wozzer72 asked me to make these changes.

rchitranshi:
[New endpoint to request data ownership with permissions](https://trello.com/c/BmEiIfud/344-new-endpoint-to-request-data-ownership-with-permissions)

server/data/ownership.js
rchitranshi: Used establishment model for this purpose.

server/routes/establishments/index.js
rchitranshi: Removed "actions" from request url.

PR #1484, opened 2019-09-20
bugfix | wdf navigation issue

omarCreativeDev:


omarCreativeDev:
[Parent WDF Report - Navigation when viewing subs workers](https://trello.com/c/iho2v4Uf/372-parent-wdf-report-navigation-when-viewing-subs-workers)

PR #1488, opened 2019-09-24
HOTFIX Preventing deleting primary establishment on BU upload

wozzer72:
https://trello.com/c/S1ppD7YK
Adding a trap to prevent setting a status of delete on the primary establishment CSV record

Note - when testing the new validation message, I noticed that the validation message had `undefined` for the establishment name. It was missing the validation message `name` attribute.

As I copied the existing CSV `Establishment::missingPrimaryEstablishmentError` method, then I updated that method too to add the `name` attribute.


wozzer72:
[[## 10341 ##] Bulk Upload is allowing the parent account to be deleted](https://trello.com/c/S1ppD7YK/378-10341-bulk-upload-is-allowing-the-parent-account-to-be-deleted)

PR #1499, opened 2019-09-26
New endpoint to approve/deny an ownership change request with reason text

rchitranshi:
https://trello.com/c/R4dTkP54

### DIFF
--- a/server/routes/establishments/ownershipChange.js
+++ b/server/routes/establishments/ownershipChange.js
@@ -54,6 +54,12 @@ router.route('/').post(async (req, res) => {
             if(getRecipientUserDetails.length){
               //save records
               params.recipientUserUid = getRecipientUserDetails[0].UserUID;
+              let saveDataOwnershipRequested = await ownership.changedDataOwnershipRequested(params);
+              if(!saveDataOwnershipRequested){
+                return res.status(400).send({
+                    message: 'Invalid request',
+                });
+              }
               let changeRequestResp = await ownership.changeOwnershipRequest(params);
               if(!changeRequestResp){
                   return res.status(400).send({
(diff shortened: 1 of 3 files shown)


server/routes/establishments/ownershipChange.js
```diff
@@ -81,9 +87,82 @@ router.route('/').post(async (req, res) => {
...
+                      return res.status(201).send(resp[0]);
+                  }
+              }
+            }
+          }
+        }
+      }
+    }catch(e){
+      console.error("/establishment/:id/ownershipChange: ERR: ", e.message);
+      return res.status(503).send({});
```
wozzer72: Why have you got nested try/catch blocks?

server/routes/establishments/ownershipChange.js
```diff
@@ -81,9 +87,82 @@ router.route('/').post(async (req, res) => {
...
+// PUT request for ownership change request approve/reject
+router.route('/').put(async (req, res) => {
+  try{
+    const uuidRegex = /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/;
+
+    // Get establishment details
+    const thisEstablishment = new Establishment.Establishment(req.username);
+    try{
+      if(await thisEstablishment.restore(req.establishmentId, false)){
+        if(!thisEstablishment){
```
wozzer72: This will always be true, because you're testing on the object (which is a new from above), rather than the result of the restore!

server/routes/establishments/ownershipChange.js
```diff
@@ -81,9 +87,82 @@ router.route('/').post(async (req, res) => {
...
+          return res.status(404).send({
+            message: 'Establishment is not found',
+          });
+        }else if(thisEstablishment.isParent || thisEstablishment._parentId === null || thisEstablishment.archived){
```
wozzer72: Ideally, you should expose a getter for `_parentId` - because you're using it outside of the entity.

server/routes/establishments/ownershipChange.js
```diff
@@ -81,9 +87,82 @@ router.route('/').post(async (req, res) => {
...
+          return res.status(404).send({
+            message: 'Establishment is not a subsidiary',
+          });
+        }else{
+          const params = {
+            userUid: req.userUid,
+            subEstablishmentId: req.establishmentId,
+            approvalStatus: req.body.approvalStatus,
```
wozzer72: Where is your input validation????

server/routes/establishments/ownershipChange.js
rchitranshi: Removed inner try-catch block.

server/routes/establishments/ownershipChange.js
rchitranshi: Make changes accordingly.

server/routes/establishments/ownershipChange.js
rchitranshi: Created a getter for parentId.

server/routes/establishments/ownershipChange.js
rchitranshi: Added validation for approvalStatus.

rchitranshi:
Tested API locally. Added 404 if Establishment not found.

server/routes/establishments/ownershipChange.js
wozzer72: Actually, the await will return the true/false on the response of the restore, so it should be good after all. I'll resolve this comment. But you should test just to be sure.

PR #1503, opened 2019-09-26
bugfix | wdf reports breadcrumb issue

omarCreativeDev:


omarCreativeDev:
[After viewing the WDF data for one of it's subsidiaries the user tries to access WDF for their own workplace. But the previous sub is shown instead](https://trello.com/c/SiuQY0mL/393-after-viewing-the-wdf-data-for-one-of-its-subsidiaries-the-user-tries-to-access-wdf-for-their-own-workplace-but-the-previous-sub)

PR #1504, opened 2019-09-27
the wdf overall validation needs to still use the sequealize model as…

pjshumphreys:
… it does an update

https://trello.com/c/9mc43h4S


pjshumphreys:
[When workplace is 'fixed' so that it now meets requirements, and staff already meeting, the overall wDF flag is not updated to 'meeting](https://trello.com/c/9mc43h4S/395-when-workplace-is-fixed-so-that-it-now-meets-requirements-and-staff-already-meeting-the-overall-wdf-flag-is-not-updated-to-meeti)

PR #1507, opened 2019-09-30
Correct the eligibility check for workplace fields

pjshumphreys:
https://trello.com/c/73vAATTU

The overall wdf eligibility check was testing whether the establishment fields have ever been valid in the current financial year, not whether they were together with all valid staff.

I'm using a previously unused establishment column "EstablishmentWdfEligibility" to store a copy of last wdf eligibility date which is not audited so can be blanked out if the est fields aren't valid.

As part of this ticket I've also run the establishment business object through a linter and static analysis tool called "semistandard" to clean up the code a bit and to find the usage of undeclared variables

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -812,35 +812,41 @@ class Establishment extends EntityValidator {
             updatedBy: savedBy.toLowerCase()
           };
 
-          // every time the establishment is saved, need to calculate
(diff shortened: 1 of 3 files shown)


pjshumphreys:
[Eligibility check for workplace is not correct](https://trello.com/c/73vAATTU/18-eligibility-check-for-workplace-is-not-correct)

PR #1527, opened 2019-10-08
Trickle feed validation

pjshumphreys:
https://trello.com/c/QgCEoU60

I've amended the bulk upload validation to use a chunked transfer response and periodically output a space character to keep the http socket alive. This won't make the response any quicker, but should prevent it from being closed due to a timeout

### DIFF
--- a/server/routes/establishments/bulkUpload.js
+++ b/server/routes/establishments/bulkUpload.js
@@ -419,20 +419,18 @@ router.route('/uploaded').put(async (req, res) => {
         } else {
           const fileNameElements = myFile.Key.split('/');
           const fileName = fileNameElements[fileNameElements.length - 1];
-          returnData.push(
-            generateReturnData(
-              {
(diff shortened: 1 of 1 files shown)


pjshumphreys:
[Scaling for Bulk Upload](https://trello.com/c/QgCEoU60/31-scaling-for-bulk-upload)

server/routes/establishments/bulkUpload.js
```diff
@@ -445,7 +443,16 @@ router.route('/uploaded').put(async (req, res) => {
 
 router.route('/validate').put(async (req, res) => {
   // manage the request timeout
-  req.setTimeout(config.get('bulkupload.validation.timeout') * 1000);
+  // req.setTimeout(config.get('bulkupload.validation.timeout') * 1000);
```
Negi-Anurag: If its not required in future, then we can remove it.

PR #1535, opened 2019-10-10
Added Registrations endpoint and added a registrations screen

aaron-russell:


src/app/features/search/search.component.ts
```diff
@@ -46,14 +73,31 @@ export class SearchComponent implements OnInit {
...
+      error => this.onError(error)
+    );
+  }
+
+  public getRegistrations(): Observable<Registrations[]> {
+    return this.http.get<Registrations[]>('/api/admin/registrations/');
+  }
+
+  public userApproval(data: object) {
+    console.log('We are about to query the API');
```
Negi-Anurag: Can we remove this console.log

src/app/features/search/search.component.ts
```diff
@@ -46,14 +73,31 @@ export class SearchComponent implements OnInit {
...
     }
+    this.getRegistrations().subscribe(
+      data => {
+        this.registrations = data;
+      },
+      error => this.onError(error)
+    );
+  }
+
+  public getRegistrations(): Observable<Registrations[]> {
```
Negi-Anurag: Service should be in different file.

aaron-russell: I just followed how they'd put the search ones in the same file rather than splitting them out

src/app/features/search/search.component.ts
```diff
@@ -46,14 +73,31 @@ export class SearchComponent implements OnInit {
...
       this.form.type = 'establishments';
       this.form.usernameLabel = 'Postcode';
       this.form.nameLabel = 'NDMS ID';
       this.form.subTitle = 'Establishment Search';
       this.form.title = 'Define your search criteria';
       this.form.buttonText = 'Search Establishments';
+    } else {
+      this.form.type = 'registrations';
     }
+    this.getRegistrations().subscribe(
```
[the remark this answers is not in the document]
aaron-russell: You shouldn't need to when it uses Http as it will only complete once

src/app/features/search/search.component.ts
aaron-russell: Done

src/app/features/search/search.component.ts
aaron-russell: Moved into a service with a model

src/app/features/search/search.component.ts
```diff
@@ -46,14 +49,22 @@ export class SearchComponent implements OnInit {
...
       this.form.type = 'establishments';
       this.form.usernameLabel = 'Postcode';
       this.form.nameLabel = 'NDMS ID';
       this.form.subTitle = 'Establishment Search';
       this.form.title = 'Define your search criteria';
       this.form.buttonText = 'Search Establishments';
+    } else {
+      this.form.type = 'registrations';
     }
+    this.registrationsService.getRegistrations().subscribe(
```
mksrivastava: Can we put this block in a different method and call here then set return in this.registration

aaron-russell: Sure, Done

PR #1537, opened 2019-10-10
Bulk Upload Delete Error

pjshumphreys:
https://trello.com/c/ZXO2w83F

I've added a database connection pool acquire timeout of 60 seconds, changed the way transactions are managed when deleting an establishment and have added a new constructor parameter to the Establishment and Worker business objects so as to prevent the wdf calculation from being run when it shouldn't be in bulk upload to reduce the number of queries that get run

### DIFF
--- a/server/config/accessibility.yaml
+++ b/server/config/accessibility.yaml
@@ -9,7 +9,9 @@ aws:
     enabled: false
 
 db:
-    pool: 15
+    pool:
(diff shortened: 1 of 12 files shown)


pjshumphreys:
[Bulk Upload Delete Error](https://trello.com/c/ZXO2w83F/46-bulk-upload-delete-error)

PR #1549, opened 2019-10-14
Store ahmp correctly from bulk upload

pjshumphreys:
http://trello.com/c/JnTcZOra


pjshumphreys:
@aaron-russell They confused me as well :)

PR #1556, opened 2019-10-15
Added initial integration tests that Warren wrote

aaron-russell:
These are not complete, and most don't work but these can then be modified if any API calls are modified


aaron-russell:
@pjshumphreys totally agree but least we have the folder structure and some initial passing tests

PR #1561, opened 2019-10-16
permit average and contracted hours to be appriopriately output from …

pjshumphreys:
…the bulk upload system when the worker's record shows they are on a zero hours contract


pjshumphreys:
[Bulk Upload - Contracted/Average Hours not being pulled through](https://trello.com/c/mMfj4zKB/63-bulk-upload-contracted-average-hours-not-being-pulled-through)

PR #1562, opened 2019-10-16
Feature/downloadable excel report for parent wdf report fe

mksrivastava:


src/app/core/services/report.service.ts
```diff
@@ -41,4 +41,11 @@ export class ReportService {
       responseType: 'blob' as 'json',
     });
   }
+
+  getPatrentWDFReport(workplaceUid: string): Observable<HttpResponse<Blob>> {
```
pjshumphreys: There's a typo here. Should be  getParentWDFReport right?

src/app/features/reports/pages/workplaces/workplaces.component.ts
```diff
@@ -39,6 +51,24 @@ export class WorkplacesComponent implements OnInit {
     );
   }
 
+  public downloadWdfParentReport(event: Event) {
+    event.preventDefault();
+    this.subscriptions.add(
+      this.reportService
+        .getPatrentWDFReport(this.primaryWorkplace.uid)
```
pjshumphreys: Typo here also  it seems

src/app/core/services/report.service.ts
mksrivastava: yes. Corrected. Thanks

src/app/features/reports/pages/workplaces/workplaces.component.ts
mksrivastava: yes. Corrected. Thanks

src/app/core/services/report.service.ts
pjshumphreys: thanks :)

PR #1593, opened 2019-10-23
Feature/swap ownership reject request

mksrivastava:


src/app/core/model/my-workplaces.model.ts
```diff
@@ -36,3 +36,8 @@ export interface WorkPlaceReference {
   name: string;
   uid: string;
 }
+
+export enum RejectOptions {
+  NO = 'No, continue with no reason',
```
pjshumphreys: Does this refer to the ownership change request being denied? If so, a reason must be provided surely?

src/app/shared/components/reject-request-dialog/reject-request-dialog.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(RejectRequestDialogComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
```
pjshumphreys: I think we should start having more unit tests than just the component could be instantiated. Don't add empty test suites like this.

mksrivastava: this is the default one created automatically through command. So I'll remove it.

src/app/core/model/my-workplaces.model.ts
mksrivastava: No this is the option which displays when reject request dialog open.

PR #1615, opened 2019-10-31
Updated the Lambda functions so we can sync data with CQC in seperate requests

aaron-russell:


CQCLocationAPICall/aws/secrets.js
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+      return myLocalSecrets.DB_ROOT_CRT;
+    }
+  } else {
+    throw new Error('Unknown secrets');
+  }
+};
+
+
+module.exports.initialiseSecrets = initialiseSecrets;
```
pjshumphreys: for this you could do: 

module.exports = {
initialiseSecrets,
dbHost,
dbPass,
jwtSecret,
slackUrl,
govNotify,
dbAppUserKey,
dbAppUserCertificate,
dbAppRootCertificate,
adminUrl
};

CQCLocationAPICall/handler.js
```diff
@@ -44,7 +45,7 @@ module.exports.handler =  async (event, context) => {
           });
           break;
       }
-
+      models.sequelize.close();
```
pjshumphreys: I think the code looks better with this. Presumably this is to close off any connections this model open still right?

CQCLocationAPICall/handler.js
```diff
@@ -76,7 +77,7 @@ module.exports.handler =  async (event, context) => {
           changes.push(response.data.changes[i]);
         }
       } while (nextPage != null);
-
+      console.log('There are ' + changes.length + ' changes');
```
pjshumphreys: Nice. It's a shame we don't have consistent logging across the app so we can do more than this

CQCLocationAPICall/handler.js
```diff
@@ -125,20 +129,20 @@ module.exports.handler =  async (event, context) => {
   //
   async function populateAllLocations(url){
     let response;
-    let nextPage='/locations?page=22&perPage=1000';
-  
+    let nextPage='/locations?page=1&perPage=1000';
```
pjshumphreys: interesting, Why was it using page 22 before?

CQCLocationAPICall/models/index.js
```diff
@@ -1,20 +1,60 @@
-'use strict';
```
pjshumphreys: I'd prefer if this line was kept

CQCLocationAPICall/serverless.yml
```diff
@@ -19,11 +19,11 @@ service: cqc-location-call # NOTE: update this with your service name
 
 provider:
   name: aws
-  runtime: nodejs8.10
+  runtime: nodejs10.x
```
pjshumphreys: cool cool

CQCLocationAPICall/handler.js
aaron-russell: Yeah, obviously we don't want this for the API but we need to close the connections before the process is finished.

CQCLocationAPICall/handler.js
aaron-russell: The comment above describes it, it's a total hack, this part of the function is if you want o completely refresh your data from scrath. So if the process fails you are able to containue from where it failed

CQCLocationAPICall/models/index.js
aaron-russell: Okay, I'll put this back in

CQCLocationAPICall/aws/secrets.js
aaron-russell: This was just lifted steaight off the API to get everything working.

PR #1623, opened 2019-11-05
Fixed-error-when-trying-to-approve-swap-ownership-request

Negi-Anurag:
https://trello.com/c/HPxMyT3B/29-error-when-trying-to-approve-swap-ownership-request


src/app/features/notifications/notification/notification.component.ts
```diff
@@ -58,8 +58,10 @@ export class NotificationComponent implements OnInit, OnDestroy {
         ownerRequestChangeUid: this.notification.typeContent.ownerChangeRequestUID,
         approvalStatus: 'APPROVED',
         approvalReason: '',
+        rejectionReason: null,
```
pjshumphreys: do we really want a separate reason text whether the request was aproved or denied. I would expect the reason text to represent why either decision was made

PR #1636, opened 2019-11-07
Added checks for correct swap ownership

Negi-Anurag:
https://trello.com/c/F1VXqNLm/25-16-parent-and-subsidiaries-swap-ownership-of-data-between-workplaces-workplace-b-receives-request

### DIFF
--- a/server/models/classes/establishment.js
+++ b/server/models/classes/establishment.js
@@ -2264,7 +2264,7 @@ class Establishment extends EntityValidator {
   }
 
   //method to fetch all establishment details by establishment id
-  static async fetchAndUpdateEstablishmentDetails(id, data, flag = false) {
+  static async fetchAndUpdateEstablishmentDetails(id, data, establishmentIsParent = false) {
     if (!id) {
       throw new EstablishmentExceptions.EstablishmentRestoreException(null,
         null,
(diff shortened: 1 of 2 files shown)


server/models/classes/establishment.js
```diff
@@ -2264,7 +2264,7 @@ class Establishment extends EntityValidator {
   }
 
   //method to fetch all establishment details by establishment id
-  static async fetchAndUpdateEstablishmentDetails(id, data) {
+  static async fetchAndUpdateEstablishmentDetails(id, data, flag = '') {
```
pjshumphreys: This should be a boolean rather than a string right?

Negi-Anurag: Its a default parameter, so we have set it to empty string.

server/routes/ownershipRequest/index.js
```diff
@@ -70,7 +70,7 @@ router.route('/:id').put(async (req, res) => {
             dataPermissions: checkOwnerChangeRequest[0].permissionRequest
           };
 
-          let recieverUpdate = await Establishment.Establishment.fetchAndUpdateEstablishmentDetails(req.establishment.id, objToUpdate);
+          let recieverUpdate = await Establishment.Establishment.fetchAndUpdateEstablishmentDetails(req.establishment.id, objToUpdate, 'owershipTransfer');
```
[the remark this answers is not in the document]
Negi-Anurag: already changed to true as now we sending boolean value

PR #1656, opened 2019-11-13
Feature/sort column staff record

mksrivastava:


angular.json
```diff
@@ -29,7 +29,11 @@
                 "output": "./assets/"
               }
             ],
-            "styles": ["./src/styles.scss"]
+            "styles": ["./src/styles.scss"],
+            "scripts": [
+              "./node_modules/jquery/dist/jquery.js",
```
Negi-Anurag: We should not use Jquery with angular.

angular.json
```diff
@@ -85,7 +89,10 @@
             "tsConfig": "src/tsconfig.spec.json",
             "karmaConfig": "karma.conf.js",
             "styles": ["src/styles.scss"],
-            "scripts": [],
+            "scripts": [
+              "./node_modules/jquery/dist/jquery.js",
+              "./node_modules/datatables.net/js/jquery.dataTables.js"
```
Negi-Anurag: Can we use any other package which do not require Jquery ?

mksrivastava:
Raised another PR to achieve this functionality without using any data table plugins.

PR #1664, opened 2019-11-14
fixed issue in bulk upload that was introduced by cross validations

pjshumphreys:
https://trello.com/c/E0TJFgHm


pjshumphreys:
[BU - Workers getting deleted when set to update](https://trello.com/c/E0TJFgHm/52-bu-workers-getting-deleted-when-set-to-update)

PR #1684, opened 2019-11-19
Changes in Parent WDF report as per new feedback given.

rchitranshi:


pjshumphreys:
FYI, I've spoken with Pete to confirm this list of changes is correct

PR #1711, opened 2019-11-26
Fixed some feedback points on parent wdf excel report.

rchitranshi:


server/routes/reports/wdf/parent.js
```diff
@@ -87,42 +87,26 @@ const propsNeededToComplete = ('MainService,EmployerTypeValue,Capacities,Service
...
-      let answerKey = -1;
-      capicityDetails.forEach((capacity, key) => {
-        if(capacity.ServiceID === value.MainServiceFKValue){
-          foundMainService = true;
-          answerKey = key;
-        }
-      });
-      if(foundMainService){
-        value.Capacities = (capicityDetails[answerKey].Answer === null)? 'Missing': capicityDetails[answerKey].Answer;
+  for(let i = 0; i< establishmentData.length; i++) {
```
aaron-russell: Why don't you use .map rather than a for loop?

rchitranshi: I didn't use it because inside the loop I need separate instance objects(like i in for loop), as I am waiting for some functions to return their values first. I tried it with forEach and map. But in both cases, it wasn't waiting for the response and because of that, my values were becoming undefined.

aaron-russell:
> I didn't use it because inside the loop I need separate instance objects(like i in for loop), as I am waiting for some functions to return their values first. I tried it with forEach and map. But in both cases, it wasn't waiting for the response and because of that, my values were becoming undefined.  

That's fine

PR #1736, opened 2019-12-03
Added logic to fetch parent name and postcode.

Negi-Anurag:
Added logic to fetch all parent name and postcode.

https://trello.com/c/8RIBCbtd/8-10-parent-and-subsidiaries-subsidiary-linking-to-parent-request-modal


Negi-Anurag:
> parent

Why parent uid is required ?

mksrivastava:
How you uniquely identified the parent workplace using a name or postal code only

PR #1739, opened 2019-12-04
Changed functionality : In case of parent establishment, removed filter. Now all parent's workplace and worker will be there inside report.

rchitranshi:
https://trello.com/c/VlPBCEkP/31-parent-wdf-not-seeing-all-staff-records


rchitranshi:
[Parent WDF - Not seeing all staff records](https://trello.com/c/VlPBCEkP/31-parent-wdf-not-seeing-all-staff-records)

PR #1751, opened 2019-12-06
Skip terms & condition for admin user and removed unused check

mksrivastava:


src/app/features/migrated-user-terms-conditions/migrated-user-terms-conditions.component.ts
```diff
@@ -83,7 +83,6 @@ export class MigratedUserTermsConditionsComponent implements OnInit {
   }
 
   private continue(): void {
-    this.userService.migratedUserTermsAccepted$.next(true);
```
aaron-russell: Will this cause an issue for someone who is migrated?

mksrivastava: There is no column available in db named with migratedUserTermsAccepted and no property available in the response during the login. This is an old code and was not removed. So good to remove it.

aaron-russell: It's done based on their first login date, if they've never been logged in, it will be false, I just wamted to make sure a Migrated user wont always see the screen, that's all

mksrivastava: Migrated user won't always see the screen after accepting the agreed terms

PR #1767, opened 2019-12-10
save and cancel functionality

mksrivastava:


src/app/shared/components/link-to-parent/link-to-parent-dialog.component.ts
```diff
@@ -153,10 +155,10 @@ export class LinkToParentDialogComponent extends DialogComponent implements OnIn
 
   /**
    * Function is used to close dialog window after successful confirmation
-   * @param {boolean} true to close dialog
+   * @param {any} true to close dialog after response or null to close without action
    * @return {void}
    */
-  public closeDialogWindow(confirm: boolean) {
+  public closeDialogWindow(confirm: any) {
```
Negi-Anurag: any specific reason its type is changed to any

mksrivastava: yes. As mentioned in param comment. True to close the dialog after response or null to close without action

PR #1777, opened 2019-12-12
Hopefully a fix for the login issues

aaron-russell:


src/app/features/login/login.component.ts
```diff
@@ -117,42 +117,40 @@ export class LoginComponent implements OnInit, OnDestroy, AfterViewInit {
     console.log('Testing we have access to localstorage');
```
Negi-Anurag: Remove all console.log

aaron-russell: These are there for debugging purposes as this is a live issue, we are trying to diagnose the problem. These will help understand if it sorts the problem. Once we know it all works, they will be removed.

PR #1783, opened 2019-12-13
Added no as an option for the other jobs property

aaron-russell:


aaron-russell:
[(4) Adding `No` option to other job roles](https://trello.com/c/AgfEtXnu/2-4-adding-no-option-to-other-job-roles)

PR #1784, opened 2019-12-16
Be link to parent approve

Negi-Anurag:


server/data/linkToParent.js
```diff
@@ -155,3 +155,67 @@ exports.getSubEstablishmentName = async params =>
...
+exports.updateNotification = async params =>
+  db.query(updateNotificationQuery, {
+    replacements: {
+      notificationUid: params.notificationUid,
+      isViewed: false,
+    },
+    type: db.QueryTypes.UPDATE,
+  });
+
+  const insertNotificationQuery =
```
rchitranshi: This function is already there inside the notification data file. You can use that one.

server/data/linkToParent.js
```diff
@@ -155,3 +155,67 @@ exports.getSubEstablishmentName = async params =>
...
+            type: params.type,
+            typeUid: params.typeUid,
+            recipientUserUid: params.recipientUserUid,
+            isViewed: false,
+            createdByUserUID: params.userUid
+        },
+        type: db.QueryTypes.INSERT
+    });
+
+    const getSubUserDetailsQuery = `
```
rchitranshi: Also this we have created for swap ownership.

PR #1796, opened 2019-12-17
Polling on wdf report

rchitranshi:
Polling on wdf report


server/config/config.js
```diff
@@ -357,6 +357,37 @@ const config = convict({
       },
     }
   },
+  wdfExcelReport: {
```
aaron-russell: We need to get rid of this, no need for extra bucket

PR #1856, opened 2020-01-09
Location id changes for fe

mksrivastava:


src/app/core/services/establishment.service.ts
```diff
@@ -49,6 +49,19 @@ interface MainServiceRequest {
     other?: string;
   };
 }
+interface CQCDatilsRequest {
```
aaron-russell: Should this be CQC Details?

src/app/core/services/establishment.service.ts
```diff
@@ -49,6 +49,19 @@ interface MainServiceRequest {
...
+  addressLine1: string;
+  addressLine2: string;
```
aaron-russell: Address Line 2 can be optional too

src/app/features/workplace/select-workplace/select-workplace.component.html
```diff
@@ -0,0 +1,48 @@
+<app-error-summary *ngIf="submitted && form.invalid" [formErrorsMap]="formErrorsMap" [form]="form"> </app-error-summary>
```
aaron-russell: Could these not use the existing templates?

src/app/core/services/establishment.service.ts
mksrivastava: ok change it to CQCLocationChangeRequest

src/app/core/services/establishment.service.ts
mksrivastava: made it optional and in other models as well.

src/app/features/workplace/select-workplace/select-workplace.component.html
mksrivastava: yes made the changes for common template

PR #1867, opened 2020-01-14
added check for parent role

Negi-Anurag:


src/app/features/workers/staff-record/staff-record.component.html
```diff
@@ -9,10 +9,10 @@ <h1 class="govuk-heading-l govuk-!-margin-bottom-0">
 
     <div class="govuk-grid-column-one-half govuk-util__align-right">
         <a
-        *ngIf="(canEditWorker && isParent) || isAdmin"
+        *ngIf="(canEditWorker && isParent) || (isAdmin && isParent)"
```
mksrivastava: || (isAdmin && isParent)" is not  required. (canEditWorker && isParent) will work for admin user

Negi-Anurag: let me test first, if it works I will remove the above condition.

PR #1906, opened 2020-01-23
Feature/parent and subsidiaries add a workplace

rchitranshi:


rchitranshi:
> Why old files changes are there

Because it was not merged into develop.

PR #1940, opened 2020-01-30
Show services that are null and what else they need

aaron-russell:


aaron-russell:
[[## 14793 ##] Head Office Services only available under non CQC regulated main services](https://trello.com/c/dnlvVIn6/27-14793-head-office-services-only-available-under-non-cqc-regulated-main-services)

PR #2067, opened 2020-04-07
Fixed the training report library, Fixed the issue around duplicate staff and mandatory training fixes

aaron-russell:


aaron-russell:
[Training Report encoding issue](https://trello.com/c/GyTQ4IYk/2-training-report-encoding-issue)

aaron-russell:
[Mandatory Training on Training Report](https://trello.com/c/22rOAKVx/3-mandatory-training-on-training-report)

server/routes/reports/trainingReport/training.js
```diff
@@ -127,24 +134,13 @@ const updateProps = 'JobName,Category,Title,Expires'.split(',');
...
-          }
-        });
-        if (!foundTrn) {
-          reportData.push({ ID: value.ID, NameOrIdValue: value.NameOrIdValue, MandatoryCount: 0, NonMandatoryCount: 0, Count: 0});
-        }
-      }
-    });
-  }
+const createExpireExpiringData = async (trainingData) => {
+  return trainingData.filter((trainingWorker, index, self) => self.findIndex(t => trainingWorker.ID === t.ID) === index).map(
```
chadtomkiss-madetech, line 138: Lovveee this!

Formatting isn't something that should be debated in a PR, but here we have `.filter` and `.map` inline which is hard to parse.

Are we using Prettier? There's a `.prettierrc.yml` but I'm not sure it's being used.

As an example, this is what it would do:

```
return trainingData
  .filter(
    (trainingWorker, index, self) =>
      self.findIndex((t) => trainingWorker.ID === t.ID) === index
  )
  .map((training) => {
    return {
      ID: training.ID,
      NameOrIdValue: training.NameOrIdValue,
      MandatoryCount: 0,
      NonMandatoryCount: 0,
      Count: 0,
    };
  });
```

It's a lot easier to see what functions are being applied and even what the object structure looks like.

server/routes/reports/trainingReport/training.js
```diff
@@ -132,12 +132,20 @@ const updateProps = 'JobName,Category,Title,Expires'.split(',');
  * Helper Function used to create expired/expiring traing data
  *
  * @param {Object} trainingData
- * @param {Object} All customized expired/expiring training report data
+ * @return {Object} All workers with traiing
```
chadtomkiss-madetech, line 135: Fix typo 
```suggestion
 * @return {Object} All workers with training
```

PR #2068, opened 2020-04-07
Fix: WDF confirmation

chadtomkiss-madetech:
**Issue**

When confirming the Workplace details for WDF funds, the logic was checking against the current eligibility which included the eligibility of each Worker.

This meant that the confirming Workplace information would not be possible if you had Workers who were not eligible.

**Changes**

I have updated `displayConfirmationPanel` to check that the date the Workplace was last eligible is before the current effective from date.


chadtomkiss-madetech:
[[## 18187 ##] No confirm and submit button for WDF on the workplace tab](https://trello.com/c/tKYr4Zdu/12-18187-no-confirm-and-submit-button-for-wdf-on-the-workplace-tab)

PR #2070, opened 2020-04-08
Fix: Confirm WDF dialog

chadtomkiss-madetech:
**Issue**

It seems the `WdfWorkplaceConfirmationDialogComponent` was used on the WDF report page but was imported in the `workplace.module.ts` NgModule and not the `reports.module.ts`

I have checked the usage of this component and it looks like we can remove it from `workplace.module.ts`

The WDF dialog used in the components that are under `workplace.module.ts` is `WdfWorkerConfirmationDialogComponent` which is a different dialog.

**Work done**

- Moved `WdfWorkplaceConfirmationDialogComponent` to `reports.module.ts`


chadtomkiss-madetech:
[[## 18187 ##] No confirm and submit button for WDF on the workplace tab](https://trello.com/c/tKYr4Zdu/12-18187-no-confirm-and-submit-button-for-wdf-on-the-workplace-tab)

PR #2071, opened 2020-04-08
Fix: Incorrect total count for Training & Qualifications

chadtomkiss-madetech:
**Issue**

The total # of records was subtracting the # of missing mandatory training records.

This meant that, if a worker had 1 training record and 1 missing training record, it'd calculate as 0.

**Work done**

- Removed missing training record number being included in the number of total records calculation.


chadtomkiss-madetech:
[Mandatory training - Incorrect Counts](https://trello.com/c/1PX6iwP1/15-mandatory-training-incorrect-counts)

PR #2083, opened 2020-04-14
Fixes for defects on Mandatory training and WDF from Develop to Staging

aaron-russell:


src/app/features/reports/pages/wdf/wdf.component.ts
```diff
@@ -4,7 +4,7 @@ import { JourneyType } from '@core/breadcrumb/breadcrumb.model';
 import { Establishment } from '@core/model/establishment.model';
 import { WDFReport } from '@core/model/reports.model';
 import { URLStructure } from '@core/model/url.model';
-import { Eligibility } from '@core/model/wdf.model';
+import { Eligibility, WDFValue } from '@core/model/wdf.model';
```
chadtomkiss-madetech, line 7: I don't think we need this

PR #2085, opened 2020-04-14
Fix: Count of training and quals records for a worker

chadtomkiss-madetech:
**Issue**

The calculation of training records was incorrect - it was making it either the number of up-to-date records or the sum of expired/expiring/missing.

**Work done**

Updated the calculation to be a simple length of training records


chadtomkiss-madetech:
[Mandatory training - Incorrect Counts](https://trello.com/c/1PX6iwP1/15-mandatory-training-incorrect-counts)

PR #2092, opened 2020-04-21
Feat: Change Workplace ID

chadtomkiss-madetech:
#### Issue
Admins would like the ability to change the Workplace ID when approving registrations

#### Work done

- Moved registrations into their own components
- Updated `/api/admin/approval/` endpoint to accept an nmdsId and also check it's unique
- Added a `firstErrorPipe` to get the first error for a form control (so that you don't display multiple errors under an input)
- Added `testing-library/angular` to easily interact with Angular components in tests
- Added tests

#### Tests
- [ ] Yes
- [x] Yes but could do with more


src/app/app.component.ts
```diff
@@ -9,8 +9,6 @@ import { NestedRoutesService } from '@core/services/nested-routes.service';
 import { Angulartics2GoogleTagManager } from 'angulartics2/gtm';
 import { filter, take, takeWhile } from 'rxjs/operators';
 
-enableProdMode();
```
chadtomkiss-madetech, line 12: Having `enableProdMode()` strips some keys from objects that's needed for running tests

i.e `fixture.debugElement`

I've taken this out but may need to wrap in an `if` to only enable when not running the tests

server/routes/admin/approval/index.js
```diff
@@ -84,16 +85,35 @@ router.route('/').post(async (req, res) => {
       console.log(error);
     }
   }else{
+    const nmdsId = req.body.nmdsId;
```
chadtomkiss-madetech, line 88: This endpoint is missing tests

src/app/features/search/registrations/registrations.component.ts
```diff
@@ -0,0 +1,30 @@
...
+    public registrationsService: RegistrationsService,
+  ) {}
+
+  ngOnInit() {
+    this.getRegistrations();
+  }
+  public getRegistrations() {
+    this.registrationsService.getRegistrations().subscribe(
+      data => {
+        this.registrations = data;
```
chadtomkiss-madetech, line 21: I had trouble testing Observables - I think that the `subscribe` isn't being run when first rendering a component.

src/app/features/search/registrations/registrations.component.ts
```diff
@@ -0,0 +1,30 @@
...
+      },
+      error => this.onError(error)
+    );
+  }
+  public handleRegistration(index: number) {
```
chadtomkiss-madetech, line 26: Fix spacing

server/routes/admin/approval/index.js
```diff
@@ -2,10 +2,11 @@
 const express = require('express');
 const router = express.Router();
 const models = require('../../../models');
+const Sequelize = require('sequelize');
+const { ne } = Sequelize.Op;
```
aaron-russell, line 6: Does this name help identify it in the future?

src/app/app.component.ts
aaron-russell: https://stackoverflow.com/a/39678498 you can enable it in certain environments

aaron-russell:
You may also want to connect a Trello ticket

chadtomkiss-madetech:
[Change WorkplaceID on Registration approval](https://trello.com/c/CdH9q1rN/40-change-workplaceid-on-registration-approval)

server/routes/admin/approval/index.js
chadtomkiss-madetech: Thanks for questioning this!

It's a "not equal" operator and the naming was based on the docs

Do you think it needs a comment/link to the docs for a list of operators?

src/app/app.component.ts
chadtomkiss-madetech: I have added a `environment.test.ts` and set the default `environment.ts` to true.

server/routes/admin/approval/index.js
aaron-russell: Yeah, makes sense, as reading that, I wasn't sure what ne stood for. THing we have used `op` elsewhere but I don't feel this is any better! :/

PR #2094, opened 2020-04-27
Allow Administrators to delete Workplaces

two-first-names:
This adds the capability for administrators to delete Workplaces.

The flow to do so is as follows:
- Click Search Establishments
- Search for Establishment, by postcode, ID, etc
- Click Establishment, this switches context to the Establishment
- Click Delete Workplace and Confirm
- This will _archive_ the Workplace and return the user to the Search Establishment screen with their original login context.

### DIFF
--- a/src/app/core/services/auth.service.ts
+++ b/src/app/core/services/auth.service.ts
@@ -17,6 +17,7 @@ export class AuthService {
   private _isAuthenticated$: BehaviorSubject<boolean> = new BehaviorSubject(null);
   private jwt = new JwtHelperService();
   private previousUser: string;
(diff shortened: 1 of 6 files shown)


aaron-russell:
There's also no unit tests. for this code. WOuld be good to make sure we are testing the front end. However, I know @chadtomkiss-madetech has added some unit testing libraries to the front end.

g-m-taylor:
[Administrators can delete workplaces - Administrators can archive workplaces](https://trello.com/c/TVXWYRIG/32-administrators-can-delete-workplaces-administrators-can-archive-workplaces)

PR #2105, opened 2020-05-03
Feature/approval refactor

claresudbery:
Admin approval endpoint: Refactored, with full coverage from unit tests, integration tests and front end tests.


server/routes/admin/approval/index.js
```diff
@@ -3,140 +3,212 @@ const express = require('express');
...
+
+    // Make sure we have the matching user
+    if ((login && login.id) && (username === login.username)) {
+      const workplace = await _findWorkplace(login.user.establishment.id);
+      const user = await _findUser(login.user.id);
+      
+      var workplaceIsUnique = await _workplaceIsUnique(login.user.establishment.id, req.body.nmdsId);
+      if (!workplaceIsUnique) {
+        return res.status(400).json({
+          nmdsId: `This workplace ID (${req.body.nmdsId}) belongs to another workplace. Enter a different workplace ID.`,
```
aaron-russell, line 39: We wouldn't normally send back this message as part of the nmdsId, instead it would be an error message. The front end should then use the status or the message to undersyand what the problem is.

server/routes/admin/approval/index.js
```diff
@@ -3,140 +3,212 @@ const express = require('express');
...
+  }
+};
+
+const _approveOrRejectNewWorkplace = async (req, res) => {
+  const nmdsId = req.body.nmdsId;
+
+  var workplaceIsUnique = await _workplaceIsUnique(req.body.establishmentId, req.body.nmdsId);
+  if (!workplaceIsUnique) {
+    return res.status(400).json({
+      nmdsId: `This workplace ID (${nmdsId}) belongs to another workplace. Enter a different workplace ID.`,
```
aaron-russell, line 62: Same here too.

server/test/unit/routes/admin/approval/index.spec.js
```diff
@@ -0,0 +1,594 @@
...
+    return foundUser;
+  } else {
+    return null;
+  }
+});
+
+sinon.stub(models.establishment, 'findOne').callsFake(async (args) => {
+  if (args.where.id === testWorkplace.id) {
+    return foundWorkplace;
+  } else if (args.where.id[Sequelize.Op.ne] === boobyTrappedEstablishmentId) {
```
aaron-russell, line 74: Not quite sure what boobyTrappedEstablishmentId is?

claresudbery:
[Change WorkplaceID on Registration approval](https://trello.com/c/CdH9q1rN/40-change-workplaceid-on-registration-approval)

server/routes/admin/approval/index.js
chadtomkiss-madetech: @aaron-russell just for clarification, if the status code is 400 then express adds the json to an error object

the reason `nmdsId` is used is that I wrote some code in the front-end component to take the key of an error message and map it to a front-end field to display the error.

https://github.com/NMDSdevopsServiceAdm/SopraSteria-SFC/blob/327bff4fd174342ff3b2ff3218594375cd6a524a/src/app/features/search/registration/registration.component.ts#L60-L63

https://github.com/NMDSdevopsServiceAdm/SopraSteria-SFC/blob/327bff4fd174342ff3b2ff3218594375cd6a524a/src/app/features/search/registration/registration.component.ts#L69-L79

server/routes/admin/approval/index.js
```diff
@@ -3,140 +3,212 @@ const express = require('express');
...
+const _findUser = async (loginId) => {
+  return await models.user.findOne({
+    where: {
+      id: loginId
+    },
+    attributes: ['id']
+  });
+};
+
+const _approveNewUser = async (login, workplace, nmdsId, res) => {
```
chadtomkiss-madetech, line 133: We have a mix of methods where some are returning Sequelize results and some are returning responses.

Do you think we should move the Sequelize methods into the `models.user` class as custom functions?

i.e 

```
models.user.findByLogin(loginId)
```

server/routes/admin/approval/index.js
```diff
@@ -3,140 +3,212 @@ const express = require('express');
...
+const _rejectNewUser = async (user, workplace, res) => {
+  // TODO: Email saying they've been rejected
+  // Remove the user
+  try {
+    if (user && workplace) {
+      const deleteduser = await user.destroy();
+      const deletedworkplace = await workplace.destroy();
+      if (deleteduser && deletedworkplace) {
+        return res.status(200).json({ status: '0', message: userRejectionConfirmation });
+      } else {
```
chadtomkiss-madetech, line 165: IMO `else's` are a code smell and can usually be avoided - especially when you're returning.

Check out https://medium.com/better-programming/why-you-need-to-stop-using-else-statements-5b1fd09dea9e which has some good examples.

By negating an `if` you can return early so that the main body of a function (no indentation) is the success route / main purpose.

PR #2116, opened 2020-05-05
Delete broken tests

chadtomkiss-madetech:
**Work done**

- Removed broken tests from Angular - these were specs that had components which had been changed but the test hadn't been updated

- Added `npm run test-headless` to `.travis.yml`


.travis.yml
```diff
@@ -38,7 +38,7 @@ before_script:
     - echo $B_NAME
 script:
   - npm run build
-  # - npm run test
+  - npm run test
```
aaron-russell, line 41: Should this be `npm run test headless`?

PR #2119, opened 2020-05-06
Fix: Duplicate Location ID for Bulk Upload

chadtomkiss-madetech:
**Work done**

- Added `getDuplicateLocationError()` method to Establishment CSV model
- Added check to ensure location IDs are unique on the `establishments/bulkUpload/validate` endpoint
- Removed a bunch of unused code from `server/test/unit/models/Bulkimport/csv/establishments.spec.js`

**Tests**

The original plan was to update the Establishment CSV model to copy the cross validation with Workers CSV, but in the end I moved the check to the API endpoint (as this was closer to the check for duplicate `LOCALESTID`).

Have tried running the API endpoint tests but I get an SNS error which I think is to post to Slack for new registrations.


aaron-russell:
The tests should run @chadtomkiss-madetech you may want to run that individual file first

chadtomkiss-madetech:
@aaron-russell

I've moved the duplicate location check into it's own function and exported it for testing.

Have also added a factory for generating test CSV data.

PR #2130, opened 2020-05-13
Duplicate Worker IDs fix

two-first-names:
- Fixes unique and change unique worker ID checking.
- Testing for the following scenarios:
  - Changing ID to an already existing ID
  - Adding two IDs of the same value
  - Changing to two differing IDs


aaron-russell:
[(2) BU Validation - Error - Duplicate workerID](https://trello.com/c/YsEBompx/35-2-bu-validation-error-duplicate-workerid)

server/test/unit/routes/establishments/bulkUpload.spec.js
```diff
@@ -51,6 +56,188 @@ describe('/server/routes/establishment/bulkUpload.js', () => {
...
+          myWorkers[1], allKeys, changeKeyNoWhitespace, keyNoWhitespace, allWorkersByKey, myAPIWorkers, csvWorkerSchemaErrors
+        )) {
+          allWorkersByKey[keyNoWhitespace] = thisWorker.lineNumber;
+
+          // to prevent subsequent Worker duplicates, add also the change worker id if CHGUNIQUEWORKERID is given
+          if (changeKeyNoWhitespace) {
+            allWorkersByKey[changeKeyNoWhitespace] = thisWorker.lineNumber;
+          }
+        }
+      });
```
chadtomkiss-madetech, line 234: This loop is duplicated in both tests - would it be worth moving to a function for future tests to run the same `bulkUpload.checkDuplicateWorkerID` code?

PR #2141, opened 2020-05-18
Updated dependancies to the latest versions

aaron-russell:
Had to keep back the following libraries due to them not been compatible:
- seqelize
- uuid
- govuk-frontend

If we want to update these we will need to refactor some code.

We currently still have 4 security vulnerabilities. 2 low and 2 high. However, there is currently no fix for these.

#### Tests
- [ ] Yes


aaron-russell:
[Dependency updates](https://trello.com/c/MaKSoyNG/45-dependency-updates)

PR #2146, opened 2020-05-19
Fix: Alignment on training by category table

chadtomkiss-madetech:
**Work done**

Fix issue where table cells were aligned in the middle, but needed to be aligned at the top - this was an issue when a training category had multiple training statuses (expired, expiring soon etc)

#### Tests
- [ ] Yes


chadtomkiss-madetech:
thanks to @two-first-names for fishing out the class I needed to add 👍

PR #2147, opened 2020-05-19
Last Updated Training

two-first-names:
#### Tests
- [X] Yes

### DIFF
--- a/src/app/shared/components/trianing-link-panel/training-link-panel.component.spec.ts
+++ b/src/app/shared/components/trianing-link-panel/training-link-panel.component.spec.ts
@@ -1,0 +1,60 @@
+import { DashboardComponent } from '@features/dashboard/dashboard.component';
+import { render, within } from '@testing-library/angular';
+import { HomeTabComponent } from '@features/dashboard/home-tab/home-tab.component';
+import { SharedModule } from '@shared/shared.module';
+import { Router, RouterModule } from '@angular/router';
+import { WindowRef } from '@core/services/window.ref';
+import { RouterTestingModule } from '@angular/router/testing';
+import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { PermissionsService } from '@core/services/permissions/permissions.service';
+import { UserService } from '@core/services/user.service';
+import { NotificationsService } from '@core/services/notifications/notifications.service';
+import { MockPermissionsService } from '@core/test-utils/MockPermissionsService';
+import { MockNotificationsService } from '@core/test-utils/MockNotificationsService';
+import { MockUserService } from '@core/test-utils/MockUserService';
+import { HttpClient } from '@angular/common/http';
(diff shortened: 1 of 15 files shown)


src/app/shared/components/trianing-link-panel/trianing-link-panel.component.html
```diff
@@ -1,25 +1,4 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-one-third">
```
aaron-russell, line 2: Are these links no longer needed?

two-first-names: Yeah, was discussed in meeting after standup this morning.

src/app/shared/components/trianing-link-panel/trianing-link-panel.component.component.ts
```diff
@@ -28,10 +37,29 @@ export class TrainingLinkPanelComponent implements OnInit, OnDestroy {
...
-        if (data && data.id) {
+        if (data && data.id !== undefined) {
           this.establishmentUid = data.uid;
+          this.subscriptions.add(
+            this.workerService.getAllWorkers(data.uid).subscribe(
+              workers => {
+                workers.forEach((worker: Worker) => {
+                  if (worker.trainingCount > 0) {
+                    this.subscriptions.add(
+                      this.workerService.getTrainingRecords(data.uid, worker.uid)
```
chadtomkiss-madetech, line 51: There's a lot of HTTP requests to get training records for every worker that's shown.

`getAllWorkers` on L46 also loops over each worker on the server side to get all of the training records them.

Not sure if I'm over optimising but do you think amending the server side response to include the latest updated record date would be a more performant option?

src/app/shared/components/trianing-link-panel/training-link-panel.component.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    const router = injector.get(Router) as Router;
+
+    return {
+      component,
+      establishmentService
+    };
+  }
+
+  it('should render a TrainingLinkPanelComponent', async () => {
+    const component = await setup();
```
chadtomkiss-madetech, line 57: This doesn't destructure like in the test on L62

`setup` returns an object with both the component and the establishment service so I'm not sure the naming `component` is technically correct.

two-first-names: Fair!

src/app/shared/components/trianing-link-panel/trianing-link-panel.component.component.ts
two-first-names: Worker does have a lastUpdated on it, but no idea if that relates to training or the worker as a whole.

PR #2157, opened 2020-05-20
Feat: Add schema for Become a Parent Request

chadtomkiss-madetech:
**Work done**

- Added `Sequelize-cli` for generating and running migrations
- Added migration for `Approvals`

#### Tests
- [ ] Yes

### DIFF
--- a/migrations/20200519170545-create_approvals_table.js
+++ b/migrations/20200519170545-create_approvals_table.js
@@ -1,0 +1,48 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('Approvals', {
+      EtablishmentID: {
+        type: Sequelize.DataTypes.INTEGER,
+        references: {
+          model: {
+            tableName: 'Establishment',
+            schema: 'cqc'
+          },
+          key: 'EstablishmentID'
+        },
+        allowNull: false
+      },
+      UserID: {
+        type: Sequelize.DataTypes.INTEGER,
+        references: {
+          model: {
+            tableName: 'User',
+            schema: 'cqc'
+          },
+          key: 'RegistrationID'
+        },
+        allowNull: false
+      },
+      RejectionReason: {
+        type: Sequelize.DataTypes.TEXT,
+        allowNull: true,
+      },
+      Data: {
+        type: Sequelize.DataTypes.JSON,
+        allowNull: true,
+      },
+      createdAt: Sequelize.DataTypes.DATE,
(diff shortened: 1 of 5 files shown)


migrations/20200519170545-create_approvals_table.js
```diff
@@ -0,0 +1,50 @@
...
+        references: {
+          model: {
+            tableName: 'Establishment',
+            schema: 'cqc'
+          },
+          key: 'EstablishmentID'
+        },
+        allowNull: false
+      },
+      UserID: {
```
aaron-russell, line 17: Should we match the column name here?

migrations/20200519170545-create_approvals_table.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable('Approvals', {
+      EtablishmentID: {
```
aaron-russell, line 6: You've spelled Establishment wrong

server/config/sequelize.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = {
+  development: {
+    username: appConfig.get('db.username'),
+    password: appConfig.get('db.password'),
+    database: appConfig.get('db.database'),
+    host: appConfig.get('db.host'),
+    port: appConfig.get('db.port'),
+    dialect: appConfig.get('db.dialect'),
+    migrationStorageTableSchema: 'cqc',
```
aaron-russell, line 11: Can this be set by appConfig too?

package.json
```diff
@@ -76,6 +76,7 @@
     "rxjs": "^6.5.2",
     "rxjs-compat": "^6.5.2",
     "sequelize": "^4.44.3",
+    "sequelize-cli": "^5.5.1",
```
aaron-russell, line 79: Is the version mismatch between the cli and sequelize going to cause an issue?

chadtomkiss-madetech: `sequelize --help` returns Sequelize CLI [Node: 13.8.0, CLI: 5.5.1, ORM: 4.44.3]

When creating a migration, you get:

```
up: (queryInterface, Sequelize)
```

The `Sequelize` argument that's passed is our installed 4.44.3 version - so as long as any Sequelize methods used in a migration are available in the version we use, we're good.

server/config/sequelize.js
chadtomkiss-madetech: This isn't defined in `appConfig`, and is also hardcoded in the models that are setup.

PR #2165, opened 2020-05-26
Feat: Add become a parent endpoint

chadtomkiss-madetech:
**Work done**

- Added endpoint to create a BecomeAParent Approval request
- Added functions to Approvals model for abstracting queries
- Added tests

#### Tests
- [x] Yes

### DIFF
--- a/server.js
+++ b/server.js
@@ -47,6 +47,7 @@ var serviceUsers = require('./server/routes/serviceUsers');
 var workingTrainingCategories = require('./server/routes/workerTrainingCategories');
 var nurseSpecialism = require('./server/routes/nurseSpecialism');
 var availableQualifications = require('./server/routes/availableQualifications');
+var approvals = require('./server/routes/approvals');
 
 // admin route
 var admin = require('./server/routes/admin');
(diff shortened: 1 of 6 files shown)


server.js
```diff
@@ -47,6 +47,7 @@ var serviceUsers = require('./server/routes/serviceUsers');
 var workingTrainingCategories = require('./server/routes/workerTrainingCategories');
 var nurseSpecialism = require('./server/routes/nurseSpecialism');
 var availableQualifications = require('./server/routes/availableQualifications');
+var approvals = require('./server/routes/approvals');
```
aaron-russell, line 50: Why var?

server/routes/approvals/index.js
```diff
@@ -0,0 +1,9 @@
+// default route for admin
+const express = require('express');
+const router = express.Router();
+
+const becomeAparent = require('./becomeAParent');
```
aaron-russell, line 5: This casing...

server.js
chadtomkiss-madetech: @aaron-russell Using var was just to align with the rest of the variables.. should they all be changed to const? If so, I think actioning that as a sep PR would be better given it'd be changing unrelated things.

PR #2166, opened 2020-05-27
Feat/approve parent

claresudbery:
#### Tests
- [ ] Yes


src/app/features/search/search.component.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+    establishment: {
+      uid: testEstablishmentUid,
+      name: testOrgname
+    }
+  }
+];
+
+describe('SearchComponent', () => {
+
+  // Everything commented out for now because of the error "Property 'switchWorkplaceService' does not exist on type 'WrapperComponent'"
```
aaron-russell, line 38: Don't like these comments been left in... though I don't know what to do with them. I'm assuming this will not be looked at again?

server/routes/admin/parent-approval/index.js
```diff
@@ -0,0 +1,92 @@
+// default route for admin/approval
```
chadtomkiss-madetech, line 1: Don't think this comment is needed

server/routes/admin/parent-approval/index.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const _updateApprovalStatus = async (approvalId, status) => {
+  let singleApproval = await models.Approvals.findbyId(approvalId);
+  singleApproval.Status = status;
+  await singleApproval.save();
+};
+
+const _makeWorkplaceIntoParent = async (id) => {
+  let workplace = await models.establishment.findbyId(id);
```
chadtomkiss-madetech, line 71: What happens if, for example, a workplace is deleted by another user before they hit approve/reject?

I think we should check that we get a result returned by the queries that are run.

claresudbery:
Acted on feedback

claresudbery:
[Become a parent workplace - Administrator can approve or reject request](https://trello.com/c/k57nBFYF/75-become-a-parent-workplace-administrator-can-approve-or-reject-request)

PR #2167, opened 2020-05-27
feat/parent-service: Service has new method to create new parent requ…

claresudbery:
…ests.

#### Tests
- [ ] Yes
- [x ] No, they are not required for this change

### DIFF
--- a/src/app/core/services/parent-requests.service.ts
+++ b/src/app/core/services/parent-requests.service.ts
@@ -18,7 +18,7 @@ export class ParentRequestsService {
     return this.http.post<any>('/api/admin/parent-approval/', data);
   }
 
-  public becomeParent(data: object) {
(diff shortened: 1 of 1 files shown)


src/app/core/services/parent-requests.service.ts
```diff
@@ -17,4 +17,8 @@ export class ParentRequestsService {
   public parentApproval(data: object) {
     return this.http.post<any>('/api/admin/parent-approval/', data);
   }
+
+  public becomeParent(data: object) {
```
chadtomkiss-madetech, line 21: I don't think this data object is needed as nothing will be sent given it's determined by the token.

chadtomkiss-madetech: Oh, it seems http POST wants data ;D

claresudbery: OK fixed

PR #2172, opened 2020-05-28
Fix: Worker training categories 404

chadtomkiss-madetech:
**Issue**

When trying to get the training by category and there's no results returned from the database, it sends a 404, but it was sending back HTML instead of JSON.

#### Tests
- [x] Yes


claresudbery:
[BUG: New workplaces display no data if they have no training records](https://trello.com/c/CgUm8CGF/108-bug-new-workplaces-display-no-data-if-they-have-no-training-records)

PR #2176, opened 2020-05-28
Feat/parent request order

claresudbery:
#### Tests
- [ ] Yes

### DIFF
--- a/server/models/approvals.js
+++ b/server/models/approvals.js
@@ -78,6 +78,9 @@ module.exports = (sequelize, DataTypes) => {
           as: 'User',
           attributes: ['FullNameValue']
         }
+      ],
+      order: [
+        ['createdAt', 'DESC']
       ]
     });
   }
@@ -99,6 +102,9 @@ module.exports = (sequelize, DataTypes) => {
           as: 'User',
           attributes: ['FullNameValue']
         }
+      ],
+      order: [
+        ['createdAt', 'DESC']
       ]
     });
   }
@@ -120,6 +126,9 @@ module.exports = (sequelize, DataTypes) => {
           as: 'User',
           attributes: ['FullNameValue']
         }
+      ],
+      order: [
+        ['createdAt', 'DESC']
       ]
     });
   }
--- a/server/models/approvals.js
+++ b/server/models/approvals.js
@@ -103,9 +103,6 @@ module.exports = (sequelize, DataTypes) => {
           attributes: ['FullNameValue']
         }
       ],
-      order: [
-        ['createdAt', 'DESC']
-      ]
     });
   }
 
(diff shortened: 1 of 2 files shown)


server/models/approvals.js
```diff
@@ -99,6 +102,9 @@ module.exports = (sequelize, DataTypes) => {
           as: 'User',
           attributes: ['FullNameValue']
         }
+      ],
+      order: [
+        ['createdAt', 'DESC']
```
chadtomkiss-madetech, line 107: This query will only return 1 result so doesn't need ordering.

server/models/approvals.js
```diff
@@ -120,6 +126,9 @@ module.exports = (sequelize, DataTypes) => {
...
```
chadtomkiss-madetech, line 131: Same here.. shouldn't need to do any ordering for this query.

src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -79,7 +79,7 @@ <h3 class="govuk-heading-s govuk-!-margin-top-1">Related content</h3>
         <li *ngIf="canViewDataPermissionsLink">
           <a (click)="setDataPermissions($event)" href="#">Set data permissions</a>
         </li>
-        <li *ngIf="canLinkToParent">
+        <li *ngIf="canLinkToParent && !workplace.isParent">
```
aaron-russell, line 82: This is done on the back end already

src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -90,7 +90,7 @@ <h3 class="govuk-heading-s govuk-!-margin-top-1">Related content</h3>
         <li *ngIf="canRemoveParentAssociation">
           <a (click)="removeLinkToParent($event)" href="#">Remove link to my parent organisation</a>
         </li>
-        <li *ngIf="canBecomeAParent && !linkToParentRequestedStatus">
+        <li *ngIf="canBecomeAParent && !linkToParentRequestedStatus && !workplace.isParent">
```
aaron-russell, line 93: Same here

src/app/features/dashboard/home-tab/home-tab.component.html
chadtomkiss-madetech: @aaron-russell The permission check always returns true when you're an admin. So this change is to hide the links from Admins if the workplace they are viewing is already a parent.

aaron-russell: @chadtomkiss-madetech that plays into what I said the other day, Permissions is broken... That shouldn't be possible

PR #2186, opened 2020-06-02
Fixed bug where pending workplaces weren't been shown without at least 1 workplace

aaron-russell:
#### Tests
- [ ] Yes


aaron-russell:
[[## 19336 ##] Pending status of a subsidiary registration](https://trello.com/c/xKWroP4Y/110-19336-pending-status-of-a-subsidiary-registration)

chadtomkiss-madetech:
@aaron-russell I think there's multiple if statements in this component that could be tested to check we are being shown the expected content? Especially as it was categorised as a bug.

PR #2189, opened 2020-06-02
Feat: Allow admins to set the location id to something not in the DB

chadtomkiss-madetech:
**Context**

We want admins to be able to set the Location ID to something that does not exist in the database.

The reason we want to set it to a non existent location is because there may be a known Location ID in the CQC dataset which does not yet exist in our database.

**Work done**

Added the ability for admins to set the location ID to something that doesn't exist in the database.

The API will return the current address for the Establishment along with the LocationID that was searched for but not found.

As a result, the UI will behave as normal, and set the Location ID on the Workplace to something that doesn't exist.

#### Tests
- [x] Yes

### DIFF
--- a/server/routes/locations.js
+++ b/server/routes/locations.js
@@ -19,6 +19,45 @@ const getLocations = async (req, res, matching) => {
     locationData.push(createLocationDetailsObject(data));
   }
 
+  // If the user is an Admin and the Location was not found, we want them to be able to use the location ID that they searched for.
+  if (locationData.length === 0 && req.role === 'Admin') {
+    const establishment = await models.establishment.findByPk(
+      req.establishment.id,
+      {
+        attributes: [
+          'NameValue',
+          'address1',
+          'address2',
+          'town',
+          'county',
+          'postcode',
+          'isRegulated',
+        ],
+        include: [
+          {
+            model: models.services,
+            as: 'mainService',
+            attributes: ['name'],
+          },
+        ],
+      }
+    );
+
+    const data = {
+      locationid: req.params.locationId,
+      locationname: establishment.NameValue,
+      addressline1: establishment.address1,
+      addressline2: establishment.address2,
+      towncity: establishment.town,
+      county: establishment.county,
+      postalcode: establishment.postcode,
+      isRegulated: establishment.isRegulated,
+      ...(establishment.mainService) && { mainservice: establishment.mainService.name }
+    };
+
+    locationData.push(createLocationDetailsObject(data));
+  }
+
   if (matching) {
     let currentEstablishments = await models.establishment.findAll({
       where: {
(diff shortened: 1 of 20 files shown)


server/test/unit/routes/locations.spec.js
```diff
@@ -71,7 +79,103 @@ describe('locations route', () => {
...
+          address1: '63 Skills for Care Lane',
+          address2: '',
+          town: 'Leeds',
+          county: 'North West',
+          postcode: 'LS1 3BE',
+          isRegulated: '123',
+          mainService: {
+            name: '123',
+          },
+        };
```
chadtomkiss-madetech, line 99: Use `test-data-bot` for this

server/test/unit/routes/locations.spec.js
```diff
@@ -71,7 +79,103 @@ describe('locations route', () => {
...
+      });
```
chadtomkiss-madetech, line 153: Use `test-data-bot` for this

server/test/unit/routes/locations.spec.js
```diff
@@ -71,7 +79,103 @@ describe('locations route', () => {
         }
       }, {status: updateStatus, json: updateJson, send: updateJson}, true);
     });
+
+    it('should return a non existent location if the user is an admin', async () => {
```
chadtomkiss-madetech, line 83: This test name could make it clearer that it returns the establishments current location

server/routes/locations.js
```diff
@@ -32,8 +32,53 @@ const getLocations = async (req, res, matching) => {
...
+      const data = {
+        locationid: req.params.locationId,
+        locationname: establishment.NameValue,
+        addressline1: establishment.address1,
+        addressline2: establishment.address2,
+        towncity: establishment.town,
+        county: establishment.county,
+        postalcode: establishment.postcode,
+        isRegulated: establishment.isRegulated,
+        ...(establishment.mainService) && { mainservice: establishment.mainService.name }
```
aaron-russell, line 68: Not quite sure what the && here is doing?

server/test/unit/routes/locations.spec.js
aaron-russell: I agree, test name doesn't make sense, especially as establishment does exist, just not in CQC (As of yet)

server/routes/locations.js
chadtomkiss-madetech: So this is a way to conditionally add key/values to an object.

Rather than:

```
const data = {
    name: 'Workplace Name',
}

if (establishment.mainservice)
  data.mainservice = establishment.mainservice.name;
}

if (establishment.Location) {
  data.location = establishment.location.name
}
```

```
const data = {
  name: 'Workplace Name',
  ...(establishment.mainService) && { mainservice: establishment.mainService.name },
  ...(establishment.Location) && { location: establishment.Location.name },
}
```

You can read more about the technique here: https://medium.com/@mikeh91/conditionally-adding-keys-to-javascript-objects-using-spread-operators-and-short-circuit-evaluation-acf157488ede

aaron-russell: Didn't know that, thank you, that's cool!

PR #2192, opened 2020-06-04
Remove unused capacity values

aaron-russell:
# Changes
- Added a new util for generating the correct capacities
- Added the util to the endpoints for changing a main service and other services

#### Tests
- [X] Yes

### DIFF
--- a/server/test/unit/utils/correctCapacities.spec.js
+++ b/server/test/unit/utils/correctCapacities.spec.js
@@ -83,7 +83,7 @@ describe('correctCapacities util', () => {
           }
         }
       });
-      const capacities = await correctCapacities(establishment, null, null);
+      const capacities = await correctCapacities(establishment);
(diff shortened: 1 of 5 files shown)


server/test/unit/utils/correctCapacities.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  question: 'Number of people using the service on the completion date',
+  service:
+  {
+    id: 9,
+    category: 'Adult day',
+    name: 'Day care and day services'
+  }
+}];
+
+sinon.stub(models.establishmentCapacity, 'findAll').callsFake(async (args) => {
```
chadtomkiss-madetech, line 68: stubs need to go in a `beforeEach` and you need a `sinon.restore()` in an `afterEach()`

Otherwise we will get the already stubbed error when running the full suite (if other tests have stubbed findAll etc)

PR #2213, opened 2020-06-11
Feat/generic approval

claresudbery:
#### Tests
- [x ] Yes


chadtomkiss-madetech:
@claresudbery 

Do you think that `userName` is confusing because it's the user's actual name but `username` is widely known as a login variable?

What if we had:

```
user: {
  name: 'Mary Poppins',
}
```

So it's `user.name` to access their name instead.

PR #2218, opened 2020-06-16
[## 16372 ##] Bulk upload pre populating the UNIQUEWORKERID on the training file

two-first-names:
Uses Worker.localIdentifier instead of Worker.key.

#### Tests
- [x] Yes


aaron-russell:
Can I help get this unit tested @two-first-names ?

server/test/unit/routes/establishments/bulkUpload.spec.js
```diff
@@ -480,4 +487,86 @@ describe('/server/routes/establishment/bulkUpload.js', () => {
...
+        ]
+      }, true);
+
+      await bulkUpload.exportToCsv('', [establishment], 'foo', 'training', [
+        {
+          maxqualifications: "10"
+        }
+      ], responseSend);
+
+      expect(lines[1]).to.equal(`${establishment.name},,${BUDI.trainingCategory(BUDI.FROM_ASC, 1)},,,,,`);
```
chadtomkiss-madetech, line 528: Does the `1` here related to the training id? Might be worth assigning it to a variable to give it meaning.

two-first-names: Category ID, good catch.

PR #2221, opened 2020-06-16
Feat: CQC Status Change > Develop

chadtomkiss-madetech:


src/app/features/workplace/select-main-service/select-main-service.component.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    return {
+      component
+    };
+  }
+
+  it('should render', async () => {
+    const { component } = await setup();
+
+    expect(component).toBeTruthy();
+  });
```
aaron-russell, line 38: Should there be more tests around this?

PR #2231, opened 2020-06-17
Added slack alert if there's a large bulk upload

aaron-russell:
#### Tests
- [X] Yes

### DIFF
--- a/server/test/unit/routes/establishments/bulkUpload.spec.js
+++ b/server/test/unit/routes/establishments/bulkUpload.spec.js
@@ -5,6 +5,7 @@ const sinon = require('sinon');
 const rfr = require('rfr');
 const { build, fake } = require('@jackfranklin/test-data-bot');
 
+const slack = rfr('server/utils/slack/slack-logger');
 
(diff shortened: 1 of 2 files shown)


server/test/unit/routes/establishments/bulkUpload.spec.js
```diff
@@ -5,6 +5,7 @@ const sinon = require('sinon');
 const rfr = require('rfr');
 const { build, fake } = require('@jackfranklin/test-data-bot');
 
+const slack = rfr('server/utils/slack/slack-logger');
 
 const dbmodels = rfr('server/models');
 sinon.stub(dbmodels.status, 'ready').value(false);
```
g-m-taylor, line 11: This isnt so much a review as a question to the Devs(as I know this code wasnt added on the PR but :  wont this stub bleed into other tests? as its outside of the describe?

aaron-russell: I think `sinon.restore() ` will also include that one, but then it wont be reinitiated.

PR #2251, opened 2020-06-23
Adding Sentry, disabling datadog

two-first-names:
#### Tests
- [ ] Yes


server/utils/security/isAuthenticated.js
```diff
@@ -23,8 +23,7 @@ exports.isAuthorised = (req, res , next) => {
         req.username = claim.sub;
         req.userUid = claim.userUid;
         req.user = {
-          id: claim.userUid,
-          username: claim.sub
+          id: claim.userUid
```
aaron-russell, line 26: What is this doing? This seems like a large change to the JWT?

two-first-names: It's putting the user ID in `req.user.id`, this is mostly so Honeycomb and Sentry can have user context, since they get it from the express request.

If we prefer, we can not pass this to them at all, but I think it'd be useful to correlate requests by users.

PR #2266, opened 2020-06-26
Changed server command to start with nodemon running.

aaron-russell:
This means when you're building anything, it will autorefresh the frontend and autorestart the back end!

#### Tests
- [ ] Yes


chadtomkiss-madetech:
@aaron-russell Do you think we should add nodemon as a dev dependency so devs don't have to install it globally?

aaron-russell:
> 
> 
> @aaron-russell Do you think we should add nodemon as a dev dependency so devs don't have to install it globally?

It already is!

PR #2273, opened 2020-06-26
Updated the postcode lookup to only sho LA if only 1 returned

aaron-russell:
- Removed limit of 1 from the select for first half of postcode
- Only returned la if only 1 in array

#### Tests
- [X] Yes

### DIFF
--- a/server/routes/reports/localAuthorityReport/user.js
+++ b/server/routes/reports/localAuthorityReport/user.js
@@ -112,13 +112,13 @@ const identifyLocalAuthority = async postcode => {
 
   // must escape the string to prevent SQL injection
   const fuzzyCssrIdMatch = await models.sequelize.query(
-    `select "Cssr"."CssrID", "Cssr"."CssR" from cqcref.pcodedata, cqc."Cssr" where postcode like \'${escape(firstHalfOfPostcode)}%\' and pcodedata.local_custodian_code = "Cssr"."LocalCustodianCode" group by "Cssr"."CssrID", "Cssr"."CssR" limit 1`,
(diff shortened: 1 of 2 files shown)


server/routes/reports/localAuthorityReport/user.js
```diff
@@ -112,13 +112,13 @@ const identifyLocalAuthority = async postcode => {
...
   const fuzzyCssrIdMatch = await models.sequelize.query(
-    `select "Cssr"."CssrID", "Cssr"."CssR" from cqcref.pcodedata, cqc."Cssr" where postcode like \'${escape(firstHalfOfPostcode)}%\' and pcodedata.local_custodian_code = "Cssr"."LocalCustodianCode" group by "Cssr"."CssrID", "Cssr"."CssR" limit 1`,
+    `select "Cssr"."CssrID", "Cssr"."CssR" from cqcref.pcodedata, cqc."Cssr" where postcode like \'${escape(firstHalfOfPostcode)}%\' and pcodedata.local_custodian_code = "Cssr"."LocalCustodianCode" group by "Cssr"."CssrID", "Cssr"."CssR"`,
     {
       type: models.sequelize.QueryTypes.SELECT
     }
   );
 
-  if (fuzzyCssrIdMatch && fuzzyCssrIdMatch[0] && fuzzyCssrIdMatch[0] && fuzzyCssrIdMatch[0].CssrID) {
+  if (fuzzyCssrIdMatch && fuzzyCssrIdMatch[0] && fuzzyCssrIdMatch.length === 1 && fuzzyCssrIdMatch[0].CssrID) {
```
chadtomkiss-madetech, line 121: Do you think we need `.length` on `fuzzyCssrIdMatch` if `fuzzyCssrIdMatch[0]` is already checking there's an element in the array?

PR #2277, opened 2020-06-30
Live > Develop

chadtomkiss-madetech:
Opening this PR to discuss changes that are on live but not in develop

#### Tests
- [ ] Yes


server/config/config.js
```diff
@@ -502,9 +502,9 @@ if (config.get('aws.secrets.use')) {
     // DB rebind
     config.set('db.host', AWSSecrets.dbHost());
     config.set('db.password', AWSSecrets.dbPass());
-    config.set('db.client_ssl.data.certificate', AWSSecrets.dbAppUserCertificate().replace(/\\n/g, "\n"));
-    config.set('db.client_ssl.data.key', AWSSecrets.dbAppUserKey().replace(/\\n/g, "\n"));
-    config.set('db.client_ssl.data.ca', AWSSecrets.dbAppRootCertificate().replace(/\\n/g, "\n"));
+    //config.set('db.client_ssl.data.certificate', AWSSecrets.dbAppUserCertificate().replace(/\\n/g, "\n"));
+    //config.set('db.client_ssl.data.key', AWSSecrets.dbAppUserKey().replace(/\\n/g, "\n"));
+    //config.set('db.client_ssl.data.ca', AWSSecrets.dbAppRootCertificate().replace(/\\n/g, "\n"));
```
aaron-russell, line 507: This can't happen on dev

chadtomkiss-madetech: @aaron-russell Should we do something similar to

```
config.get('aws.secrets.use')
```

Something like:

```
if (config.get('db.use_ssl')) {
}
```

aaron-russell: All env's use Secrets, only some use SSL

PR #2281, opened 2020-07-01
Feat/l areport polling

g-m-taylor:
#### Tests
- [ ] Yes
Tried to add tests to the polling service but couldnt figure out what to test the functions individually. Most of it is S3, or database functions.  Spent a day working on it, after which Aaron and I decided it was taking up too much time.  Will have a look again in the next ticket. 


server/utils/fileLock.js
```diff
@@ -0,0 +1,193 @@
...
+  region: String(config.get('bulkupload.region')),
+});
+const Bucket = String(config.get('bulkupload.bucketname'));
+
+// Prevent multiple report requests from being ongoing simultaneously so we can store what was previously the http responses in the S3 bucket
+// This function can't be an express middleware as it needs to run both before and after the regular logic
+
+const reportsAvailable = ['la','training'];
+
+const acquireLock = async function(report,logic, req, res) {
```
aaron-russell, line 15: Noticed there's a spacing incosistencey (I know it's petty)

server/utils/fileLock.js
```diff
@@ -0,0 +1,193 @@
...
+        throw new Error('Response code was not numeric');
+      }
+    })
+    .catch(err => {
+      console.error('Report::responseGet: getting data returned an error:', err);
+
+      res.status(404).send({
+        message: 'Not Found',
+      });
+    });
```
aaron-russell, line 153: Why use then()? We should really use async await to keep the code consitent with the rest of the codebase

server/utils/fileLock.js
chadtomkiss-madetech: @aaron-russell If only there was a way to automate this 🙈 

We should definitely look into Prettier for formatting.

I've certainly held back from making formatting comments.

migrations/20200630081256-update_establishment.js
```diff
@@ -0,0 +1,29 @@
...
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.removeColumn(
+      {
+        tableName: 'Establishment',
+        schema: 'cqc'
+      },
+      'LaReportLockHeld',
+    );
+  }
+};
```
chadtomkiss-madetech, line 29: This will need a SQL file adding to the SFC-DB repo until we have migrations being run during deployment :(

server/routes/reports/localAuthorityReport/index.js
```diff
@@ -10,7 +10,8 @@ const userReport = require('./user');
 const adminReport = require('./admin');
 
 router.use('/admin', [isAdmin, adminReport]);
-router.use('/:id', [isAuthorisedEstablishment, userReport]);
+// router.use('/:id/report', [isAuthorisedEstablishment, userReport]);
```
chadtomkiss-madetech, line 13: Remove commented out route?

server/routes/reports/localAuthorityReport/user.js
```diff
@@ -1291,237 +1340,238 @@ const updateStaffRecordsSheet = (
...
+};
 
 module.exports = router;
-module.exports.LAReport = LAReport;
-module.exports.getReport = getReport;
 module.exports.identifyLocalAuthority = identifyLocalAuthority;
+router.route('/report').get(reportLock.acquireLock.bind(null, 'la', reportGet));
+router.route('/lockstatus').get(reportLock.lockStatusGet.bind(null, 'la'));
+router.route('/unlock').get(reportLock.releaseLock.bind(null, 'la'));
+router.route('/response/:buRequestId').get(reportLock.responseGet);
```
chadtomkiss-madetech, line 1577: Should this go before the `module.exports` stuff?

server/utils/fileLock.js
```diff
@@ -0,0 +1,193 @@
...
+const releaseLock = async (report,req, res, next) => {
+  const establishmentId = req.query.subEstId || req.establishmentId;
+  if (!reportsAvailable.includes(report)){
+    console.error('Lock *NOT* acquired.');
+    res.status(500).send({
+      message: `reportType not correct`,
+    });
+    return;
+  }
+  const LockHeldTitle = report + "ReportLockHeld";
```
chadtomkiss-madetech, line 76: This is dynamically building a column in the DB but I don't think that's clear - what does `report` contain?

server/utils/fileLock.js
```diff
@@ -0,0 +1,193 @@
...
+        await models.establishment.update(
+          {
+            [LockHeldTitle]: false,
+          },{
+            where: {
+              id: establishmentId,
+            }
+          });
+  }
+  if (res !== null) {
```
chadtomkiss-madetech, line 88: Would res ever be null?

server/utils/fileLock.js
```diff
@@ -0,0 +1,193 @@
...
+  });
+  res
+    .status(200) // don't allow this to be able to test if an establishment exists so always return a 200 response
+    .send(
+      currentLockState.length === 0
+        ? {
+          establishmentId,
+          reportLockHeld: true,
+        }
+        : currentLockState[0]
```
chadtomkiss-madetech, line 184: I think having ternary logic in send makes this hard to read

Does currentLockState[0] contain the same object?

You could do:

```
if (currentLockState.length > 0) {
  return res.status(200).send(currentLockState[0]);
}

return res.status(200).send({establishmentId, reportLockHeld: true});
```

server/utils/fileLock.js
g-m-taylor: ahh yeah we actually changed this together but it somehow ended up in my stash

server/utils/fileLock.js
g-m-taylor: Im not quite sure what you mean with this comment?

server/utils/fileLock.js
g-m-taylor: 🤷‍♀️  this was existing code. I assumed it was security stuff

server/utils/fileLock.js
g-m-taylor: isnt that doing something different though? its checking if length is > 0 rather than 0?

server/utils/fileLock.js
aaron-russell: I think it is difficult to read, maybe changing the name of that const in order to demonstrate what you're doing?

migrations/20200630081256-update_establishment.js
g-m-taylor: done

server/utils/fileLock.js
g-m-taylor: oh thats so it can be used as a func rather than api endpoint. But il remove it and we can put it in again when its needed

PR #2284, opened 2020-07-01
Fix: Add gzip compression

chadtomkiss-madetech:
**Work done**

- Added `compression` middleware in order to return gzip encoded responses.

#### Tests
- [ ] Yes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6136,7 +6136,6 @@
       "version": "2.0.18",
       "resolved": "https://registry.npmjs.org/compressible/-/compressible-2.0.18.tgz",
       "integrity": "sha512-AF3r7P5dWxL8MxyITRMlORQNaOA2IkAFaTr4k7BUumjPtRpGDTZpl0Pb1XCO6JeDCBdp126Cgs9sMxqSjgYyRg==",
-      "dev": true,
       "requires": {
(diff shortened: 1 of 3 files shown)


chadtomkiss-madetech:
> Is this going to be regression tested?

Do you think it needs to be?

It adds:

```
Content-Encoding: gzip
```

to the responses.

aaron-russell:
> 
> 
> > Is this going to be regression tested?
> 
> Do you think it needs to be?
> 
> It adds:
> 
> ```
> Content-Encoding: gzip
> ```
> 
> to the responses.

But it's also compressing the endpoints and the FE resources, also any reason we're not using brotil too?

chadtomkiss-madetech:
> > > Is this going to be regression tested?
> > 
> > 
> > Do you think it needs to be?
> > It adds:
> > ```
> > Content-Encoding: gzip
> > ```
> > 
> > 
> > to the responses.
> 
> But it's also compressing the endpoints and the FE resources, also any reason we're not using brotil too?

@aaron-russell 

They all have `X-Powered-By: Express` headers and the compression lib doesn't yet have support for Brotli :(

Have looked at sites like Twitter that return XHR requests of JSON and they are returned gzipped compressed too.

PR #2287, opened 2020-07-02
Feat/la report

g-m-taylor:
Merging LA Report into Develop, ready to be merged into test. 

#### Tests
- [ ] Yes


server/utils/fileLock.js
```diff
@@ -0,0 +1,172 @@
...
+    attributes: [['EstablishmentID', 'establishmentId'], [LockHeldTitle, 'reportLockHeld']],
+    where: {
+      id: establishmentId
+    }
+  }).then(res => {
+    return res.map(row => {
+      return row.dataValues;
+    });
+  });
+  if (currentLockState.length >= 0) {
```
chadtomkiss-madetech, line 162: I think this will return when it's both empty and has values.. is that intended behaviour?

chadtomkiss-madetech: Think it should be `currentLockState.length > 0`?

PR #2325, opened 2020-07-14
Feat: Add prod deploy config and manifest

chadtomkiss-madetech:
**Work done**

- Add production deployment step
- Add `manifest.prod.yml`
- Update `npm run build:prod` to only run on tagged builds.

#### Tests
- [ ] Yes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,7 @@ jobs:
             cache-path: ~/project/node_modules
             cache-version: v2
         - run: |
-            if [ "$CIRCLE_BRANCH" = "live" ]; then
+            if [ -n "$CIRCLE_TAG" ]; then
                 npm run build:prod
             else
                 npm run build
@@ -330,26 +330,21 @@ workflows:
             live_subdomain: sfcstaging
             live_domain: cloudapps.digital
             workspace_path: .
-        - hold_preprod:
-            type: approval
-            requires:
-              - test_frontend
-              - test_backend
-              - performance
-            filters:
-              branches:
-                only:
-                  - live
         - blue_green:
             name: preprod_deploy
             appname: sfcuatblue
             env_prefix: SFC_PREPROD
             requires:
-              - hold_preprod
+              - test_frontend
+              - test_backend
+              - performance
             filters:
+              # ignore any commit on any branch by default
               branches:
-                only:
-                  - live
+                ignore: /.*/
+              # only act on version tags
+              tags:
+                only: /^v[0-9]+(\.[0-9]+)*$/
             endpoint: https://api.cloud.service.gov.uk
             org: dhsc-skills-for-care-nmds-sc-2
             space: production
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             cache-path: ~/project/node_modules
             cache-version: v2
         - run: |
-            if [ "$CIRCLE_BRANCH" = "live" ]; then
+            if [ -n "$CIRCLE_TAG" ]; then
```
chadtomkiss-madetech, line 136: Changing the if statement to be based on the presence of a `CIRCLE_TAG` environment variable rather than the branch name being set to "live"

.circleci/config.yml
```diff
@@ -330,26 +330,21 @@ workflows:
...
-        - hold_preprod:
-            type: approval
-            requires:
-              - test_frontend
-              - test_backend
-              - performance
-            filters:
-              branches:
-                only:
-                  - live
```
chadtomkiss-madetech, line 342: Moved `hold_preprod` hold step to be a `hold_prod` hold step.

This will make pre-prod deploys automatic based on a new tag (and the approval step will be to deploy to production)

.circleci/config.yml
```diff
@@ -330,26 +330,21 @@ workflows:
...
+              - performance
             filters:
+              # ignore any commit on any branch by default
               branches:
-                only:
-                  - live
+                ignore: /.*/
+              # only act on version tags
+              tags:
+                only: /^v[0-9]+(\.[0-9]+)*$/
```
chadtomkiss-madetech, line 347: We have changed the filter for `preprod_deploy` to be based on a tag regex rather than the `live` branch

.circleci/config.yml
```diff
@@ -330,26 +330,21 @@ workflows:
...
-                  - live
         - blue_green:
             name: preprod_deploy
             appname: sfcuatblue
             env_prefix: SFC_PREPROD
             requires:
-              - hold_preprod
+              - test_frontend
+              - test_backend
+              - performance
```
chadtomkiss-madetech, line 340: We have changed the `requires` step so that the pre prod deployment now depends on the tests/performance jobs

.circleci/config.yml
```diff
@@ -360,7 +355,38 @@ workflows:
...
+            type: approval
+            requires:
+              - preprod_deploy
+            filters:
+              # ignore any commit on any branch by default
+              branches:
+                ignore: /.*/
+              # only act on version tags
+              tags:
+                only: /^v[0-9]+(\.[0-9]+)*$/
```
chadtomkiss-madetech, line 368: Added a hold step for production that will ignore all branches and run for matching tags

.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             cache-path: ~/project/node_modules
             cache-version: v2
         - run: |
```
chadtomkiss-madetech, line 135: `- run: |` this allows you to have multi line bash commands

PR #2328, opened 2020-07-14
Test > Live

chadtomkiss-madetech:
**Changes introduced**

- Undo config for CircleCI pre-prod/prod deployments
- Add `manifest.prod.yml` 

#### Tests
- [ ] Yes


manifest.prod.yml
```diff
@@ -0,0 +1,7 @@
+applications:
+- name: sfcuatgreen
+  memory: 8G
+  disk_quota: 4GB
+  instances: 10
+  buildpack: https://github.com/cloudfoundry/nodejs-buildpack.git#v1.6.32
+stack: cflinuxfs3
```
chadtomkiss-madetech, line 7: This manifest was only used by CircleCI so is fine to leave in

.circleci/config.yml
```diff
@@ -133,7 +133,7 @@ jobs:
             cache-path: ~/project/node_modules
             cache-version: v2
         - run: |
-            if [ "$CIRCLE_BRANCH" = "live" ]; then
+            if [ -n "$CIRCLE_TAG" ]; then
```
chadtomkiss-madetech, line 136: This is to only run `npm run build:prod` on tags, but no Workflow has a tag filter defined so this will always run `npm run build`

PR #2343, opened 2020-07-20
Test > Live

chadtomkiss-madetech:
**Changes introduced**

- Added config for pre-prod/prod deployment using CircleCI

#### Tests
- [ ] Yes


.circleci/config.yml
```diff
@@ -401,3 +424,38 @@ workflows:
...
+            requires:
+              - prod_deploy
+            filters:
+              # ignore any commit on any branch by default
+              branches:
+                ignore: /.*/
+              # only act on version tags
+              tags:
+                only: /^v[0-9]+(\.[0-9]+)*$/
+            env: production
```
aaron-russell, line 461: Bit concerned as what we have on live currently works?

chadtomkiss-madetech: @aaron-russell This job will trigger a Sentry release when a deployment to production happens.

Currently works for dev/staging etc.

It'll also run after the deployment has been successful so no issues from me

PR #2346, opened 2020-07-20
Fix: Check app exists before stopping it

chadtomkiss-madetech:
**Issue**

If you stop an app that does not exist, the cf CLI tool will exit with an error code - by using `|| echo` we can make sure the exit code is 0.

If the app already exists, and it fails to stop - then `cf delete -f` will force delete the app anyway.

#### Tests
- [ ] Yes


chadtomkiss-madetech:
If the app DOES exist, and it fails to stop - the rename will fail because the app already exists.

I think this is fine because CircleCI will fail to deploy.

PR #2350, opened 2020-07-21
Develop > test

g-m-taylor:
Sentry changes

#### Tests
- [ ] Yes
- [ x ] No, they are not required for this change


server/config/config.js
```diff
@@ -465,16 +465,25 @@ const config = convict({
     dsn: {
       doc: 'Sentry Endpoint',
       format: String,
-      default: '',
+      default: 'https://59c078b68dc0429aa404e59920f288fd@o409195.ingest.sentry.io/5281212',
```
chadtomkiss-madetech, line 468: If we have the default DSN set here I think we can remove it from `server/config/development.yaml` etc? Especially as we can also set it via `SENTRY_DSN` env vars too

server/config/test.yaml
```diff
@@ -50,3 +50,5 @@ bulkupload:
 public:
     download:
         baseurl: https://sfc-public-staging.s3.eu-west-2.amazonaws.com/public/download
+sentry:
+  dsn: https://59c078b68dc0429aa404e59920f288fd@o409195.ingest.sentry.io/5281212
```
aaron-russell, line 54: You don't need this if it's default in config.js

aaron-russell: I have approved it anyway, as it't not going to affect anything but I would open another PR to remove it

PR #2353, opened 2020-07-21
Fix: Conflicts

chadtomkiss-madetech:
**Issue**

In order to resolve some merge conflicts, we needed to merge `live` branch, but we were unable to because of some changes that were specific to `live` could not be on `develop`/`test` branches.

The changes was related to whether to use the client SSL certs to connect to the DB and as we are using GovPaaS in production but AWS in dev/staging, we don't require an SSL cert in production.

As a fix, I have wrapped the config in an if statement to check whether to use client ssl or not.

#### Tests
- [ ] Yes


src/app/features/reports/components/reports-header/reports-header.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">
```
g-m-taylor, line 3: Sorry forgot about this.  This needs to  <h1 class="govuk-heading-xl">

PR #2381, opened 2020-07-30
Added benchmarks database config 

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server/config/benchmarks.yaml
+++ b/server/config/benchmarks.yaml
@@ -18,9 +18,6 @@ db:
         min: 10
         max: 10
     ssl: true
-    client_ssl:
-        status: false
-        usingFiles: false
 
 slack:
     level : 3     # enables Slack notifications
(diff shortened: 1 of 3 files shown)


server/config/benchmarks.yaml
```diff
@@ -0,0 +1,46 @@
...
+      enabled: true
+      registrations: arn:aws:sns:eu-west-1:364648107127:registrations-dev
+      feedback: 	arn:aws:sns:eu-west-1:364648107127:feedback-dev
+
+db:
+    pool:
+        min: 10
+        max: 10
+    ssl: true
+    client_ssl:
```
chadtomkiss-madetech, line 21: I think you can remove the `client_ssl` block as this is default to the status to false and not use it

server/config/benchmarks.yaml
```diff
@@ -0,0 +1,46 @@
...
+    bucketname: sfc-bulkupload-dev
+    validation:
+        timeout: 300
+        storeIntermediaries: true
+    completion:
+        timeout: 180
+
+public:
+    download:
+        baseurl: https://sfc-public-dev.s3.eu-west-2.amazonaws.com/public/download
```
chadtomkiss-madetech, line 44: Same with the `baseurl` - it's default to this in the config

PR #2384, opened 2020-07-31
Added release drafter

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -12,6 +12,7 @@ categories:
       - 'bug'
   - title: '🧰 Maintenance'
     label: 'chore'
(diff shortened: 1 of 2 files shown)


.github/release-drafter.yml
```diff
@@ -0,0 +1,23 @@
+name-template: 'SKILLSFORCARE Version 1 Sprint '
+tag-template: 'v1'
```
chadtomkiss-madetech, line 2: Question on versioning - how will this set the next version of the release?

aaron-russell: It wouldn't - we would have to that before we approve it. Unless we change our versioning strategy

PR #2385, opened 2020-07-31
Fix: Validate staff totals in Establishment CSV

chadtomkiss-madetech:
#### Tests
- [ ] Yes


server/models/BulkImport/csv/establishments.js
```diff
@@ -1308,69 +1308,25 @@ class Establishment {
...
-      // Is the establishment only shared lives (code 19)?
-     notSharedLivesOnly = this._mainService.id !== 19;
-
-    // Is the establishment only head office services (code 16)?
-     notHeadOfficeOnly = this._mainService.id !== 16;
-    }
-    if (this._totalPermTemp === employedWorkers + nonEmployedWorkers) {
-      if (notHeadOfficeOnly) {
-        if (employedWorkers === 0) {
-          if(false) {
```
chadtomkiss-madetech, line 1345: None of these rules were being run and haven't been tested.

After speaking to Jackie, we felt that it would be best to remove them and approach them when we get round to implementing them properly.

server/models/BulkImport/csv/establishments.js
```diff
@@ -2140,7 +2096,7 @@ class Establishment {
     fetchMyEstablishmentsWorkers
   }) {
     // if establishment isn't being added or updated then exit early
-    if (!(['NEW', 'UPDATE'].includes(this._status))) {
+    if (!(['NEW', 'UPDATE', 'NOCHANGE'].includes(this._status))) {
```
chadtomkiss-madetech, line 2099: Even if a Workplace is set to `NOCHANGE` we want to validate that the staff match the totals provided.

server/test/factories/worker/csv.js
```diff
@@ -23,7 +23,7 @@ module.exports = build('WorkerCSV', {
     STARTDATE : moment(fake(f => f.date.past())).format('YYYY'),
     STARTINSECT : moment(fake(f => f.date.past())).format('YYYY'),
     APPRENTICE : '999',
-    EMPLSTATUS : '1',
+    EMPLSTATUS : 1,
```
chadtomkiss-madetech, line 26: The code expects this value to be an integer

aaron-russell: None of the other numbers are integer's is it worth either converting them all or parsing the int when needed?

server/test/unit/models/Bulkimport/csv/establishments-old.spec.js
```diff
@@ -0,0 +1,3230 @@
+const expect = require('chai').expect;
```
aaron-russell, line 1: Why old spec?

chadtomkiss-madetech: I wanted to avoid adding tests to a spec that had 3.5k lines and was prone to having new tests c&p what's there already.

I re-named so that this spec runs, but new tests will hopefully default to the new spec that have examples of using factories/sinon etc to set up Establishments rather than 100s of lines of CSV data.

server/test/factories/worker/csv.js
chadtomkiss-madetech: So I ran into the issue when writing the tests

 I wouldn't change them here unless we're confident that all of the code expects them to be integers (I have a feeling some code does `parseInt` and some doesn't) 

I would action as a sep PR as it'll potentially have a wider impact than just this change.

aaron-russell: Yeah, are we confident that this won't impact anything with this once changing only?

chadtomkiss-madetech: The only thing it would impact is the tests, and they would fail if this change introduced an issue.

Looking at the code, `parseInt(this._currentLine.EMPLSTATUS, 10);` seems to be used throughout.

PR #2400, opened 2020-08-11
Benchmarking Qualifications tile: getting actual values into the tile

g-m-taylor:
#### Tests
- [x] Yes

### DIFF
--- a/server/models/services.js
+++ b/server/models/services.js
@@ -62,6 +62,15 @@ module.exports = function(sequelize, DataTypes) {
       as:"benchmarksData"
     });
   };
+  const seniorCareWorker = 25;
+  const careWorker = 10;
+  const communitySupport = 11;
+  const employmentSupport = 12;
+  const adviceGuidance = 3;
+  const technician = 29;
+  const otherCare = 20;
+  const nurseAssistant = 16;
+  Services.careProvidingStaff = [seniorCareWorker, careWorker, communitySupport, employmentSupport, adviceGuidance, technician, otherCare, nurseAssistant];
 
 
   return Services;
--- a/server/models/services.js
+++ b/server/models/services.js
@@ -62,15 +62,17 @@ module.exports = function(sequelize, DataTypes) {
       as:"benchmarksData"
     });
   };
-  const seniorCareWorker = 25;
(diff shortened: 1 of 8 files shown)


server/models/services.js
```diff
@@ -62,6 +62,15 @@ module.exports = function(sequelize, DataTypes) {
...
   };
+  const seniorCareWorker = 25;
+  const careWorker = 10;
+  const communitySupport = 11;
+  const employmentSupport = 12;
+  const adviceGuidance = 3;
+  const technician = 29;
+  const otherCare = 20;
+  const nurseAssistant = 16;
+  Services.careProvidingStaff = [seniorCareWorker, careWorker, communitySupport, employmentSupport, adviceGuidance, technician, otherCare, nurseAssistant];
```
g-m-taylor, line 73: Wasn't entirely sure if this is the best way? I wanted to show what the numbers/IDs correlate to, rather than just have a list of IDs. Figured this would be better than adding an other column on the db?

server/models/worker.js
```diff
@@ -1064,5 +1059,18 @@ module.exports = function(sequelize, DataTypes) {
     });
   };
 
+  Worker.specificJobs = function (establishmentId,jobArray) {
```
g-m-taylor, line 1062: does this need an async / await?

server/models/services.js
chadtomkiss-madetech: You *could* use comments:

```
Services.careProvidingStaff = [
  25, // Senior Care Worker
  10 // Care Worker
];
```

I do get a bit concerned when IDs from the DB are hardcoded in the code - there's no guarantee they'll stay the same across all environments / databases.

If you were to add a column to the services table, would that make things more clear?

```
SELECT * FROM workers JOIN services ON services.id = workers.MainServiceFKValue WHERE services.isCareProvider = 1;
```

Also what if 90% of services were care providing services? Would having a JS array of 100s of IDs still make sense?

server/models/worker.js
chadtomkiss-madetech: @g-m-taylor I think so yes as Sequelize functions should return a promise

PR #2404, opened 2020-08-12
Benchmarks turnover: adding in actual values

g-m-taylor:
#### Tests
- [x] Yes

### DIFF
--- a/server/routes/establishments/benchmarks/index.js
+++ b/server/routes/establishments/benchmarks/index.js
@@ -16,6 +16,7 @@ router.route('/').get(async (req, res) => {
     if (tiles.includes('pay')) reply.tiles.pay = await pay(establishmentId);
     if (tiles.includes('sickness')) reply.tiles.sickness = await sickness(establishmentId);
     if (tiles.includes('qualifications')) reply.tiles.qualifications = await qualifications(establishmentId);
+    if (tiles.includes('turnover')) reply.tiles.turnover = await turnover(establishmentId);
 
     reply = await comparisonGroupData(reply, benchmarkComparisonGroup);
     return res.status(200).json(reply);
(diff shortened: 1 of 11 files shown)


server/routes/establishments/benchmarks/index.js
```diff
@@ -69,6 +70,41 @@ const pay = async (establishmentId) => {
...
+        percentOfPermTemp = (leavers / permTemptCount);
+      }else{
+        stateMessage = 'check-data';
+      }
+    } else{
+      stateMessage = 'no-data';
+    }
+  }else{
+    stateMessage = 'no-workers';
+  }
```
chadtomkiss-madetech, line 94: I think there's too many `if/else` branches here and that `percentOfPermTemp` calculation is in the deepest nest but is actually the main purpose of the function.

There's a good example on using guard clauses to tackle this problem here: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

g-m-taylor: Ok, I had to create an other function to do this, but have a look and see if this is what you mean?

server/routes/establishments/benchmarks/index.js
```diff
@@ -70,28 +70,25 @@ const pay = async (establishmentId) => {
...
+  if (!establishment || establishment.NumberOfStaffValue === 0 ||
+    workerCount !== establishment.NumberOfStaffValue) {
+    return {percentOfPermTemp: 0, stateMessage: 'no-workers'};
+  }
+  const permTemptCount = await models.worker.permAndTempCountForEstablishment(establishmentId);
+  const leavers = await models.establishmentJobs.leaversForEstablishment(establishmentId);
+  if (establishment.LeaversValue !== 'With Jobs') {
+    return {percentOfPermTemp: 0, stateMessage: 'no-data'};
+  }
+  if ((leavers / permTemptCount) > 9.95) {
```
chadtomkiss-madetech, line 85: Duplication of `(leavers / permTemptCount)` for both the `if` and the `percentOfPermTemp` calculation - could use a variable.

server/routes/establishments/benchmarks/index.js
```diff
@@ -70,28 +70,25 @@ const pay = async (establishmentId) => {
...
-    stateMessage = 'no-workers';
+  if (!establishment || establishment.NumberOfStaffValue === 0 ||
+    workerCount !== establishment.NumberOfStaffValue) {
+    return {percentOfPermTemp: 0, stateMessage: 'no-workers'};
+  }
+  const permTemptCount = await models.worker.permAndTempCountForEstablishment(establishmentId);
+  const leavers = await models.establishmentJobs.leaversForEstablishment(establishmentId);
+  if (establishment.LeaversValue !== 'With Jobs') {
+    return {percentOfPermTemp: 0, stateMessage: 'no-data'};
+  }
```
chadtomkiss-madetech, line 84: This `if` is right between the consts and where the consts are used - if you move this `if` above the consts you'll get all of your early returns together and then the main flow is the calculation 👍

PR #2413, opened 2020-08-14
Feat/back to admin

claresudbery:
#### Tests
- [x ] Yes


claresudbery:
> Can we use the enum? I saw somewhere else role === Roles.Admin

Yup, good spot. Done that.

src/app/core/components/header/header.component.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  it('should render a HeaderComponent', async () => {
+    const { component } = await setup();
+    expect(component).toBeTruthy();
+  });
+
+  describe('Back to admin link', () => {
+    it('should display a Back to admin link if user is an admin and is logged in', async () => {
+      const { component } = await setup(true, 0, true);
```
chadtomkiss-madetech, line 56: I find knowing the meaning of `true, 0, true` to be difficult.

You could do something like:


```
const isAdmin = true;
const subsidiaries = 0;
const isLoggedIn = true;

const { component } = await setup(isAdmin, subsidiaries, isLoggedIn);
```

or use destructuring:

```
const { isAdmin, subsidiaries, isLoggedIn } = [true, 0, true];

const { component } = await setup(isAdmin, subsidiaries, isLoggedIn);
```

PR #2440, opened 2020-08-21
Updated turnover to accept a higher precision

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/migrations/20200821074704-updateBenchmarks.js
+++ b/migrations/20200821074704-updateBenchmarks.js
@@ -1,0 +1,26 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.changeColumn(
+      {
+        tableName: 'Benchmarks',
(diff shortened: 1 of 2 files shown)


migrations/20200821074704-updateBenchmarks.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.changeColumn(
+      {
+        tableName: 'Benchmarks',
+        schema: 'cqc'
+      },
+      'Turnover', {
+        type: Sequelize.DataTypes.DECIMAL(5, 2),
```
chadtomkiss-madetech, line 11: @aaron-russell Do we need to update the import script in the SFC-DB repo that creates a new table and renames it?

How do you see managing the changes across both migrations and the import script?

g-m-taylor (approved):

PR #2469, opened 2020-09-03
Benchmarks/getaddress.io

g-m-taylor:
#### Tests
- [] Yes
- [x] No, I found it difficult to test    - can't set API key in circleCI


server/test/unit/utils/getAddressAPI.spec.js
```diff
@@ -0,0 +1,11 @@
+const expect = require('chai').expect;
+
+const getAddressAPI = require('../../../utils/getAddressAPI');
+
+describe('fileLock util', () => {
+  it("should return the correct data from API", async () => {
+    const data = await getAddressAPI.getAddressAPI.getPostcodeData("M46FW");
+    expect(data.addresses[0].district).to.equal("Manchester");
+  });
```
aaron-russell, line 9: Is this going to work on Circle without an API key?

server/utils/getAddressAPI.js
```diff
@@ -0,0 +1,17 @@
...
+        const postcodeData = await axios.get(`https://api.getAddress.io/find/${postcode.replace(/ /g, '')}?api-key=${config.get('getAddress.apikey')}&expand=true`);
+        return postcodeData.data;
+      } catch (err) {
+        console.error(err);
+        return false;
+      }
+    }
+  };
+module.exports.getAddressAPI = getAddressAPI;
+module.exports.getAddressAPI.getPostcodeData = getAddressAPI.getPostcodeData;
```
aaron-russell, line 17: Do you need to export the child if you're exporting the parent too?

aaron-russell (changes requested):

server/test/unit/utils/getAddressAPI.spec.js
g-m-taylor: Thats what I was testing :P Turns out: nope

server/utils/getAddressAPI.js
```diff
@@ -0,0 +1,16 @@
+var config = require('../../server/config/config');
```
aaron-russell, line 1: Why is this a var?

aaron-russell (changes requested):

aaron-russell (approved):

PR #2492, opened 2020-09-16
Stopped SLV validation on BU

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server/models/BulkImport/csv/establishments.js
+++ b/server/models/BulkImport/csv/establishments.js
@@ -2210,7 +2210,7 @@ class Establishment {
 
       this._validateReasonsForLeaving();
 
-      this._validateNoChange();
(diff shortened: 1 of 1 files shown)


g-m-taylor (approved):

server/models/BulkImport/csv/establishments.js
```diff
@@ -2210,7 +2210,7 @@ class Establishment {
 
       this._validateReasonsForLeaving();
 
-      this._validateNoChange();
+      // this._validateNoChange();
```
g-m-taylor, line 2213: Might be good to add a reason? aka : not currently working - disabling til fix can be found?

aaron-russell: Sure

PR #2497, opened 2020-09-18
WIP: Apply prettier and ng lint code fixes

chadtomkiss-madetech:
#### Tests
- [ ] Yes


aaron-russell:
If we are going to do thes, we many want to rerun it @chadtomkiss-madetech ?

PR #2507, opened 2020-09-23
Changing where the guard is for benchmarks

g-m-taylor:
#### Tests
- [] Yes


src/app/app-routing.module.ts
```diff
@@ -98,7 +98,7 @@ const routes: Routes = [
       {
         path: 'benchmarks',
         loadChildren: () => import('./shared/components/benchmarks-tab/benchmarks.module').then(m => m.BenchmarksModule),
-        canActivate: [CheckPermissionsGuard],
+        canActivate: [],
         data: {
           permissions: ['canViewBenchmarks'],
```
aaron-russell, line 103: This can be removed

aaron-russell (changes requested):

aaron-russell (approved):

PR #2516, opened 2020-09-28
Flu jab form text link bug

madeviv:
#### Tests
- [ ] Yes


chadtomkiss-madetech:
@madeviv Do you think there's value in having a test to ensure the link is correct depending on whether you are viewing a primary workplace or not?

g-m-taylor (approved):

PR #2520, opened 2020-09-30
Get new locking into Preprod

aaron-russell:
#### Tests
- [ ] Yes


aaron-russell:
Think we may get a lot of this because of the linting changes

PR #2545, opened 2020-10-07
Added new S3 locking to bulk upload

aaron-russell:
Reverts NMDSdevopsServiceAdm/SopraSteria-SFC#2544


package.json
```diff
@@ -119,6 +119,7 @@
     "aws-sdk-mock": "^4.5.0",
     "chai": "^4.2.0",
     "codelyzer": "^5.1.2",
+    "delay": "^4.4.0",
```
chadtomkiss-madetech, line 122: Does `delay` get used?

chadtomkiss-madetech (changes requested):

package.json
aaron-russell: No, but it's useful to have as a dev dependency in the future

chadtomkiss-madetech: @aaron-russell given we removed a tonne of packages that were not used, I think I would add it as a dependency at the time there is a change that requires it?

chadtomkiss-madetech (approved):

PR #2559, opened 2020-10-16
Fixed 122 LGTM Issues

aaron-russell:
#### Tests
- [ ] Yes


server/aws/kinesis.js
```diff
@@ -7,13 +7,13 @@ let kinesis = null;
...
     apiVersion: '2013-12-02',
-    region
+    region,
   });
 };
 
 const ACTIONS = ['created', 'updated', 'deleted'];
 
-const establishmentPump = async (action, establishment)  => {
+const establishmentPump = async (action, establishment) => {
```
chadtomkiss-madetech, line 16: Did the Kinesis stuff get removed?

server/routes/login.js
```diff
@@ -19,7 +17,7 @@ const sendMail = require('../utils/email/notify-email').sendPasswordReset;
 
 const tribalHashCompare = (password, salt, expectedHash) => {
```
chadtomkiss-madetech, line 18:
```
// if this found login account is a migrated tribal account, and there is no current hash, then
//  we need to first validate password using tribal hashing
```

There is a process to migrate hashes when logging in.. do we know if there are any logins that have yet to be migrated? If not, we can probably remove this code completely.

server/aws/kinesis.js
aaron-russell: I think so actually, I can remove this file, not a problem

server/routes/login.js
aaron-russell: I'm not sure about that, I would guess so, think we will know once we start deleting old accounts

server/sendEmail.js
```diff
@@ -1,93 +1,94 @@
 // this is simply to test sending emails via AWA SES
```
chadtomkiss-madetech, line 1: Can this file be deleted? It's tied to Warrens email etc.

server/routes/login.js
aaron-russell: Think we're best waiting for this one

chadtomkiss-madetech (approved):

PR #2579, opened 2020-10-28
Feat: Upgrade GovUK Front-end

chadtomkiss-madetech:
**Work done**

- Upgraded `govuk-frontend` to `v3.9.1`

**Changes introduced**

- Update `grey-1/2/3` colours to `light/mid/dark-grey`
- Update SVG icon colours
- Update Tabs markup
- Fix Start now buttons to include the Arrow SVG
- Replaced custom colours with GovUK specific colours (some of these were the same)

#### Tests
- [ ] Yes


src/assets/scss/components/_tables.scss
```diff
@@ -21,7 +21,7 @@ $error_row_padding: $govuk-gutter-half;
 }
 
 .govuk-error-table__row {
-  border: $error_border_size solid $shiraz;
+  border: $error_border_size solid $govuk-error-colour;
```
aaron-russell, line 24: Why are these different compared to `govuk-colour('whatever')`?

chadtomkiss-madetech: The guidelines say:

> Only use the variables in the context they’re designed for. For example, if you wanted to use red to represent some data in a graph you should use govuk-colour("red") rather than $govuk-error-colour

Given it's for a `govuk-error` class it felt appropriate.

There's probably other cases which could use Sass variables instead of the `govuk-colour` function.

aaron-russell (approved):

PR #2580, opened 2020-10-28
Metric page

madeviv:
#### Tests
- [ ] Yes


server/routes/establishments/benchmarks/index.js
```diff
@@ -35,144 +57,141 @@ const getMetaData = async (reply,benchmarkComparisonGroup) => {
...
-      reply.tiles[key].comparisonGroup.hasValue =  benchmarkComparisonGroup[key] !== null  ;
-      if (!reply.tiles[key].comparisonGroup.hasValue) {
-        reply.tiles[key].comparisonGroup.stateMessage = 'no-data';
-      }
+
+const comparisonGroupData = (reply, benchmarkComparisonGroup) => {
+  Object.keys(reply.tiles).map((key) => {
+    if (benchmarkComparisonGroup) {
+      Object.keys(comparisonJson).map((comparison) => {
+        const item = comparison === 'comparisonGroup' ? key : key + comparison;
```
chadtomkiss-madetech, line 66: Does casing matter here? 

Before it was `${key}GoodCQC` but now it would be `${key}goodCqc`

server/routes/establishments/benchmarks/index.js
```diff
@@ -35,144 +57,141 @@ const getMetaData = async (reply,benchmarkComparisonGroup) => {
...
+        const value = benchmarkComparisonGroup[item];
+        reply.tiles[key][comparison].value = value ? parseInt(value) : 0;
+        reply.tiles[key][comparison].hasValue = benchmarkComparisonGroup[item] !== null;
```
chadtomkiss-madetech, line 69: `benchmarkComparisonGroup[item]` can be `value`

server/routes/establishments/benchmarks/index.js
aaron-russell: I changed the model to match, it makes it match the JSON which I think works better

chadtomkiss-madetech: Ahh good spot!

Are we ok with the casing on the model being `sicknessgoodCqc` and the DB column being `SicknessGoodCQC`?

There's 3 differences with the casing which is a bit confusing I think?

Also I don't think `sicknessgoodCqc` matches any casing convention?

chadtomkiss-madetech (approved):

PR #2659, opened 2020-11-10
Added read about comparison group

madeviv:
#### Tests
- [ ] Yes

### DIFF
--- a/server/routes/establishments/benchmarks/index.js
+++ b/server/routes/establishments/benchmarks/index.js
@@ -4,21 +4,15 @@ const models = require('../../../models');
 const clonedeep = require('lodash.clonedeep');
 
 const comparisonJson = {
-  comparisonGroup: {
-    value: 0,
-    hasValue: false,
-    stateMessage: 'no-data',
-  },
-  goodCqc: {
-    value: 0,
-    hasValue: false,
-    stateMessage: 'no-data',
-  },
-  lowTurnover: {
-    value: 0,
-    hasValue: false,
-    stateMessage: 'no-data',
-  },
+  value: 0,
+  hasValue: false,
+  stateMessage: 'no-data',
+};
+
+const comparisonGroupsJson = {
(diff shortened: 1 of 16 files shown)


server/routes/establishments/benchmarks/index.js
```diff
@@ -89,7 +87,7 @@ const pay = async (establishmentId) => {
       value: averagePaidAmount,
       hasValue: stateMessage.length === 0,
     },
-    ...clonedeep(comparisonJson),
+    ...buildComparisonGroupMetrics('pay', benchmarkComparisonGroup),
```
chadtomkiss-madetech, line 90: Do you think `'pay', 'turnover'` etc should be an enum rather than hardcoded strings?

src/app/core/utils/benchmarks-util.ts
```diff
@@ -0,0 +1,21 @@
+import { BenchmarkValue, Tile } from '@core/model/benchmarks.model';
+
+export class BenchmarksUtil {
+  static hasValue(tile: Tile, benchmarkValueCallback: (_: Tile) => BenchmarkValue): boolean {
+    if (tile) {
```
chadtomkiss-madetech, line 5: You could return early so that the main body of the function is what happens when you have the required params set.

i.e

```
if (!tile) {
  return false;
}

return benchmarkValueCallback(tile)?.hasValue;
```

chadtomkiss-madetech (approved):

server/routes/establishments/benchmarks/index.js
madeviv: If they're needed outside of this instance that is always a possibility

src/app/core/utils/benchmarks-util.ts
madeviv: I can do. It's the same number of conditionals and return statements but it does read better.

PR #2684, opened 2020-11-17
Feat/user ranking

madeviv:
#### Tests
- [ ] Yes


server/models/worker.js
```diff
@@ -1194,21 +1194,10 @@ module.exports = function (sequelize, DataTypes) {
...
-          [sequelize.Op.not]: null,
-        },
-      },
-    });
-  };
-  Worker.careworkersTotalHourlyPaySum = async function (establishmentId) {
-    return this.sum('AnnualHourlyPayRate', {
+
+  Worker.averageHourlyPay = async function (establishmentId) {
+    return this.findOne({
```
aaron-russell, line 1199: Just wanted to check this definitely gets the average and not just one value?

server/routes/establishments/benchmarks/index.js
```diff
@@ -73,18 +75,19 @@ const buildMetric = (metricValue) => {
...
-  const careworkersWithHourlyPayCount = await models.worker.careworkersWithHourlyPayCount(establishmentId);
-  let averagePaidAmount = 0;
+  let value = 0;
   let stateMessage = '';
-  if (careworkersWithHourlyPayCount > 0) {
-    let paidAmount = await models.worker.careworkersTotalHourlyPaySum(establishmentId);
-    averagePaidAmount = (paidAmount * 100) / careworkersWithHourlyPayCount;
+
+  const averageHourlyPay = await models.worker.averageHourlyPay(establishmentId);
+  if (averageHourlyPay.amount !== null) {
```
chadtomkiss-madetech, line 82: There is `averageHourlyPay.amount !== null` and `parseInt(averageHourlyPay.amount * 100)` in multiple places

So every time we call `averageHourlyPay` we need to null check and `* 100`.

Normally I would say move the calculation internally so that the function always returns the `* 100` value but the `null` check is also used to determine the `stateMessage`

Not sure what our options are :joy: but just wanted to flag as a potential refactor

server/routes/establishments/benchmarks/rankings/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+  return {
+    currentRank,
+    maxRank,
+    hasValue: true
+  }
+}
+
+router.route('/pay').get(async (req, res) => {
+  const establishmentId = req.establishmentId;
```
chadtomkiss-madetech, line 37: This is only used once, so it could be inlined?

```
const responseData = await pay(req.establishmentId);
```

chadtomkiss-madetech (approved):
This is fantastic work @madeviv <3

server/test/unit/routes/establishments/benchmarks/rankings.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    })
+
+    it('should be response with maxRank equal to number of comparison group rankings + current establishment', async () => {
+      const establishmentId = 123;
+
+      sinon.stub(models.worker,'averageHourlyPay').returns({ amount: 15.00 })
+      sinon.stub(models.benchmarksPay, 'getComparisonGroupRankings').returns([
+        { CssrID: 123, MainServiceFK: 1, pay: 1400, EstablishmentFK: 456 },
+        { CssrID: 123, MainServiceFK: 1, pay: 1600, EstablishmentFK: 789 }
+      ]);
```
aaron-russell, line 52: Shouldn't the current establishment be in the ranking too (maybe 2 different tests?)

src/app/shared/components/benchmark-metric/ranking-content/ranking-content.component.html
```diff
@@ -0,0 +1,12 @@
...
+</ng-container>
+<ng-template [ngSwitch]="state" #noRank>
+  <p *ngSwitchCase="'no-data'" class="govuk-!-font-size-27">
+    You've not added any data about hourly pay yet. We cannot show where you rank.
+  </p>
+  <p *ngSwitchCase="'no-comparison-data'" class="govuk-!-font-size-27">
+    We do not have enough data to show this comparison yet. Your workplace cannot be ranked without a comparison group.
+  </p>
+  <p *ngSwitchDefault class="govuk-!-font-size-27">&nbsp;</p>
+</ng-template>
```
aaron-russell, line 12: We should be able to use the same error as in the metric data if I remember rightly, just needs `We cannot show where you rank.` on the end

src/app/shared/components/benchmark-metric/ranking-content/ranking-content.component.html
```diff
@@ -0,0 +1,12 @@
+<ng-container *ngIf="showRank; else noRank">
+  <p class="govuk-!-font-size-27">Your workplace is ranked {{ rank }} in your comparison group.</p>
+</ng-container>
+<ng-template [ngSwitch]="state" #noRank>
+  <p *ngSwitchCase="'no-data'" class="govuk-!-font-size-27">
+    You've not added any data about hourly pay yet. We cannot show where you rank.
+  </p>
+  <p *ngSwitchCase="'no-comparison-data'" class="govuk-!-font-size-27">
```
aaron-russell, line 8: No point in having the scss file for this component if it's empty

src/app/shared/components/benchmarks-tab/metric/metric.component.spec.ts
```diff
@@ -108,4 +117,21 @@ describe('BenchmarksMetricComponent', () => {
...
+
+    fixture.whenStable()
+    //const lowestRank = getByText('3Lowest ranking');
+    //const highestRank = getByText('Highest ranking 1');
+    const currentRank = getByText('Highest ranking 1');
+
+    //expect(lowestRank).toBeTruthy();
+    //expect(highestRank).toBeTruthy();
+    expect(currentRank).toBeTruthy();
+  });*/
```
aaron-russell, line 136: If this is not needed, should be removed

aaron-russell (changes requested):

server/models/worker.js
madeviv: Yes it does. findAll would return an array.

server/routes/establishments/benchmarks/index.js
madeviv: averageHourlyPay.amount !== null only happens once

server/test/unit/routes/establishments/benchmarks/rankings.spec.js
madeviv: Check the real implementation of benchmarksPay to answer your question

src/app/shared/components/benchmark-metric/ranking-content/ranking-content.component.html
madeviv: The only place 'We do not have enough data to show this comparison yet' is shared is in the `barchart-options-builder` and `benchmark-tile.component`

src/app/shared/components/benchmarks-tab/metric/metric.component.spec.ts
madeviv: It is I need to fix it. Can't figure out why it isn't working

aaron-russell (approved):

PR #2686, opened 2020-11-18
Turnover ranking

madeviv:
#### Tests
- [X] Yes
- [] No, they are not required for this change

### DIFF
--- a/server/models/benchmarksTurnover.js
+++ b/server/models/benchmarksTurnover.js
@@ -1,5 +1,5 @@
 module.exports = function (sequelize, DataTypes) {
-  const BenchmarksPay = sequelize.define(
+  const BenchmarksTurnover = sequelize.define(
     'benchmarksTurnover',
     {
       CssrID: {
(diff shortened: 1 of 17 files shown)


server/models/benchmarksTurnover.js
```diff
@@ -0,0 +1,59 @@
+module.exports = function (sequelize, DataTypes) {
+  const BenchmarksPay = sequelize.define(
```
aaron-russell, line 2: This should be `BenchmarksTurnover`

src/app/shared/components/benchmark-metric/barchart/barchart-options-builder.ts
```diff
@@ -123,7 +123,7 @@ export class BarchartOptionsBuilder {
   private formatLabel(): Highcharts.AxisLabelsFormatterCallbackFunction {
     return function () {
       const bold = this.isFirst ? 'govuk-!-font-weight-bold' : 'govuk-!-font-weight-regular';
```
aaron-russell, line 125: You can get rid of this line

aaron-russell (changes requested):

server/models/benchmarksTurnover.js
madeviv: Good catch

aaron-russell (approved):

aaron-russell (approved):

PR #2726, opened 2020-11-30
Added to data layer when first time login (along with the survey only…

aaron-russell:
… showing once)

#### Tests
- [ ] Yes

### DIFF
--- a/src/app/features/dashboard/home-tab/home-tab.component.ts
+++ b/src/app/features/dashboard/home-tab/home-tab.component.ts
@@ -108,10 +108,10 @@ export class HomeTabComponent implements OnInit, OnDestroy {
       );
     }
 
-    if (!this.workplace?.employerType) {
+    if (!this?.workplace?.employerType) {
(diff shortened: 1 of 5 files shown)


g-m-taylor (approved):

src/app/features/dashboard/home-tab/home-tab.component.ts
```diff
@@ -88,6 +97,13 @@ export class HomeTabComponent implements OnInit, OnDestroy {
         }),
       );
     }
+
+    if (!this?.workplace?.employerType) {
```
chadtomkiss-madetech, line 101: At what point would `employerType` be populated so that `firstTimeLogin` is false?

aaron-russell: When they enter the workplace information, it's the first question, it's also the same rule as we show them on the home page tab 'What to do first'

chadtomkiss-madetech (approved):

PR #2736, opened 2020-12-03
Feat: Benchmark PDFs for Sickness, Quals, Pay & Turnover

chadtomkiss-madetech:
**Work done**

- Updated `BuildBenchmarksPdf` to work with individual benchmarks
- Refactored creation of PDF into smaller funcs (for adding header/footer/about the data etc)
- Added `canvg` to convert SVGS into Canvas elements so that the charts can be rendered in the PDF
- Remove `html2canvas` as a top level dependency (is pulled in via `jspdf`)

#### Tests
- [ ] Yes


src/app/core/services/pdf.service.ts
```diff
@@ -3,76 +3,183 @@ import { Establishment } from '@core/model/establishment.model';
...
+      switch (page) {
+        case 1:
+          return 540;
+        case 2:
+          return 380;
+      }
+    },
+  } as ReportInterface,
+};
+
```
chadtomkiss-madetech, line 44: Not sure how I feel about this - it's used to calculate where the footer at the bottom should go on each page

src/app/core/services/pdf.service.ts
```diff
@@ -3,76 +3,183 @@ import { Establishment } from '@core/model/establishment.model';
...
 import { PdfHeaderComponent } from '@features/pdf/header/pdf-header.component';
 import { PdfWorkplaceTitleComponent } from '@features/pdf/workplace-title/pdf-workplace-title.component';
+import Canvg from 'canvg';
 import { jsPDF } from 'jspdf';
 
 export interface PdfComponent {
   content: ElementRef;
 }
 
+export enum ReportType {
```
chadtomkiss-madetech, line 13: Should the terminology be `Report`? We already have Training Reports etc and this is for the Benchmarks PDF

src/app/core/services/pdf.service.ts
```diff
@@ -3,76 +3,183 @@ import { Establishment } from '@core/model/establishment.model';
...
   providedIn: 'root',
 })
 export class PdfService {
+  public doc = new jsPDF('p', 'pt', 'a4');
+  public ptToPx = 1.3333333333;
+  public scale = 0.5;
+  public width = 1000;
+  public spacing = 50;
+  public y = 20;
+  public ypx = (this.y * this.ptToPx) / this.scale;
```
chadtomkiss-madetech, line 55: I moved these to be class properties in order to reference them in other functions but there might be a better way?

src/app/core/services/pdf.service.ts
madeviv: Would it not be easier to get ReportType and ReportInterface and just have a separate BuildMetricsPdf public function on PdfService

src/app/core/services/pdf.service.ts
madeviv: Would it not be easier to get rid of ReportType and ReportInterface and just have a separate BuildMetricsPdf public function on PdfService

src/app/shared/components/benchmarks-tab/benchmarks-tab.component.ts
```diff
@@ -66,7 +66,13 @@ export class BenchmarksTabComponent implements OnInit, OnDestroy {
...
     $event.preventDefault();
     try {
-      return await this.pdfService.BuildBenchmarksPdf(this.elRef, this.aboutData.aboutData, this.workplace);
+      await this.pdfService.BuildBenchmarksPdf(
+        this.elRef,
+        this.aboutData.aboutData,
+        this.workplace,
+        ReportType.dashboard,
+        'Benchmarks.pdf',
+      );
```
madeviv, line 75: I would probably do away with having a catch all BuildBenchmarksPdf function and create a separate function for metrics then you don't need to increase the number of arguments

src/assets/scss/modules/_utils.scss
```diff
@@ -1,3 +1,7 @@
+.util__hidden {
+  display: none !important;
+}
+
```
aaron-russell, line 4: Missed a trick not calling it .ninja! Should we be using `!important`?

aaron-russell (changes requested):

src/assets/scss/modules/_utils.scss
chadtomkiss-madetech: Funny enough there is already:

```
.hidden {
  display: none !important;
}
```

in `benchmarks-tab.component.scss`

I don't think the `!important` is an issue - I think utility classes are a good option for using it? I also don't think we would ever want to override the display of an element that has a `hidden` class on it?

Suggestion: move `.hidden` to `_utils.scss` and replace `util__hidden` classes with `hidden`?

aaron-russell: Sounds good

chadtomkiss-madetech: @aaron-russell it was an old class that wasn't used - we don't have an About the data link on the individual tile pages so don't need to hide the component

aaron-russell (approved):

PR #2783, opened 2020-12-14
Adding Who should do bulk upload

g-m-taylor:
#### Tests
- [ ] Yes


madeviv (changes requested):
Same thing with the other pr. Folder structure has changed.

Also you've removed all the components from the page. How will we test it?

g-m-taylor:
> Same thing with the other pr. Folder structure has changed.
> 
> Also you've removed all the components from the page. How will we test it?

Test what? The components were arn't using anymore?

madeviv:
You've removed the components that allow you to do the uploading

g-m-taylor:
> You've removed the components that allow you to do the uploading. How can we test changes with no BU components?
Am I missing something? 
Dont we have a ticket to redesign that component? Aka the outcome of the drop and drop spike.  Why would we test it before we actually do the work for that?  Also the component is still on the original bulk upload page.

madeviv:
If you take a look at that component on my PR it might make a bit more sense. We're developing the new functionality in parallel with the existing functionality so there will be some duplication as we iterate. The new page should get it's own components as the needs arise to diverge from the existing functionality

aaron-russell (dismissed):

g-m-taylor:
> If you take a look at that component on my PR it might make a bit more sense. We're developing the new functionality in parallel with the existing functionality so there will be some duplication as we iterate. The new page should get it's own components as the needs arise to diverge from the existing functionality

surely its better to add those components in when the need arises rather than mix old and new components mish-mashed together...

madeviv (approved):

PR #2786, opened 2020-12-15
Changed permissions and added some permissions middleware

aaron-russell:
#### Tests
- [ ] Yes


server/models/cache/singletons/permissions.js
```diff
@@ -276,158 +280,224 @@ let ALL_PERMISSIONS = [
...
+      throw new Error(error);
+    }
+    const becomeAParentRequest = await models.Approvals.findOne({
+      where: {
+        EstablishmentID: req.establishmentId,
+        Status: 'Pending',
+        ApprovalType: 'BecomeAParent',
+      },
+      order: [['createdAt', 'DESC']],
+    });
```
madeviv, line 317: Could potentially move the declaration of `thisEstablishment` and `becomeAParentRequest` closer to where they are used.

madeviv (approved):

server/models/cache/singletons/permissions.js
chadtomkiss-madetech: We could potentially make a method on the model to find this as well:

```
models.Approvals.findLatestParentRequest(req.establishmentId);
```

g-m-taylor (approved):

PR #2787, opened 2020-12-16
Feat/delete report

g-m-taylor:
#### Tests
- [x] Yes


server/routes/reports/deleteReport/report.js
```diff
@@ -84,7 +78,16 @@ const addCSSRData = async (establishmentsData) => {
...
-
+const getLastUpdate = (establishment) => {
+  if (establishment.workers.length > 0) {
+    if (moment(establishment.updated).isAfter(moment(establishment.workers[0].updated))) {
+      return establishment.updated;
+    } else {
+      return establishment.workers[0].updated;
+    }
+  }
+  return establishment.updated;
```
chadtomkiss-madetech, line 89: If we return early and switch the `isAfter` to the worker then we can avoid nesting / elses to improve readability:

```
if (establishment.workers.length === 0) {
  return establishment.updated;
}

if (moment(establishment.workers[0].updated).isAfter(establishment.updated))) {
  return establishment.workers[0].updated;
}

return establishment.updated;
```

madeviv (approved):

server/routes/reports/deleteReport/report.js
```diff
@@ -114,15 +117,15 @@ const fillData = (reportData, laData, WS1) => {
...
+        establishment.nmdsId.trim(),
         establishment.id,
         region,
         la,
         establishment.mainService.name,
         establishment.EmployerTypeValue,
         excelUtils.formatBool(establishment.isRegulated),
         parentName,
-        moment(establishment.updated).add(monthsToBeDelete, 'months').format('DD-MM-YYYY'),
+        new Date(moment(getLastUpdate(establishment)).add(monthsToBeDelete, 'months').format('MM-DD-YYYY')),
```
chadtomkiss-madetech, line 128: Does this work? It's passing a formatted string to `new Date` which then would no longer have the formatting?

chadtomkiss-madetech (approved):

server/routes/reports/deleteReport/report.js
g-m-taylor: yeah exceljs seems to expect a Date to be able to format the column as a date

PR #2811, opened 2020-12-22
Bug/cant see add user: Fixing bug where User can't see Add a user button

g-m-taylor:
If a workplace user doesn't have a name (this can only happen if the user was migrated) then the Add a user Button disappears.  Added a check to avoid this in the future.

#### Tests
- [x ] Yes


g-m-taylor:
[[## 25217 ##] Workplaces not having the add user option](https://trello.com/c/SbPCereZ/347-25217-workplaces-not-having-the-add-user-option)

aaron-russell (approved):

PR #2817, opened 2020-12-23
Added permission check to user endpoints

aaron-russell:
#### Tests
- [ ] Yes


server/routes/establishments/la.js
```diff
@@ -316,4 +317,10 @@ const formatLAResponse = (establishment, primaryAuthority = null) => {
   return response;
 };
 
+router.route('/').get(hasPermission('canViewEstablishment'), getLAs);
+router.route('/').post(hasPermission('canEditEstablishment'), updateLAs);
+// These don't look to be used so could be removed
```
g-m-taylor, line 322: Do we need this comment?

g-m-taylor (approved):

server/routes/establishments/la.js
aaron-russell: Yeah, I think so, as I don't want to get rid of them as part of this, but it's something that needs to be flagged I think

PR #2839, opened 2021-01-04
Bu/last bulk uploaded - Adding in the last bulk uploaded date

g-m-taylor:
#### Tests
- [x] Yes


src/app/features/bulk-upload-v2/bulk-upload-page/bulk-upload-page.component.ts
```diff
@@ -35,7 +36,16 @@ export class BulkUploadPageV2Component implements OnInit, OnDestroy {
   ngOnInit() {
     this.breadcrumbService.show(JourneyType.BULK_UPLOAD);
     this.establishment = this.establishmentService.primaryWorkplace;
-    this.setupFormErrorsMap();
```
madeviv, line 38: @g-m-taylor was this supposed to be removed?

src/app/features/bulk-upload-v2/bulk-upload-page/bulk-upload-page.component.ts
```diff
@@ -35,7 +36,16 @@ export class BulkUploadPageV2Component implements OnInit, OnDestroy {
   ngOnInit() {
     this.breadcrumbService.show(JourneyType.BULK_UPLOAD);
     this.establishment = this.establishmentService.primaryWorkplace;
```
madeviv, line 38: Is it possible to get the last bulk uploaded date from the here rather than creating a new endpoint?

src/app/features/bulk-upload-v2/bulk-upload-page/bulk-upload-page.component.ts
g-m-taylor: nope :/ added it back in

src/app/features/bulk-upload-v2/bulk-upload-page/bulk-upload-page.component.ts
g-m-taylor: I did think about that, but that would have meant adding it to the object as currently establishment doesnt store that. So seemed cleaner to add an endpoint rather than fetch it back every time we get an establishment (as we dont use it anywhere else)

madeviv: But we would only get it as many times as we already get the establishment right? So it wouldn't make any difference to the number of times we call the backend just that that piece of data would be included in the request. And it's data from the same table

aaron-russell: I don't think this is a good enough reason to not approve. If it's not needed anywhere else, it doesn't add much to have an endpoint. And as this is already built. I would go ahead with it.

aaron-russell (approved):

src/app/features/bulk-upload-v2/bulk-upload-page/bulk-upload-page.component.ts
madeviv: I'm looking at it more from code maintenance it's more code to have new endpoints and models. Also the `establishmentService.primaryWorkplace` is called once and cached locally until the data changes IIRC

g-m-taylor: Might be my misunderstanding of how establishment is received in establishmentService. I think it uses the .restore method. which would mean having to edit model, the restore function, add a property to the property manager, and creating a the property file? Or am I looking at the wrong function?

madeviv: You only need to add a property and update the property manager stuff for extended properties e.g. db columns with `SavedAt` or `CreatedBy`. In this cause you would update the restore method to set `lastBulkUploaded` on the class and then update the `toJSON` function to include that it the response. Then it should just be available client side and IIRC it is cached client side.

g-m-taylor: ohhhh I getcha!

aaron-russell (approved):

PR #2845, opened 2021-01-07
Error Report Page

aaron-russell:
#### Tests
- [X] Yes


server/routes/establishments/bulkUpload/errorReport.js
```diff
@@ -0,0 +1,156 @@
...
+    const errorReport = await getErrorReport(req.establishmentId);
+
+    await s3.saveResponse(req, res, 200, errorReport);
+  } catch (error) {
+    await s3.saveResponse(req, res, 404);
+    throw new Error(error);
+  }
+};
+
+const createTableHeader = (currentWorksheet) => {
```
chadtomkiss-madetech, line 67: On a second glance at this file, I am thinking there's a bunch of CSV generation code in a file that's responsible for routing.

Think we can abstract this out

server/utils/errorWarningArray.js
```diff
@@ -0,0 +1,39 @@
...
+  const filterErrors = (msg) => msg.errCode && msg.errType;
+  const filterWarnings = (msg) => msg.warnCode && msg.warnType;
+
+  const result = [];
+
+  report
+    .filter((msg) => (type === 'error' ? filterErrors(msg) : filterWarnings(msg)))
+    .sort((a, b) => a.lineNumber - b.lineNumber)
+    .map((item) => {
+      const { lineNumber, name, source, worker, ...uniqueInfo } = item;
```
chadtomkiss-madetech, line 13: `...uniqueInfo` is actually the common info for an item - think we should rename this

server/utils/errorWarningArray.js
```diff
@@ -0,0 +1,39 @@
...
+        ...(worker && { worker }),
+      };
+
+      const existingCode =
+        type === 'error'
+          ? result.find((res) => res.errCode === item.errCode)
+          : result.find((res) => res.warnCode === item.warnCode);
+
+      if (existingCode) {
+        existingCode.items.push(errWarn);
```
chadtomkiss-madetech, line 27: Not sure if I'm reading this wrong, but should something happen with `existingCode`? It doesn't seem to be used.

If this isn't a bug, then I think we can simplify the `if/else` by negating the `if`

src/app/core/model/bulk-upload.model.ts
```diff
@@ -75,6 +75,34 @@ export interface ValidatedFile {
   username: string;
   deleted?: number;
 }
+export interface ErrorReportError {
```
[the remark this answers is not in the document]
chadtomkiss-madetech: Interfaces can be extended, so we could have an interface that holds the common variables and have interfaces that only include the different items

i.e

```
interface ErrorReport {
  origin
  items...
}

interface ErrorReportError extends ErrorReport {
  errCode...
}

chadtomkiss-madetech (approved):

PR #2846, opened 2021-01-07
Feat/remove flu banner 

g-m-taylor:
removing api and service for workplace flu

#### Tests
- [ ] Yes


aaron-russell (changes requested):
Do  we need to get rid of the endpoints too?

aaron-russell (approved):

PR #2882, opened 2021-01-19
Added isAdmin to datalayer

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/src/app/features/dashboard/home-tab/home-tab.component.ts
+++ b/src/app/features/dashboard/home-tab/home-tab.component.ts
@@ -98,6 +98,10 @@ export class HomeTabComponent implements OnInit, OnDestroy {
       );
     }
 
+    this.window.dataLayer.push({
+      isAdmin: this.user.role === 'Admin',
+    });
+
     if (!this?.workplace?.employerType) {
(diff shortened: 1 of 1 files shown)


src/app/features/dashboard/home-tab/home-tab.component.ts
```diff
@@ -98,6 +98,10 @@ export class HomeTabComponent implements OnInit, OnDestroy {
       );
     }
 
+    this.window.dataLayer.push({
+      isAdmin: this.user.role === 'Admin',
```
chadtomkiss-madetech, line 102: @aaron-russell Do you think this should be added to the dataLayer for all pages and not just the home tab?

Would there be a situation where someone goes directly to a page (maybe bookmarked etc) that would skip this component being rendered and miss out on the `isAdmin` flag being set?

aaron-russell: Potentially, however, it's only for the HotJar survey atm which works on the home tab. But thought this could be used in the future for other reasons.

chadtomkiss-madetech (approved):

PR #2910, opened 2021-01-26
Postcode import

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/CQCLocationAPICall/importPostcodeData.js
+++ b/CQCLocationAPICall/importPostcodeData.js
@@ -1,0 +1,49 @@
+const fs = require('fs');
+const path = require('path');
+const csv = require('fast-csv');
+const models = require('./models/index');
+
+try {
+  fs.readdir('./pcodefile', async (error, files) => {
+    if (error) {
+      console.error(error);
+    }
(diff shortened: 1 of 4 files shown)


CQCLocationAPICall/importPostcodeData.js
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs');
+const path = require('path');
+const csv = require('fast-csv');
+const models = require('./models/index');
+
+try {
+  fs.readdir('./pcodefile', async (error, files) => {
+    if (error) {
+      console.error(error);
```
chadtomkiss-madetech, line 9: Should this continue if there's an error? Or should we return here?

CQCLocationAPICall/importPostcodeData.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    console.log(files);
```
chadtomkiss-madetech, line 12: Do we need this `console.log` or was it for testing?

chadtomkiss-madetech (approved):

PR #2911, opened 2021-01-26
Feat/bu delete files

joannafawl:
#### Changes
- Deleting file on bulk upload page deletes file in S3

#### Tests
- [x] Yes


g-m-taylor (approved):

server/test/unit/routes/establishments/bulkUpload/s3.spec.js
```diff
@@ -0,0 +1,54 @@
+// const expect = require('chai').expect;
```
aaron-russell, line 1: Need to remove this line

PR #2915, opened 2021-01-27
Feat/multi file type

g-m-taylor:
#### Tests
- [x] Yes


server/routes/establishments/bulkUpload/uploadFiles.js
```diff
@@ -106,219 +103,136 @@ const uploadedPost = async (req, res) => {
...
+        file.header,
+      );
+      break;
+    case 'Worker':
+      passedCheck = new WorkerCsvValidator(file.importedData[firstRow], firstLineNumber).preValidate(file.header);
+      break;
+    case 'Training':
+      passedCheck = new TrainingCsvValidator(file.importedData[firstRow], firstLineNumber).preValidate(file.header);
+      break;
+  }
```
chadtomkiss-madetech, line 129: What do you think about a factory to return the type of validator based on the file type?

All of the arguments to the validators are the same so you could do something like

```
const validator = getValidator(file);
const passedCheck = validator.preValidate(file.header);
```

src/app/core/test-utils/MockBulkUploadService.ts
```diff
@@ -17,7 +17,24 @@ const ValidatedFileBuilder = build('ValidatedFile', {
   },
 });
 
-export const ValidatedFile = ValidatedFileBuilder();
+export const TrainingFile = ValidatedFileBuilder({
+  overrides: {
+    fileType: 'Training',
```
chadtomkiss-madetech, line 22: Love this 😍

src/app/features/bulk-upload-v2/drag-and-drop-files-list/drag-and-drop-files-list.component.ts
```diff
@@ -172,11 +174,26 @@ export class DragAndDropFilesListComponent implements OnInit, OnDestroy {
...
 
+  private checkForDuplicateTypes(): boolean {
+    const fileTypesArr = this.uploadedFiles.map(function (file) {
+      return file.fileType;
+    });
+    const isDuplicate = fileTypesArr.some(function (item, idx) {
+      return fileTypesArr.indexOf(item) != idx;
+    });
+
+    return isDuplicate;
```
chadtomkiss-madetech, line 194: When we assign a variable and return straight away we can just return the `fileTypesArr.some` part directly I think

src/app/features/bulk-upload-v2/drag-and-drop-files-list/drag-and-drop-files-list.component.html
```diff
@@ -29,7 +29,7 @@
           </p>
         </td>
         <td class="govuk-table__cell" [class.govuk-error-table__cell]="file.errors">
-          {{ file.fileType ? bulkUploadFileTypeEnum[file.fileType] : getFileType(file.filename) }}
+          {{ file.fileType ? file.fileType : '-' }}
```
chadtomkiss-madetech, line 32: Is `bulkUploadFileTypeEnum` and `getFileType` used anywhere else? Or does the fact we no longer use them mean we can remove them from the codebase?

chadtomkiss-madetech (approved):

PR #2936, opened 2021-02-01
Fix: capture timeout for Karma

chadtomkiss-madetech:
**Issue**

When running the tests, Chrome will start and then compile the assets before running the tests. However, the default timeout is 60s and the time to compile is ~90s.

There doesn't seem to be an elegant wait to wait for `ng build` before trying to run the tests.

**Work done**

- Have upped the `captureTimeout` to 120s

#### Tests
- [ ] Yes


chadtomkiss-madetech:
The build process is getting to 95% before killing Chrome.. before it was getting to 68%.. have upped the timeout to 140s

g-m-taylor (approved):

PR #2972, opened 2021-02-09
Feat: Keep Workplaces Active screen

chadtomkiss-madetech:
**Work done**

- Added `Emails` tab to admin navigation
- Initial template for sending emails to inactive workplaces
- Added dialog to confirm whether emails would be sent

**Note**

This is a static design of the Keep Workplaces Active screen and has no working functionality - merging into a feature branch rather than `test`

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/search/emails/dialogs/send-emails-confirmation-dialog/send-emails-confirmation-dialog.component.html
+++ b/src/app/features/search/emails/dialogs/send-emails-confirmation-dialog/send-emails-confirmation-dialog.component.html
@@ -1,10 +1,6 @@
-<h1 id="dialogHeading" class="govuk-panel--warning-prompt">You're about to permanently delete.</h1>
+<h1 id="dialogHeading" >Are you sure you want to send these emails?</h1>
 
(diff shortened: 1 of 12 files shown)


src/app/features/search/emails/dialogs/send-emails-confirmation-dialog/send-emails-confirmation-dialog.component.html
```diff
@@ -0,0 +1,6 @@
+<h1 id="dialogHeading" >Are you sure you want to send these emails?</h1>
+
+<p>This will send emails to <strong>5,723</strong> workplaces.</p>
```
aaron-russell, line 3: That's a lot of workplaces!

aaron-russell: Maybe leave a TODO saying this needs updating with the correct number

aaron-russell (approved):

PR #3026, opened 2021-02-24
Feat/bu admin skip button

duncanc19:
- Created component with admin Skip button for the missing reference pages
- Tested the conditional rendering of Skip button depending on whether admin 

#### Tests
- [X] Yes


src/app/features/bulk-upload-v2/bulk-upload-references/missing-refs-submit-exit-buttons/missing-refs-submit-exit-buttons.component.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor(public authService: AuthService) {}
+
+  @Input() public cta?: string;
+  @Input() public exit?: string;
+  @Input() public return?: URLStructure;
+  @Output() public skip: EventEmitter<any> = new EventEmitter();
+
+  onSkip(): void {
+    this.skip.emit(null);
+  }
```
aaron-russell, line 19: Could you extend the existing SubmitExitComponent?

src/app/features/bulk-upload-v2/bulk-upload-references/missing-refs-submit-exit-buttons/missing-refs-submit-exit-buttons.component.html
```diff
@@ -0,0 +1,33 @@
...
+    </button>
+  </div>
+  <ng-container *ngIf="this.authService.isAdmin">
+    <div class="govuk-grid-column-one-third govuk-!-padding-left-9">
+      <a (click)="onSkip()" class="govuk-button govuk-button--link">Skip</a>
+      <span class="govuk-visually-hidden">or</span>
+    </div>
+  </ng-container>
+  <div class="govuk-grid-column-one-third" [ngClass]="{ 'govuk-!-padding-left-9': !this.authService.isAdmin }">
+    <ng-container *ngIf="return">
```
chadtomkiss-madetech, line 17: Do you think the `*ngIf` should be above the `govuk-grid-column-one-third` div? I think we'll have a div with no content when `return` is false :eyes:

src/app/features/bulk-upload-v2/bulk-upload-references/missing-refs-submit-exit-buttons/missing-refs-submit-exit-buttons.component.ts
duncanc19: Do you mean extend as in a subclass? I'd wondered that but most of the duplication is in the HTML file, is it worth doing to avoid duplicating the three input lines? You could add the functionality to the original submit-exit-buttons component but you'd have to add an input to say whether it should have the skip and add AuthService as a dependency to it

aaron-russell: Then any changes that that would be added to this too

aaron-russell (approved):

PR #3028, opened 2021-02-24
Update getHistory endpoint

joannafawl:
#### Changes
- Update `getHistory` endpoint to get email campaign history from database
- Update test to mock database call

#### Tests
- [x] Yes

### DIFF
--- a/server/routes/admin/email-campaigns/inactive-workplaces/index.js
+++ b/server/routes/admin/email-campaigns/inactive-workplaces/index.js
@@ -26,7 +26,6 @@ const createCampaign = async (req, res) => {
 
     const inactiveWorkplaces = await findInactiveWorkplaces.findInactiveWorkplaces();
     const history = inactiveWorkplaces.map((workplace) => {
-
       return {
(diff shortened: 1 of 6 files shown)


server/routes/admin/email-campaigns/inactive-workplaces/index.js
```diff
@@ -51,16 +52,30 @@ const createCampaign = async (req, res) => {
...
-    {
-      date: '2021-01-05',
-      emails: 1356,
-    },
-    {
-      date: '2020-12-05',
-      emails: 278,
-    },
-  ];
+  const emailCampaigns = await models.EmailCampaign.findAll({
```
chadtomkiss-madetech, line 55: We can probably move this to be a method on the `EmailCampaign` model

```
await models.EmailCampaign.getHistory()
```

I also think we need a `where` clause to filter by `inactiveWorkplaces` campaign type

chadtomkiss-madetech (changes requested):

chadtomkiss-madetech (approved):

PR #3030, opened 2021-02-24
Feat/nototalstaffbanner: No Staff / More Staff and records banner

g-m-taylor:
#### Tests
- [x] Yes


src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -23,25 +23,8 @@ <h2 class="govuk-heading-m">Home</h2>
...
-          You said the total number of staff at your workplace was {{ workplace.numberOfStaff }}, but
-          {{ workersCount }} staff records have been added.
-        </p>
-        <p>
-          You either need to delete {{ workersCount - workplace.numberOfStaff }} staff records or change the total
-          number of staff for your workplace.
-        </p>
-      </app-inset-text>
+    <ng-container *ngIf="workplace && canViewListOfWorkers && workersCount !== workplace.numberOfStaff">
+    <app-staff-mismatch-banner (selectStaffTab)="selectStaffTab($event)" (selectTotalStaff)="selectTotalStaff($event)"[workersCount]="workersCount" [numberOfStaff] ="workplace.numberOfStaff"></app-staff-mismatch-banner>
```
chadtomkiss-madetech, line 27: Extra space

```suggestion
    <app-staff-mismatch-banner (selectStaffTab)="selectStaffTab($event)" (selectTotalStaff)="selectTotalStaff($event)"[workersCount]="workersCount" [numberOfStaff]="workplace.numberOfStaff"></app-staff-mismatch-banner>
```

src/app/shared/components/workplace-summary/workplace-summary.component.ts
```diff
@@ -152,4 +158,10 @@ export class WorkplaceSummaryComponent implements OnInit, OnDestroy {
   public isNumber(value: unknown): boolean {
     return typeof value === 'number';
   }
+
+  public shouldShowStaffBanner(): boolean {
+    return (
+      this.canViewListOfWorkers && this.isNumber(this.workerCount) && !this.wdfView && this.totalStaffWarningNonWDF
```
chadtomkiss-madetech, line 164: Do you think `!this.wdfView` should be part of the `totalStaffWarningNonWDF` method?

src/app/shared/components/workplace-summary/workplace-summary.component.html
```diff
@@ -1,9 +1,7 @@
 <dl
   *ngIf="workplace"
   class="govuk-summary-list"
-  [class.govuk-summary-list--contains-warnings]="
-    totalStaffWarning || (canViewListOfWorkers && !wdfView && workplace.numberOfStaff < workerCount)
-  "
+  [class.govuk-summary-list--contains-warnings]="totalStaffWarning || shouldShowStaffBanner()"
```
chadtomkiss-madetech, line 4: On second thoughts, I think `shouldShowStaffBanner` could be more descriptive as to what that actually means, especially as it's being used for more than just showing/hiding the banner

The hidden aria for the banner says:

`Total number of staff warning`

That seems to be very similar to what `totalStaffWarning` is designed to do?

not 100% clear on the difference but it looks to be based on:

- whether they can view a list of workers
- not in WDF view

Potential refactor here?

src/app/features/dashboard/home-tab/home-tab.component.ts
```diff
@@ -338,14 +338,22 @@ export class HomeTabComponent implements OnInit, OnDestroy {
...
+
   public selectStaffTab(event: Event) {
     if (event) {
       event.preventDefault();
     }
     this.workerService.tabChanged.next(true);
   }
 
+  public selectTotalStaff(event: Event) {
+    if (event) {
```
chadtomkiss-madetech, line 350: Is there any situation where we wouldn't pass an event? Don't think we need the `if`

If we were to get `preventDefault method does not exist on undefined` then the fix would be to pass the `$event` however this `if` would actually hide that bug

chadtomkiss-madetech (changes requested):

src/app/shared/components/workplace-summary/workplace-summary.component.ts
aaron-russell: I don't think so, it's just a different function as to when to show the banner

src/app/shared/components/workplace-summary/workplace-summary.component.html
aaron-russell: We mentioned this, however, the WDF designs change all of this anyway, so thought it best to leave this as it is for now.

I have renamed the function to be a little clearer. I agree on the similarities, and I mentioned this to Billy + Rich, but they said to leave it as this was all going to change for WDF.

The hidden label is the same just because it's the heading for that table item to screenreaders. The `<dt>` tag has to be there semantically, and should not be empty for screenreaders.

src/app/features/dashboard/home-tab/home-tab.component.ts
aaron-russell: Agreed, this was just copied from the home tab

chadtomkiss-madetech (approved):

PR #3053, opened 2021-03-02
Feat: Keep workplaces active - Calculate inactive workplaces

chadtomkiss-madetech:
**Work done**

- Added `LastUpdatedEstablishments ` view to store Workplace information and when they were last updated
- Implemented `findInactiveWorkplaces`
- Updated `sendEmails` function to use `switchMap` in order to get latest # of inactive workplaces after creating the campaign
- Fixed issue where API would respond before finishing inserting campaign history (this was a bug that meant the UI would not update when re-sending a request)

#### Tests
- [x] Yes

### DIFF
--- a/server/routes/admin/email-campaigns/inactive-workplaces/findInactiveWorkplaces.js
+++ b/server/routes/admin/email-campaigns/inactive-workplaces/findInactiveWorkplaces.js
@@ -3,7 +3,7 @@ const moment = require('moment');
 const models = require('../../../../models');
 
 const findInactiveWorkplaces = async () => {
-  const lastUpdated = moment().subtract(6, 'months').format('YYYY-MM-DD');
+  await models.sequelize.query('REFRESH MATERIALIZED VIEW cqc."LastUpdatedEstablishments"');
 
   const inactiveWorkplaces = await models.sequelize.query(
     `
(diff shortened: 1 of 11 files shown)


server/routes/admin/email-campaigns/inactive-workplaces/findInactiveWorkplaces.js
```diff
@@ -1,32 +1,74 @@
...
-  ];
-}
+    };
+  });
+};
 
 module.exports = {
-  findInactiveWorkplaces
+  findInactiveWorkplaces,
 };
```
aaron-russell, line 74: Could we use something like this: https://www.npmjs.com/package/sequelize-mv-support 

And could this be moved to the models folder? Don't think it makes sense where it is currently

aaron-russell (changes requested):

server/routes/admin/email-campaigns/inactive-workplaces/findInactiveWorkplaces.js
chadtomkiss-madetech: @aaron-russell I gave this a go and ran into a bunch of issues with existing model relationships - I think this is because it uses v5 of Sequelize and we are still on v4 :(

chadtomkiss-madetech: Have moved into the models folder though 👍

aaron-russell: Awesome, thank you :)

aaron-russell (approved):

PR #3057, opened 2021-03-03
Fix/bu admin skip completed banner

duncanc19:
#### Changes

- Pass in whether there are establishments with missing references left to nextMissingPage to stop _All references added_ banner from appearing after admin skipping
- Change save in missing-staff-references to remove the current establishment instead of the first one(removes bug where admin fills in a references page after having skipped and gets taken to the wrong references page)

#### Tests
- [ ] Yes

### DIFF
--- a/src/app/features/bulk-upload-v2/bulk-upload-references/missing-staff-references/missing-staff-references-page.component.ts
+++ b/src/app/features/bulk-upload-v2/bulk-upload-references/missing-staff-references/missing-staff-references-page.component.ts
@@ -92,7 +92,7 @@ export class MissingStaffReferencesComponent extends BulkUploadReferencesDirecti
     const currentEstablishmentIndex = this.establishmentsWithMissingReferences.findIndex(
       (establishment) => establishment.uid === this.establishmentUid,
     );
-    this.nextMissingPage(this.bulkUploadService.nextMissingReferencesNavigation(currentEstablishmentIndex + 1));
+    this.nextMissingPage(this.bulkUploadService.nextMissingReferencesNavigation(currentEstablishmentIndex + 1), true);
   }
 
   protected save(): void {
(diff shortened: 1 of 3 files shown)


aaron-russell (approved):

src/app/features/bulk-upload-v2/bulk-upload-references/missing-staff-references/missing-staff-references-page.component.ts
```diff
@@ -89,10 +93,10 @@ export class MissingStaffReferencesComponent extends BulkUploadReferencesDirecti
 
   public skipPage(): void {
     this.bulkUploadService.setMissingReferencesNavigation(this.establishmentsWithMissingReferences);
-    const currentEstablishmentIndex = this.establishmentsWithMissingReferences.findIndex(
-      (establishment) => establishment.uid === this.establishmentUid,
+    this.nextMissingPage(
+      this.bulkUploadService.nextMissingReferencesNavigation(this.currentEstablishmentIndex + 1),
+      true,
```
chadtomkiss-madetech, line 98: I think that it's a tad confusing to have a `nextMissingPage` function that can actually have a URL that's `bulk-upload` and display a success message if there's no references left.

We are then faking the fact there _is_ missing references when skipping in order to not show the success message

Not sure if there's something we can do to bring more clarity to what's happening?

duncanc19: Yeah, I was wondering about having the parameter as hasSkipped in nextMissingPage, because then it reads as if it's back on the bulk upload home page and it hasn't skipped, then the banner shows. But then you're passing that in as true every time there are references left and you haven't necessarily skipped.

duncanc19: Or you could add an API call to nextMissingPage which checks if missing references is 0 when it gets back onto the bulk upload page?

chadtomkiss-madetech: So all `nextMissingPage` does is redirect and then check if it's on the bulk upload page and adds a success message

What if `skipPage` just did:

```
const nextUrl = this.bulkUploadService.nextMissingReferencesNavigation(this.currentEstablishmentIndex + 1);

this.router.navigate(nextUrl);
```

duncanc19: But then could there be a case where an admin has skipped some pages but filled in others/the last one? Then they'd still get the banner

PR #3067, opened 2021-03-04
Fixed production build in preprod

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/src/app/app-routing.module.ts
+++ b/src/app/app-routing.module.ts
@@ -24,6 +24,32 @@ import {
 import { ResetPasswordComponent } from '@features/reset-password/reset-password.component';
 import { SatisfactionSurveyComponent } from '@features/satisfaction-survey/satisfaction-survey.component';
 
+import { environment } from '../environments/environment';
+
+const bulkUploadRouting = {
+  other: {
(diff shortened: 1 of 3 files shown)


src/app/app-routing.module.ts
```diff
@@ -16,34 +18,11 @@ import { DashboardComponent } from '@features/dashboard/dashboard.component';
...
-  test: {
-    path: 'bulk-upload',
-    loadChildren: () => import('@features/bulk-upload-v2/bulk-upload.module').then((m) => m.BulkUploadV2Module),
-    data: {
-      title: 'Bulk Upload',
-    },
-  },
-  production: {
-    path: 'bulk-upload',
-    loadChildren: () => import('@features/bulk-upload/bulk-upload.module').then((m) => m.BulkUploadModule),
```
chadtomkiss-madetech, line 41: For this deployment would it be safer to update the module here so that it can be reverted in live if there's issues but still be available in staging to fix them?

aaron-russell: This code doesn't work anyway, V2 is live in preprod

chadtomkiss-madetech (approved):

PR #3100, opened 2021-03-08
Feat: Refactor find inactive workplaces

chadtomkiss-madetech:
**Issue**

Codeclimate had some complaints about the `findInactiveWorkplaces` function.

**Work done**

- Split out `findInactiveWorkplaces` and `nextEmailTemplate` logic
- Moved non route files into `server/services/email-campaigns/inactive-workplaces`

#### Tests
- [x] Yes


server/models/email-campaigns/inactive-workplaces/nextEmail.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const shouldReceive = (inactiveWorkplace) => {
+  return getTemplate(inactiveWorkplace) !== null;
+}
+
+module.exports = {
+  getTemplate,
+  shouldReceive,
+}
```
aaron-russell, line 47: Should this be part of the model as it's more logic, I know we never spoke about where to put logic but think it should be separate.

aaron-russell (changes requested):

server/models/email-campaigns/inactive-workplaces/nextEmail.js
chadtomkiss-madetech: @aaron-russell I think maybe the `model` could just be the refresh view / run the query

and the logic to build the inactive workplace object / determine next template could live elsewhere?

so then `models` file is just the DB logic

I'm leaning towards services :eyes:

so

```
server/models/inactive-workplaces/getInactiveWorkplaces.js

server/services/email-campaigns/inactive-workplaces/inactiveWorkplaces.js
server/services/email-campaigns/inactive-workplaces/nextEmail.js
```

aaron-russell: Sounds good to me, yeah!

PR #3104, opened 2021-03-09
Made sure that the object not been there wont affect logins

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server/routes/login.js
+++ b/server/routes/login.js
@@ -15,7 +15,7 @@ const formatSuccessulLoginResponse = require('../utils/login/response');
 
 const sendMail = require('../utils/email/notify-email').sendPasswordReset;
 
-const { get } = require('lodash');
+const get = require('lodash/get');
 
 const tribalHashCompare = (password, salt, expectedHash) => {
   const hash = crypto.createHash('sha256');
--- a/server/routes/login.js
+++ b/server/routes/login.js
@@ -102,6 +102,15 @@ router.post('/', async (req, res) => {
           })
         : null;
 
+    let establishmentInfo = {
+      userId:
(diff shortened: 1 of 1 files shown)


server/routes/login.js
```diff
@@ -173,8 +173,14 @@ router.post('/', async (req, res) => {
...
-              userId: establishmentUser.user.uid,
-              establishmentId: establishmentUser.establishment.uid,
+              userId:
+                establishmentUser && establishmentUser.user && establishmentUser.user.uid
+                  ? establishmentUser.user.uid
+                  : null,
+              establishmentId:
+                establishmentUser && establishmentUser.establishment && establishmentUser.establishment.uid
+                  ? establishmentUser.establishment.uid
+                  : null,
```
chadtomkiss-madetech, line 183: Could we use lodash's get for safe navigation of an object?

```
userId: _.get(establishmentUser, 'user.uid'),
establishmentId: _.get(establishmentUser, 'establishment.uid')
```

will return `undefined` if it's not there

chadtomkiss-madetech (changes requested):

server/routes/login.js
aaron-russell: Sure

server/routes/login.js
```diff
@@ -15,6 +15,8 @@ const formatSuccessulLoginResponse = require('../utils/login/response');
 
 const sendMail = require('../utils/email/notify-email').sendPasswordReset;
 
+const { get } = require('lodash');
```
chadtomkiss-madetech, line 18: I _think_ you can do `require('lodash/get')`

so then you only pull in the get function rather than the whole of lodash

chadtomkiss-madetech (approved):

PR #3134, opened 2021-03-12
Added security headers to protect both back end and front end templates

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server.js
+++ b/server.js
@@ -143,31 +143,25 @@ var unless = function (root, path, middleware) {
 
 app.disable('x-powered-by');
 
-app.use(helmet.noSniff());
 app.use(
-  helmet.referrerPolicy({
-    policy: 'strict-origin-when-cross-origin',
-  }),
-);
-app.use(helmet.xssFilter());
-app.use(
-  helmet.frameguard({
-    action: 'deny',
-  }),
-);
-app.use(
-  helmet.permittedCrossDomainPolicies({
-    permittedPolicies: 'none',
-  }),
-);
-app.use(
-  helmet.expectCt({
-    maxAge: 86400,
-  }),
-);
-app.use(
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -141,16 +141,38 @@ var unless = function (root, path, middleware) {
...
+  helmet.permittedCrossDomainPolicies({
+    permittedPolicies: 'none',
+  }),
+);
+app.use(
+  helmet.expectCt({
+    maxAge: 86400,
+  }),
+);
+
```
chadtomkiss-madetech, line 168: I think we can configure this using a single helmet function rather than multiple `app.use`?

From the docs:

```
// This...
app.use(helmet());

// ...is equivalent to this:
app.use(helmet.contentSecurityPolicy());
app.use(helmet.dnsPrefetchControl());
app.use(helmet.expectCt());
app.use(helmet.frameguard());
app.use(helmet.hidePoweredBy());
app.use(helmet.hsts());
app.use(helmet.ieNoOpen());
app.use(helmet.noSniff());
app.use(helmet.permittedCrossDomainPolicies());
app.use(helmet.referrerPolicy());
app.use(helmet.xssFilter());
```

If you want to set custom props, you can do:

```
// This sets custom options for the `referrerPolicy` middleware.
app.use(
  helmet({
    referrerPolicy: { policy: "no-referrer" },
  })
);
```

and if you want to disable any , you can do:

```
// This disables the `contentSecurityPolicy` middleware but keeps the rest.
app.use(
  helmet({
    contentSecurityPolicy: false,
  })
);
```

aaron-russell: When I did this originally it said that they had depreciated the boolean, though it could have been the true value only. I'll take a look.

g-m-taylor (approved):

PR #3140, opened 2021-03-16
Feat: Keep Workplaces Active - parents db

joannafawl:
#### Changes
- Added a function to get all parent workplaces and their subsidiaries
- Updated `getInactiveWorkplaces` endpoint to include parent workplaces
- Add tests

#### Tests
- [x] Yes


server/models/email-campaigns/inactive-workplaces/getParentWorkplaces.js
```diff
@@ -0,0 +1,55 @@
...
+        SELECT
+          ech. "establishmentID"
+        FROM
+          cqc. "EmailCampaignHistories" ech
+          JOIN cqc. "EmailCampaigns" ec ON ec. "id" = ech. "emailCampaignID"
+        WHERE
+          ec. "type" = 'inactiveWorkplaces'
+          AND ech. "createdAt" >= :currentMonth
+          AND ech. "establishmentID" = e. "ParentID");`,
+    {
```
aaron-russell, line 44: Is there a reason why this is raw SQL?

aaron-russell (changes requested):

server/models/email-campaigns/inactive-workplaces/getParentWorkplaces.js
joannafawl: It's because Sequelize doesn't support `UNION` queries

aaron-russell (approved):

PR #3152, opened 2021-03-18
Fix: 24 months plus inactive

chadtomkiss-madetech:
**Issue**

- If a workplace had been more than 24 months inactive, they were not being included in the report.

**Work done**

- Added a `matches` function to each of the templates
- Refactored `getTemplate` to use `matches` function
- 
#### Tests
- [x] Yes


server/test/unit/services/email-campaigns/inactive-workplaces/nextEmail.spec.js
```diff
@@ -95,6 +94,9 @@ describe('nextEmailTemplate', () => {
 
       const emailTemplate = nextEmail.getTemplate(inactiveWorkplace);
 
+      console.log({ emailTemplate });
+      console.log({ NextTemplate });
```
aaron-russell, line 98: Do you need console logs of these?

aaron-russell (changes requested):

joannafawl (approved):

aaron-russell (approved):

PR #3163, opened 2021-03-19
Bug/data permission not set correctly

ross-mcfadyen:
#### Tests
- [x] Yes

### DIFF
--- a/server/routes/ownershipRequest/index.js
+++ b/server/routes/ownershipRequest/index.js
@@ -22,7 +22,7 @@ const ownershipRequest = async (req, res) => {
       rejectionReason: req.body.rejectionReason,
       type: req.body.type,
     };
-    let recieverUpdate;
+    let receiverUpdate;
 
     if (!id) {
       console.error('Missing id or uid');
@@ -59,21 +59,21 @@ const ownershipRequest = async (req, res) => {
 
         //update establishment dataOwner and dataPermissions property for ownership transfer
         if (params.approvalStatus !== 'DENIED') {
-          recieverUpdate = await updateOwnership.update(req, checkOwnerChangeRequest);
+          receiverUpdate = await updateOwnership.update(req, checkOwnerChangeRequest);
         } else {
           let recieverEstablishmentDetails = new Establishment.Establishment(req.username);
           if (await recieverEstablishmentDetails.restore(req.establishment.id, false)) {
-            recieverUpdate = recieverEstablishmentDetails;
+            receiverUpdate = recieverEstablishmentDetails;
           }
         }
 
         // Update notifications after updating Establishment
-        if (recieverUpdate) {
+        if (receiverUpdate) {
           params.exsistingNotificationUid = req.body.exsistingNotificationUid;
           let updateNotificationParam = {
             exsistingNotificationUid: params.exsistingNotificationUid,
             ownerRequestChangeUid: params.ownerRequestChangeUid,
-            recipientUserUid: recieverUpdate.dataOwner !== 'Parent' ? req.userUid : params.recipientUserUid,
+            recipientUserUid: receiverUpdate.dataOwner !== 'Parent' ? req.userUid : params.recipientUserUid,
           };
           let updatedNotificationResp = await notifications.updateNotification(updateNotificationParam);
           if (updatedNotificationResp) {
--- a/server/routes/ownershipRequest/index.js
+++ b/server/routes/ownershipRequest/index.js
@@ -6,6 +6,7 @@ const Establishment = require('../../models/classes/establishment');
 const ownership = require('../../data/ownership');
 const notifications = require('../../data/notifications');
 const Authorization = require('../../utils/security/isAuthenticated');
+const updateOwnership = require('./updateOwnership');
 const { hasPermission } = require('../../utils/security/hasPermission');
 
 // PUT request for ownership change request approve/reject
@@ -21,7 +22,7 @@ const ownershipRequest = async (req, res) => {
       rejectionReason: req.body.rejectionReason,
       type: req.body.type,
     };
-    let owershipRequesterEsatblishmentId, objToUpdate;
+    let recieverUpdate;
 
     if (!id) {
       console.error('Missing id or uid');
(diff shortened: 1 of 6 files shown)


server/routes/ownershipRequest/index.js
```diff
@@ -57,94 +58,60 @@ const ownershipRequest = async (req, res) => {
...
-                    userUid: req.userUid,
-                  };
-                  let saveDataOwnershipRequested = await ownership.changedDataOwnershipRequested(clearOwnershipParam);
-                  let updateOwnershipRequest = await ownership.updateOwnershipRequest(clearOwnershipParam);
-                  if (!saveDataOwnershipRequested && !updateOwnershipRequest) {
-                    return res.status(400).send({
-                      message: 'Invalid request',
-                    });
-                  }
+        // Update notifications after updating Establishment
```
ross-mcfadyen, line 70: I haven't changed any of this, it's just been moved about in the file.

server/routes/ownershipRequest/index.js
```diff
@@ -57,94 +58,60 @@ const ownershipRequest = async (req, res) => {
...
-              req.establishment.id,
-              objToUpdate,
-              true,
-            );
-          } else {
-            let recieverEstablishmentDetails = new Establishment.Establishment(req.username);
-            if (await recieverEstablishmentDetails.restore(req.establishment.id, false)) {
-              recieverUpdate = recieverEstablishmentDetails;
-            }
+          let recieverEstablishmentDetails = new Establishment.Establishment(req.username);
```
ross-mcfadyen, line 64: I haven't changed any of this, it's just been moved about in the file.

server/routes/ownershipRequest/index.js
```diff
@@ -57,94 +58,60 @@ const ownershipRequest = async (req, res) => {
...
-            owershipRequesterEsatblishmentId = ownershipRequesterId[0].ParentID;
-            objToUpdate.dataOwner = 'Parent';
-          } else {
-            owershipRequesterEsatblishmentId = checkOwnerChangeRequest[0].subEstablishmentID;
-          }
-          requesterUpdate = await Establishment.Establishment.fetchAndUpdateEstablishmentDetails(
-            owershipRequesterEsatblishmentId,
-            objToUpdate,
-          );
+          recieverUpdate = await updateOwnership.update(req, checkOwnerChangeRequest);
```
ross-mcfadyen, line 62: This is the new function I've added that replaces lines 60-100 on the left.

server/routes/ownershipRequest/index.js
```diff
@@ -57,94 +58,60 @@ const ownershipRequest = async (req, res) => {
...
-                  let updateOwnershipRequest = await ownership.updateOwnershipRequest(clearOwnershipParam);
-                  if (!saveDataOwnershipRequested && !updateOwnershipRequest) {
-                    return res.status(400).send({
-                      message: 'Invalid request',
-                    });
-                  }
+        // Update notifications after updating Establishment
+        if (receiverUpdate) {
+          params.exsistingNotificationUid = req.body.exsistingNotificationUid;
+          let updateNotificationParam = {
```
chadtomkiss-madetech, line 73: You can use `const` if it's not re-assigned

server/routes/ownershipRequest/index.js
```diff
@@ -57,94 +58,60 @@ const ownershipRequest = async (req, res) => {
...
+            exsistingNotificationUid: params.exsistingNotificationUid,
+            ownerRequestChangeUid: params.ownerRequestChangeUid,
+            recipientUserUid: receiverUpdate.dataOwner !== 'Parent' ? req.userUid : params.recipientUserUid,
+          };
+          let updatedNotificationResp = await notifications.updateNotification(updateNotificationParam);
```
chadtomkiss-madetech, line 78: Also `const` here

server/routes/ownershipRequest/updateOwnership.js
```diff
@@ -0,0 +1,28 @@
...
+    dataPermissions: ownershipChangeRequest[0].permissionRequest,
+  };
+
+  if (currentDataOwnerDetails[0].IsParent === false && currentDataOwnerDetails[0].ParentID) {
+    updateDetails.dataOwner = 'Parent';
+    workplaceEstablishmentId = request.establishment.id;
+  } else {
+    updateDetails.dataOwner = 'Workplace';
+    workplaceEstablishmentId = ownershipChangeRequest[0].subEstablishmentID;
+  }
```
chadtomkiss-madetech, line 19: I think we could do avoid `if/else` by maybe doing something like:

```
const currentDataOwnerDetails = await ownership.getownershipRequesterId(request.establishment.id);

const isSubsidiary = currentDataOwnerDetails[0].IsParent === false && currentDataOwnerDetails[0].ParentID;

const updateDetails = {
  dataPermissions: ownershipChangeRequest[0].permissionRequest,
  dataOwner: (isSubsidiary) ? 'Parent' : 'Workplace',
}

const workplaceEstablishmentId = (isSubsidiary) ? request.establishment.id : ownershipChangeRequest[0].subEstablishmentID;

return await Establishment.fetchAndUpdateEstablishmentDetails(workplaceEstablishmentId, updateDetails);
```

chadtomkiss-madetech (commented):
Only commenting due to being on hols next week but ace work @ross-mcfadyen 😍

server/routes/ownershipRequest/updateOwnership.js
ross-mcfadyen: Fair point, I did toy with something similar, but I have a bit of a dislike for ternary unless it's very, very basic. Personally I find multiple ternary statements really difficult to read without going over it several times.

There probably is a better way though that isn't if/else, I agree. Be interested to hear what other people think....

chadtomkiss-madetech: Yeah agreed, the fact the ternary is used multiple times makes it a bit confusing actually

I think the main issue for me is:

1) Adding the same key to the object in the `if/else` but just to a diff value
2) Re-assigning `workplaceEstablishmentId` to the same value as it's original in the `if`
2) `else` is a code smell and should be avoided if possible

ross-mcfadyen: Good points. I'll see what I can do with those 3 in mind 👍

ross-mcfadyen: @chadtomkiss-madetech - After all that, I went with the ternary 😆 I tried a few other options and , on reflection, it's the cleanest way and makes sense now I've gone through it a few times.

server/routes/ownershipRequest/index.js
ross-mcfadyen: I think there are a lot of places where this could be tidied up, but I'm leaving these as they are as I haven't touched the notification code (and don't want to risk breaking it as I'm not testing the notifications). These have just moved in the file.

aaron-russell (approved):

PR #3174, opened 2021-03-22
Fix: Accessibility Audit

joannafawl:
#### Changes
- Change colour of orange banner to increase contrast
- Update `alt` text on flags so they are only used where they don't already have a description
- Change `sortBy` labels to be unique

#### Tests
- [ ] Yes

### DIFF
--- a/src/app/shared/components/tabs/tabs.component.html
+++ b/src/app/shared/components/tabs/tabs.component.html
@@ -20,11 +20,9 @@
         {{ tab.title }}
       </a>
 
-      <i
-        alt="Missing or expired record alert"
-        [class.govuk-tabs__tab--alert]="tab.alert"
-        [class.govuk-tabs__tab--red-alert]="tab.redAlert"
-      ></i>
+      <i [class.govuk-tabs__tab--alert]="tab.alert" [class.govuk-tabs__tab--red-alert]="tab.redAlert">
+        <span class="govuk-visually-hidden">Missing or expired record alert</span>
+      </i>
     </li>
   </ng-container>
 </ul>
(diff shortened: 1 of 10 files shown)


src/app/shared/components/tabs/tabs.component.html
```diff
@@ -20,7 +20,11 @@
         {{ tab.title }}
       </a>
 
-      <i [class.govuk-tabs__tab--alert]="tab.alert" [class.govuk-tabs__tab--red-alert]="tab.redAlert"></i>
+      <i
+        alt="Missing or expired record alert"
```
aaron-russell, line 24: This will need to be hidden visually instead, as an `<i>` tag can't have an alt

src/app/shared/components/total-staff-panel/total-staff-panel.component.html
```diff
@@ -35,7 +35,7 @@
       <ng-template #deleteText>delete</ng-template>
     </dt>
     <dd data-testid="changeNumber">
-      <img src="/assets/images/flag-orange.svg" alt="Warning" />
+      <img src="/assets/images/flag-orange.svg" alt="" />
```
aaron-russell, line 38: This will need an alt as it doesn't mention what the image is showing or why it's showing

aaron-russell (changes requested):

src/app/shared/components/total-staff-panel/total-staff-panel.component.html
joannafawl: I wasn't sure with this one because the alt text for orange flags is `Expiring record alert` but this one is for missing records

aaron-russell: It will be `There are Missing records` or similar

aaron-russell (approved):

PR #3186, opened 2021-03-23
Fixed bug where is no primary user in payload, would default to false

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server/models/classes/user.js
+++ b/server/models/classes/user.js
@@ -239,12 +239,7 @@ class User {
       }
 
       if (this._isBool(document.isPrimary)) {
-        // by explicitly checking for "true", don't have to worry about any other value
-        if (document.isPrimary === true) {
-          this._isPrimary = true;
-        } else {
-          this._isPrimary = false;
-        }
+        this._isPrimary = document.isPrimary;
       }
       if (document.isActive) {
         this._active = document.isActive;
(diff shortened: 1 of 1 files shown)


server/models/classes/user.js
```diff
@@ -231,7 +238,7 @@ class User {
         this._password = escape(document.password);
       }
 
-      if (document.isPrimary !== null) {
+      if (this._isBool(document.isPrimary)) {
         // by explicitly checking for "true", don't have to worry about any other value
         if (document.isPrimary === true) {
```
duncanc19, line 243: If you've already checked that it's a bool, can't you remove the if else and just assign it to document.isPrimary?

g-m-taylor (approved):

server/models/classes/user.js
```diff
@@ -217,6 +217,13 @@ class User {
     }
   }
 
+  _isBool(value) {
+    if (typeof value === 'boolean') {
+      return true;
+    }
+    return false;
+  }
```
duncanc19, line 225: Could you just do return typeof value === 'boolean'?

server/models/classes/user.js
aaron-russell: Good shout, done!

server/models/classes/user.js
aaron-russell: Another good shout, done

duncanc19 (approved):

PR #3192, opened 2021-03-25
Fix: Inactive workplaces report

chadtomkiss-madetech:
**Preface**

The inactive workplaces report had a lot of columns that were empty since we refactored how the parent object was being built. 

The reason they were empty was due to the database results being different to the objects we were stubbing in our tests - the query was renaming `NmdsID` to `btrim` and the query was also missing `ParentID` and `IsParent`.

We also noticed that subsidiaries who owned their own data were not being included in the single workplace report.

**Issues**

- We added `TRIM` to `NmdsID` and this actually names the column as `btrim`, so `workplace.NmdsID` was undefined
- We were missing `ParentID` and `IsParent` from the parent query
- We weren't including subsidiaries who owned their own data in the single workplaces report
- If a parent is inactive but has no subsidiaries, `subsidiaries` wasn't being initialised as an empty array so we were mapping over undefined when transforming
-  We were missing `.endOf('month').endOf('day')` from the parent date comparison
- There was no `console.error` when the `api/admin/email-campaigns/inactive-workplaces` endpoint threw an exception

**Work done**

- Updated query for single workplaces and parents:
  -  Fix `TRIM` alias
  - Add missing `ParentID` and `IsParent` to parent workplaces query
  - Include subsidiaries who own their own data in the single workplaces report
- Fixed `buildWorkplaces` to initialise subsidiaries as an empty array
- Fixed date comparison to always be end of the month/day
- Added `console.error` to the `api/admin/email-campaigns/inactive-workplaces` endpoint

#### Tests
- [x] Yes


server/services/email-campaigns/inactive-workplaces/findParentWorkplaces.js
```diff
@@ -9,6 +9,9 @@ const buildWorkplaces = (workplaces) => {
   return workplaces.reduce((acc, workplace) => {
     if (workplace.IsParent) {
       acc[workplace.EstablishmentID] = {
+        ...{
+          subsidiaries: [],
+        },
```
aaron-russell, line 14: Why not just add subsidiaries to the object? 
```suggestion
      subsidiaries: [],
```

aaron-russell (changes requested):

aaron-russell (approved):

PR #3269, opened 2021-04-13
Feat/second user banner

duncanc19:
#### Changes
- Create Add second user banner to Users section on home page
- Add orange flag to Users tab when Add second user banner is showing
- Change positioning of tab flags to move towards top right 

#### Tests
- [X] Yes

### DIFF
--- a/src/app/features/dashboard/dashboard.component.ts
+++ b/src/app/features/dashboard/dashboard.component.ts
@@ -32,6 +32,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
   public canViewBenchmarks: boolean;
   public workplaceUid: string;
   public showUsersTabFlag: boolean;
+  public canAddUser: boolean;
 
   constructor(
     private alertService: AlertService,
@@ -54,6 +55,7 @@ export class DashboardComponent implements OnInit, OnDestroy {
     this.canViewListOfWorkers = this.permissionsService.can(this.workplaceUid, 'canViewListOfWorkers');
     this.canViewEstablishment = this.permissionsService.can(this.workplaceUid, 'canViewEstablishment');
     this.canDeleteEstablishment = this.permissionsService.can(this.workplaceUid, 'canDeleteAllEstablishments');
+    this.canAddUser = this.permissionsService.can(this.workplace.uid, 'canAddUser');
 
     if (this.workplace) {
       this.subscriptions.add(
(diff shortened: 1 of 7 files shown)


src/app/features/dashboard/dashboard.component.ts
```diff
@@ -46,16 +48,16 @@ export class DashboardComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.authService.isOnAdminScreen = false;
     this.workplace = this.establishmentService.primaryWorkplace;
-    const workplaceUid: string = this.workplace ? this.workplace.uid : null;
-    this.canViewBenchmarks = this.permissionsService.can(workplaceUid, 'canViewBenchmarks');
-    this.canViewListOfUsers = this.permissionsService.can(workplaceUid, 'canViewListOfUsers');
-    this.canViewListOfWorkers = this.permissionsService.can(workplaceUid, 'canViewListOfWorkers');
-    this.canViewEstablishment = this.permissionsService.can(workplaceUid, 'canViewEstablishment');
-    this.canDeleteEstablishment = this.permissionsService.can(workplaceUid, 'canDeleteAllEstablishments');
+    this.workplaceUid = this.workplace ? this.workplace.uid : null;
```
aaron-russell, line 51: You can still have a type definition here if you want 
```suggestion
    this.workplaceUid: string | null = this.workplace ? this.workplace.uid : null;
```

duncanc19: It's not being declared here though, it's got the type on line 33, is that ok?

src/app/features/dashboard/dashboard.component.ts
```diff
@@ -92,6 +96,11 @@ export class DashboardComponent implements OnInit, OnDestroy {
           this.subsidiaryCount = res.subsidaries ? res.subsidaries.count : 0;
         }),
       );
+      this.subscriptions.add(
+        this.userService.getAllUsersForEstablishment(this.workplaceUid).subscribe((users) => {
+          this.showUsersTabFlag = users.length === 1 && this.canAddUser;
```
chadtomkiss-madetech, line 101: In `user-accounts-summary` component the logic is `canAddUser && users.length === 1`

Feel this reads better as the permission is the first check 👍

src/app/features/dashboard/dashboard.component.ts
chadtomkiss-madetech: @duncanc19 think it just needs the `| null` on the type to communicate it can also be set to null

chadtomkiss-madetech (approved):

PR #3283, opened 2021-04-16
Fixed bug where it wouldn't accept an email with longer than 4 tld

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server/models/classes/user/properties/emailProperty.js
+++ b/server/models/classes/user/properties/emailProperty.js
@@ -2,56 +2,54 @@
 const ChangePropertyPrototype = require('../../properties/changePrototype').ChangePropertyPrototype;
 
 exports.UserEmailProperty = class UserEmailProperty extends ChangePropertyPrototype {
-    constructor() {
-        super('Email');
+  constructor() {
+    super('Email');
+  }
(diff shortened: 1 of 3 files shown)


server/models/classes/user/properties/emailProperty.js
```diff
@@ -2,56 +2,54 @@
...
-        }
+  // concrete implementations
+  async restoreFromJson(document) {
+    if (document.email) {
+      const EMAIL_MAX_LENGTH = 120;
+      const emailRegex = /^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,63}$/;
+      if (document.email.length > 0 && document.email.length <= EMAIL_MAX_LENGTH && emailRegex.test(document.email)) {
+        this.property = document.email;
+      } else {
+        this.property = null;
```
chadtomkiss-madetech, line 21: This regex is slightly different to the email pattern on the front-end - does that need to change as well?

The `+\.` is `+.` in `src/app/core/constants/constants.ts`

chadtomkiss-madetech (approved):

PR #3294, opened 2021-04-21
Adding decryption to wdf parent report

g-m-taylor:
#### Tests
- [ ] Yes

### DIFF
--- a/server/data/parentWDFReport.js
+++ b/server/data/parentWDFReport.js
@@ -1,7 +1,8 @@
 'use strict';
 
 const db = require('../utils/datastore');
-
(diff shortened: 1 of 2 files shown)


server/data/parentWDFReport.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
 const db = require('../utils/datastore');
-
+var config = require('../../server/config/config');
```
chadtomkiss-madetech, line 4: This could be a `const`

chadtomkiss-madetech (approved):

PR #3298, opened 2021-04-22
Added indexes on audit columns

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/migrations/20210422075222-auditIndexes.js
+++ b/migrations/20210422075222-auditIndexes.js
@@ -3,9 +3,10 @@
 module.exports = {
   up: async (queryInterface, Sequelize) => {
     return queryInterface.sequelize.query(`
-      CREATE INDEX "worker_audit__worker_f_k__property_name__event_type__when" ON cqc."WorkerAudit" USING btree ("WorkerFK", "EventType", "When");
-      CREATE INDEX "establishment_audit__establishment_f_k__when" ON cqc."EstablishmentAudit" USING btree ("EstablishmentFK", "When");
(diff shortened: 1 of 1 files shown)


migrations/20210422075222-auditIndexes.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.query(`
+      CREATE INDEX "worker_audit__worker_f_k__property_name__event_type__when" ON cqc."WorkerAudit" USING btree ("WorkerFK", "EventType", "When");
```
chadtomkiss-madetech, line 6: We should create the indexes concurrently to avoid locking up the audit table

You can do that with `CREATE INDEX CONCURRENTLY`

Also the default start up timeout is 5 minutes - is there a chance this could take longer when running against production?

May need to do this kind of database work separately to deployment 🤔

aaron-russell: Seems like this sequelize has it's quirks when it comes to index's. I had a problem with adding them with `queryInterface.addIndex` so moved to this method. If I try add them via this script concurrently: `ERROR: CREATE INDEX CONCURRENTLY cannot be executed from a function or multi-command string`

Think it makes sense to do this sort of thing separately. I'll close this.

chadtomkiss-madetech: @aaron-russell Hmm.. I wonder what Sequelize is doing internally - I know that you can't create indexes concurrently in a transaction

aaron-russell: I have a feeling it wraps every migration in a transaction. I'll add them to DB repo instead and run them manually. Where did you put the various views/indexes you made?

PR #3303, opened 2021-04-22
Fix/wdf headings and tab icons

duncanc19:
#### Changes
- Move icons in tabs down a bit
- Make new WDF headings l instead of xl

#### Tests
- [ ] Yes

### DIFF
--- a/src/assets/scss/components/_tabs.scss
+++ b/src/assets/scss/components/_tabs.scss
@@ -4,29 +4,25 @@
   border: 0;
 }
 
-.govuk-tabs__list-item .govuk-tabs__tab--alert {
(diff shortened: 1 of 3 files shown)


src/assets/scss/components/_tabs.scss
```diff
@@ -28,5 +28,5 @@
   background-position: calc(100%);
   background-size: 22px;
   padding-left: 30px;
-  padding-top: 15px;
+  padding-top: 10px;
```
aaron-russell, line 31: Is there a reason why these can't all share the same CSS? Pointless having 3 bits of code to do the same thing

aaron-russell (changes requested):

aaron-russell (approved):

PR #3326, opened 2021-04-23
Bug/address format

duncanc19:
#### Changes
- Fix issue where sub building name isn't shown in addresses
- Fix issue where building number and street are on separate lines on address input page
- Refactor postcodes endpoint
- Add tests to postcodes endpoint

#### Tests
- [X] Yes


server/routes/postcodes.js
```diff
@@ -1,74 +1,85 @@
...
+    let data = results[i].dataValues;
 
-    try {
-      //Clean user submitted postcode
-      let cleanPostcode= pCodeCheck.sanitisePostcode(req.params.postcode);
+    const workplaceAndAddress = createAddressObject(data);
+    postcodeData.push(workplaceAndAddress);
+  }
+  return postcodeData;
+};
```
aaron-russell, line 16: Why not just use a map? You should be able to do something like this

```javascript
const postcodeData = results.map((result) => createAddressObject(result));
```

aaron-russell (changes requested):

aaron-russell (approved):

PR #3390, opened 2021-05-13
Feat: WDF parent overview

joannafawl:
#### Work done
- Added logic to display different WDF overview page for parents when both eligible and not eligible

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/wdf/wdf-overview/wdf-overview.component.ts
+++ b/src/app/features/wdf/wdf-overview/wdf-overview.component.ts
@@ -42,7 +42,7 @@ export class WdfOverviewComponent implements OnInit, OnDestroy {
     this.workplace = this.establishmentService.primaryWorkplace;
     this.isParent = this.workplace.isParent;
 
-    this.getParentsAndSubs();
+    this.getParentAndSubs();
     this.getWdfReport();
   }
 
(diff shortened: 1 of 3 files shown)


aaron-russell (approved):

src/app/features/wdf/wdf-overview/wdf-overview.component.ts
```diff
@@ -14,23 +17,66 @@ import { Subscription } from 'rxjs';
...
+
+  public getParentOverallWdfEligibility(): void {
+    this.parentOverallWdfEligibility = !this.workplaces.some((workplace) => {
+      return workplace.wdf.overall === false;
+    });
+  }
+
+  public getLastOverallEligibilityDate(): void {
+    if (this.parentOverallWdfEligibility) {
+      this.parentOverallEligibilityDate = moment(this.workplaces[0].wdf.overallWdfEligibility).format('D MMMM YYYY');
```
aaron-russell, line 75: Should this always be the first item? Or the latest any of their workplaces passed?

joannafawl: I was about to say that we order `workplaces` by WDF eligibility then by the latest updated on line 58, but I've just realised we put the primary workplace at the front in the line below so it won't work, could we call `orderBy` after unshifting the parent or do you think we should explicitly check the latest date?

aaron-russell: I guess it depends on the design, and what date they would be expecting?

joannafawl: The date is used to show when they met WDF requirements for all workplaces, so it's the WDF eligibility date of the last workplace to meet requirements

aaron-russell (approved):

PR #3405, opened 2021-05-14
Tech debt/ff config cat - Adding Feature Flag check to circleCI

g-m-taylor:
#### Tests
- [ ] Yes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -416,8 +416,9 @@ workflows:
           requires:
             - install
       - configcat/validate-flag-references:
-            debug: true
-            fail-on-warnings: true
+          name: feature_flag
+          debug: true
+          fail-on-warnings: true
       - blue_green:
           name: blue_green_benchmarks
           appname: sfcbenchmarks
@@ -448,7 +449,7 @@ workflows:
             - test_frontend
             - test_backend
             - performance
-            - feature_flag_staging
+            - feature_flag
           filters:
             branches:
               only:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -196,6 +196,30 @@ jobs:
             - node-deps-v2-{{ .Branch }}-{{ checksum "package-lock.json" }}
       - run:
           command: $(npm bin)/lhci autorun
+  feature_flag:
+    description: Execute a blue/green deploy  in a single job. Expects either build_steps
+      or workspace_path for assets to deploy.
```
aaron-russell, line 201: Should this description not describe the ff step?

.circleci/config.yml
```diff
@@ -386,6 +410,12 @@ workflows:
               only: /^v[0-9]+(\.[0-9]+)*$/
           requires:
             - install
+      - feature_flag:
+          filters:
+            tags:
+              only: /^v[0-9]+(\.[0-9]+)*$/
+          requires:
+            - install
```
aaron-russell, line 418: are you only running it on deploy to prod and none of the other env's? Surely this is most beneficial on staging PR's?

aaron-russell (changes requested):

.circleci/config.yml
g-m-taylor: I dont think that is what the filter is doing?  Its the same as performance and the two test steps, which run in every envo?

aaron-russell (approved):

PR #3421, opened 2021-05-20
Changed the way the worker file is downloaded to use database rather than restoring workers

aaron-russell:
#### Tests
- [ ] Yes

### DIFF
--- a/server/models/BulkImport/csv/workers.js
+++ b/server/models/BulkImport/csv/workers.js
@@ -1,5 +1,6 @@
 const BUDI = require('../BUDI').BUDI;
 const moment = require('moment');
+const get = require('lodash/get');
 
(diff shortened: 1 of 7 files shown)


server/models/BulkImport/csv/workers.js
```diff
@@ -3916,102 +3857,119 @@ class Worker {
...
+    switch (entity.AnnualHourlyPayValue) {
+      case 'Hourly':
+        // "SALARYINT"
+        columns.push(3);
+        // "SALARY"
+        columns.push('');
+        // "HOURLYRATE"
+        columns.push(entity.AnnualHourlyPayRate);
+        break;
+      case 'Anually':
```
duncanc19, line 3880: Should this case be 'Annually'?

aaron-russell: Good catch, thank you!

duncanc19 (approved):

PR #3485, opened 2021-06-10
Added preprod build

aaron-russell:
#### Tests
- [ ] Yes


joannafawl (approved):

src/environments/environment.preprod.ts
```diff
@@ -0,0 +1,6 @@
+export const environment = {
+  production: true,
+  environmentName: 'production',
```
duncanc19, line 3: Is the environment name meant to be preproduction?

PR #3502, opened 2021-06-16
Feat: Targeted emails

aaron-russell:
#### Work done
- Add functionality for admins to send emails to targeted users by selecting a user group and template
- Create `allPrimaryUsers` function in user model to get all primary users
- Add `getTemplates` function to get templates using Send In Blue API

#### Tests
- [x] Yes

### DIFF
--- a/server/services/email-campaigns/isWhitelisted.js
+++ b/server/services/email-campaigns/isWhitelisted.js
@@ -1,0 +1,13 @@
+const config = require('../../config/config');
+
+const isWhitelisted = (email) => {
+  if (!config.get('sendInBlue.whitelist')) {
+    return true;
+  }
+
+  return config.get('sendInBlue.whitelist').split(',').includes(email);
+};
+
+module.exports = {
+  isWhitelisted,
+};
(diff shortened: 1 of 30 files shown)


server/services/email-campaigns/isWhitelisted.js
```diff
@@ -0,0 +1,13 @@
+const config = require('../../config/config');
+
+const isWhitelisted = (email) => {
+  if (!config.get('sendInBlue.whitelist')) {
+    return true;
```
g-m-taylor, line 5: if there is an error in the config would this just let every email through? Would it be safer to return false?

server/services/email-campaigns/limit.js
```diff
@@ -0,0 +1,8 @@
+const { pRateLimit } = require('p-ratelimit');
+
+const limit = pRateLimit({
+      interval: 1000,
+      rate: 15, // 5 emails per second
+});
```
g-m-taylor, line 6: is this comment still correct? above it at rate:5 = 5 emails per second?

server/services/email-campaigns/isWhitelisted.js
aaron-russell: The whitelist is just used during testing to only send the relevant people, the only time it would be an issue is if there wasn't a whitelist at all.

server/services/email-campaigns/limit.js
aaron-russell: Thanks 👍

g-m-taylor (approved):

PR #3503, opened 2021-06-17
Feat/about us page

duncanc19:
#### Work done
- Created pages service which retrieves pages from DirectUS
- Created a page-content component which can be reused for future pages
- Created about us page with content and article list at side
- Added tests for page-content and about-us components
- Made article and page factories for tests 

#### Tests
- [X] Yes


src/app/shared/components/page-content/page-content.component.html
```diff
@@ -0,0 +1,4 @@
+<h1 class="govuk-heading-l govuk-!-margin-bottom-5">
+  {{ page.title }}
+</h1>
```
aaron-russell, line 3: Should you be showing the title on the page or should it be part of the content? Also does this mean there's duplicate `h1`'s?

aaron-russell (changes requested):

src/app/shared/components/page-content/page-content.component.html
duncanc19: This component contains the title and content and the title's not in the about us component. It means we can just retrieve the data from the snapshot in this component instead of having to retrieve it every time we make a new page

duncanc19: Should we change the name of the component or do you think the data should be retrieved in the about us component?

rramgolam: The `page-content` component has been renamed to `page` to cater for title + content.

aaron-russell: Does it work for screen readers? I thought the blue heading was a `h1` that's all, and screenreaders don't like 2 `h1`'s

duncanc19: For pages, there isn't a caption part to the heading, articles have the caption part but they don't use this component. The article component still has the h1 with spans for the ASC-WDS news caption and the article.title

aaron-russell (approved):

PR #3504, opened 2021-06-17
Auto confirm fix

g-m-taylor:
Fixing a bug where anytime you go onto a wdf staff page it would auto update the fields that dont need confirmation ( e.g date of birth), which in turn would change the Last Updated date for the staff without any user input.
Its now been checked so that this only happens once all other fields are up-to-date. 
There was also one field missing which has been added in.

Adding missing auto confirm field,
Moving auto confirm to only happen once every other field is confirmed.
adding tests

#### Tests
- [x] Yes

### DIFF
--- a/src/app/shared/components/staff-record-summary/staff-record-summary.component.ts
+++ b/src/app/shared/components/staff-record-summary/staff-record-summary.component.ts
@@ -1,5 +1,5 @@
 import { Location } from '@angular/common';
-import { Component, Input, OnChanges, OnInit } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Establishment } from '@core/model/establishment.model';
 import { URLStructure } from '@core/model/url.model';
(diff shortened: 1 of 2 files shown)


aaron-russell (approved):

src/app/shared/components/staff-record-summary/staff-record-summary.component.ts
```diff
@@ -87,26 +88,58 @@ export class StaffRecordSummaryComponent implements OnInit {
...
+      }
+      if (this.worker.wdf?.[fieldCheck].isEligible === 'Yes' && !this.worker.wdf?.[fieldCheck].updatedSinceEffectiveDate) {
+        if ((fieldCheck === 'careCertificate' && this.worker.careCertificate === 'Yes, completed') ||
+        (fieldCheck === 'qualificationInSocialCare' && this.worker.careCertificate === 'Yes') ){
+          this.confirmField(fieldCheck);
           continue;
         }
-        this.confirmField(field);
+        this.confirmField(fieldCheck);
       }
```
aaron-russell, line 143: What is the point in this if, it it confirms anyway?

aaron-russell (changes requested):

src/app/shared/components/staff-record-summary/staff-record-summary.component.ts
g-m-taylor: ah yeah the logic is backwards.  I have updated it and added some tests around that

aaron-russell (approved):

PR #3541, opened 2021-06-23
Bug/wdf staff record exit: 

g-m-taylor:
Fixing an error where when you go to the next staff member on the wdf, it doesnt set the return currectly.
Changed it so it doesn't pass the return address down but instead you set it in the worker/workplace service before you get to the page, and it then draws from that. 
#### Tests
- [ ] Yes


src/app/shared/components/staff-record-summary/basic-record/basic-record.component.html
```diff
@@ -10,7 +10,6 @@ <h2 class="govuk-heading-m">{{ basicTitle }}</h2>
         [explanationText]="' basic record details'"
         [link]="getRoutePath('staff-details')"
         [hasData]="true"
-        (setReturnClicked)="this.setReturn()"
```
aaron-russell, line 13: This wouldn't have worked anyway right? `this.` shouldn't be in the template?

aaron-russell (approved):
Good work! Looks good to me, not sure what they were thinking of originally...

src/app/shared/components/staff-record-summary/basic-record/basic-record.component.html
g-m-taylor: I think you can use this, its just not actually needed. but not entirely sure.

PR #3542, opened 2021-06-23
Feat/remove nino and dob show and hide for admin

rramgolam:
### Work Done
- Add permissions canViewNinoDob to permissions service and permission type to model.
- Add rendering logic for admin and non-admin cases.
- Add tests for both scenarios.

#### Tests
- [X] Yes


src/app/shared/components/staff-record-summary/personal-details/personal-details.component.html
```diff
@@ -25,13 +25,16 @@ <h2 class="govuk-heading-m">Personal details</h2>
         <div class="govuk-!-display-inline-block govuk-!-width-one-third">
           {{ ninoHidden ? '*********' : worker.nationalInsuranceNumber }}
         </div>
-        <div class="govuk-!-display-inline-block govuk-!-width-two-thirds">
-          <a class="govuk-link" href="#" (click)="toggleNinoHide($event)">
-            {{ ninoHidden ? 'Show' : 'Hide' }}
-          </a>
-        </div>
+        <ng-container *ngIf="!isAdmin; else showNino">
```
aaron-russell, line 28: Is this not going to use the new permission you have created rather than the admin flag?

aaron-russell (changes requested):

src/app/shared/components/staff-record-summary/personal-details/personal-details.component.html
rramgolam: `isAdmin` is actually coming from the new permission I created, in staff-record-summary, I fetch and set it: `this.isAdmin = this.permissionsService.can(this.workplaceUid, 'canViewNinoDob');`, then pass it along to the personal-details component. 

Shall I rename `isAdmin` to `canViewNinoDob` to make it clearer?

rramgolam: (also I realised that `else showNino` doesn't need to be there, I've pushed that commit to remove it already - hence the Outdated note)

aaron-russell: I think that makes more sense, otherwise `isAdmin` will be wrong later

aaron-russell (approved):

PR #3569, opened 2021-06-29
Bug/staffmismatchzero

g-m-taylor:
Making sure the banner shows when total staff is zero.
Moving function to OnChange so it only runs when needed. 

#### Tests
- [x] Yes

### DIFF
--- a/src/app/shared/components/workplace-summary/workplace-summary.component.html
+++ b/src/app/shared/components/workplace-summary/workplace-summary.component.html
@@ -1,7 +1,7 @@
 <dl
   *ngIf="workplace"
   class="govuk-summary-list"
-  [class.govuk-summary-list--contains-warnings]="totalStaffWarning || staffMismatchWarning()"
+  [class.govuk-summary-list--contains-warnings]="this.showTotalStaffWarning || staffMismatchWarning()"
 >
   <div class="govuk-summary-list__row">
     <dt class="govuk-summary-list__key">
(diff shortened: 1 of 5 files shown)


src/app/shared/components/workplace-summary/workplace-summary.component.html
```diff
@@ -3,8 +3,8 @@
   class="govuk-summary-list"
   [ngClass]="{
     'govuk-summary-list--contains-warnings':
-      (!wdfNewDesign || !wdfView) && (totalStaffWarning || staffMismatchWarning()),
-    'asc-summary-list--wide-value': wdfView && wdfNewDesign && totalStaffWarning
+      (!wdfNewDesign || !wdfView) && (this.showTotalStaffWarning || staffMismatchWarning()),
```
aaron-russell, line 6: What is the difference between `totalStaffWarning` and `this.totalStaffWarning`?

aaron-russell (approved):

src/app/shared/components/workplace-summary/workplace-summary.component.html
g-m-taylor: tbh nothing other than force of habit, write this. just shows up all the available items  when on webstorm.

aaron-russell: Should it be consistent then and not use this?

PR #3745, opened 2021-07-19
Added thank you component

aaron-russell:
#### Work done
- Added thank you component
- Removed old awaiting approval component

#### Tests
- [x] Yes


src/app/features/registration/registration-routing.module.ts
```diff
@@ -147,10 +145,10 @@ const routes: Routes = [
     data: { title: 'Complete' },
   },
   {
-    path: 'awaiting-approval',
-    component: RegistrationAwaitingApprovalComponent,
-    canActivate: [RegisterGuard],
-    data: { title: 'Awaiting Approval' },
+    path: 'thank-you',
+    component: ThankYouComponent,
+    // canActivate: [RegisterGuard],
```
duncanc19, line 150: If it doesn't use the guard can the comment go?

aaron-russell: I've added it back in, thank you, just makes it easy to test rather than going through the form

duncanc19 (approved):

PR #3779, opened 2021-07-26
Set dates for LA return

aaron-russell:
#### Work done
- Added a new database table for admin settings
- Added a new endpoint for getting the la return dates
- Added a new endpoint for setting the la return dates
- Created screen to modify the dates in the new admin area

#### Tests
- [X] Yes


server/routes/admin/local-authority-return/dates/index.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+};
+
+const setLAReturnDates = async (req, res) => {
+  try {
+    const startDate = req.body.laReturnStartDate;
+    const endDate = req.body.laReturnEndDate;
+
+    const [laReturnStartDateCount] = await models.AdminSettings.setValue('laReturnStartDate', formatDates(startDate));
+    const [laReturnEndDateCount] = await models.AdminSettings.setValue('laReturnEndDate', formatDates(endDate));
```
joannafawl, line 38: How come these are arrays?

server/test/unit/routes/admin/local-authority-returns/dates/index.spec.js
```diff
@@ -0,0 +1,132 @@
+const expect = require('chai').expect;
+const httpMocks = require('node-mocks-http');
+const sinon = require('sinon');
+const models = require('../../../../../../models');
+
+const { getLAReturnDates, setLAReturnDates } = require('../../../../../../routes/admin/local-authority-return/dates');
+
+describe.only('server/routes/admin/local-authority-returns/dates', async () => {
```
joannafawl, line 8:
```suggestion
describe('server/routes/admin/local-authority-returns/dates', async () => {
```

src/app/core/breadcrumb/journey.admin.ts
```diff
@@ -1,12 +1,22 @@
...
+      path: '/sfcadmin',
+      title: 'Admin',
+      children: [
+        {
+          path: '/sfcadmin/local-authorities-return',
+          title: 'Local authorities return',
+          children: [
+            {
+              path: '/sfcadmin/local-authorities-return/set-dates',
+              title: 'Set start and end date',
```
joannafawl, line 15: Will these urls be `sfcadmin/scfadmin/...`?

src/app/features/admin/admin.routing.module.ts
```diff
@@ -1,14 +1,16 @@
...
 
 const routes: Routes = [
-  {
-    path: '',
-    redirectTo: 'search',
-  },
+  // {
+  //   path: '',
+  //   redirectTo: 'search',
+  // },
```
joannafawl, line 13: Does this still need to be commented out?

server/routes/admin/local-authority-return/dates/index.js
aaron-russell: Because the output from sequelize looks like this: 

```javascript
[ 1 ]
```

1 would be the number of rows updated. We only want that number within the array.

src/app/core/breadcrumb/journey.admin.ts
aaron-russell: No, the URL's are from root

joannafawl: I thought parent routes prefix the children path? It looks a bit confusing

src/app/features/admin/admin.routing.module.ts
aaron-russell: Thanks, forgot about that. I have fixed that now.

src/app/core/breadcrumb/journey.admin.ts
aaron-russell: They are all like that. Sometimes the URL's are variables but they are all the same

joannafawl (approved):

PR #3783, opened 2021-07-27
Feat/is this your workplace directive

duncanc19:
#### Work done
- Created is-this-your-workplace directive and extend it from components in create-account and add-workplace
- Set new component in add-workplace flow to use workplace service
- Moved HTML template into folder with directive
- Added setting of selectedLocationAddress in directive

#### Tests
- [X] Yes


src/app/features/add-workplace/is-this-your-workplace/is-this-your-workplace.component.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+      spy,
+    };
+  }
+
+  it('should render a IsThisYourWorkplaceComponent', async () => {
+    const { component } = await setup();
+    expect(component).toBeTruthy();
+  });
+
+  describe('Parent journey', () => {
```
joannafawl, line 98: You could probably get rid of this describe line because it's all the parent journey

joannafawl (approved):

PR #3789, opened 2021-07-29
Added wider class to new admin screens

aaron-russell:
#### Work done
- Added new class depending on the current route you are on
- Added isAdmin flag to existing code to follow route changes

#### Tests
- [ ] Yes

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -16,7 +16,7 @@ import { filter, take, takeWhile } from 'rxjs/operators';
 })
 export class AppComponent implements OnInit {
   private baseTitle = 'Skills for Care';
-  public admin = false;
+  public isAdminSection = false;
   @ViewChild('top') top: ElementRef;
   @ViewChild('content') content: ElementRef;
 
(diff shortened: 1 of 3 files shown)


src/app/app.component.ts
```diff
@@ -6,16 +6,17 @@ import { NavigationEnd, Router } from '@angular/router';
...
 import { filter, take, takeWhile } from 'rxjs/operators';
-import { FeatureFlagsService } from '@shared/services/feature-flags.service';
 
 @Component({
   selector: 'app-root',
   templateUrl: './app.component.html',
 })
 export class AppComponent implements OnInit {
   private baseTitle = 'Skills for Care';
+  public admin = false;
```
duncanc19, line 19: I think it'd be more descriptive to call it something like isAdminSection, because it's not just based on whether the user is an admin or not

src/assets/scss/components/_grid.scss
```diff
@@ -82,3 +82,9 @@
     }
   }
 }
+
+.asc-width-container {
```
duncanc19, line 86: You could add something to the class name too to show it's related to the admin section? Like asc-admin-section-width-container or something?

duncanc19 (approved):

PR #3826, opened 2021-08-03
Feat/adding l as

Richard-Pentecost:
#### Work done
- create migration file for new local authorities table
- create model for local authorities
- create endpoint to get all local authorities
- create service to format the response from the local authorities get request
- create tests for both the service and the route

#### Tests
- [ X] Yes

### DIFF
--- a/server/test/unit/routes/admin/local-authority-returns/local-authorities/index.spec.js
+++ b/server/test/unit/routes/admin/local-authority-returns/local-authorities/index.spec.js
@@ -1,0 +1,104 @@
+const expect = require('chai').expect;
+const httpMocks = require('node-mocks-http');
+const sinon = require('sinon');
+const models = require('../../../../../../models');
+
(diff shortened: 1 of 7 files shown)


server/test/unit/routes/admin/local-authority-returns/local-authorities/index.spec.js
```diff
@@ -0,0 +1,104 @@
+const expect = require('chai').expect;
+const httpMocks = require('node-mocks-http');
+const sinon = require('sinon');
+const models = require('../../../../../../models');
+
+const { getLocalAuthorities } = require('../../../../../../routes/admin/local-authority-return/monitor');
+
+describe.only('server/routes/admin/local-authority-returns/local-authorities', async () => {
```
joannafawl, line 8:
```suggestion
describe('server/routes/admin/local-authority-returns/local-authorities', async () => {
```

joannafawl (approved):

PR #3854, opened 2021-08-09
Fix: Find workplace address back link

joannafawl:
#### Work done
- Set back link on `find-workplace-address` component to confirm details url if you come from that page 

#### Tests
- [x] Yes


src/app/features/create-account/workplace/find-workplace-address/find-workplace-address.component.ts
```diff
@@ -59,4 +60,8 @@ export class FindWorkplaceAddressComponent extends FindWorkplaceAddress {
   protected onSuccess(data: LocationSearchResponse): void {
     this.registrationService.locationAddresses$.next(data.postcodedata);
   }
+
+  public setBackLinkToConfirmDetailsPage(): void {
```
duncanc19, line 64: Can have this set to protected as well?

duncanc19 (approved):

PR #3862, opened 2021-08-12
Fix/create account snagging list

duncanc19:
#### Work done
- Added css class to password show hide to stop button getting pushed onto second line in confirm details
- Removed renderForm in select main service to stop focus bug on error handling
- Changed wording in several components(_regulated-by-cqc, find-your-workplace, is-this-your-workplace, select-main-service, new-workplace-not-found_) for the parent flow and added tests to them

#### Tests
- [X] Yes

### DIFF
--- a/src/app/features/create-account/workplace/new-select-main-service/new-select-main-service.component.ts
+++ b/src/app/features/create-account/workplace/new-select-main-service/new-select-main-service.component.ts
@@ -1,11 +1,9 @@
 import { Component } from '@angular/core';
 import { FormBuilder } from '@angular/forms';
 import { ActivatedRoute, Router } from '@angular/router';
-import { Establishment } from '@core/model/establishment.model';
 import { Service } from '@core/model/services.model';
 import { BackService } from '@core/services/back.service';
 import { ErrorSummaryService } from '@core/services/error-summary.service';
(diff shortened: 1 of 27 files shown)


src/app/features/create-account/workplace/new-select-main-service/new-select-main-service.component.ts
```diff
@@ -40,6 +37,8 @@ export class NewSelectMainServiceComponent extends SelectMainServiceDirective {
   protected async init(): Promise<void> {
     this.flow = this.route.snapshot.parent.url[0].path;
     this.isRegulated = this.registrationService.isRegulated();
+    this.workplace = this.establishmentService.primaryWorkplace;
+    this.isParent = this.workplace?.isParent;
```
joannafawl, line 41: Do you need these if this is the registration flow?

duncanc19: I think it's good to set isParent in here to make it explicit instead of it being undefined, but I could just set it to false because we don't seem to use the workplace for anything else in this component

joannafawl: Yeah I think setting it to false would be better

joannafawl (approved):

PR #3863, opened 2021-08-12
Fixed other services input not showing when selecting an other option

aaron-russell:
#### Work done
- Due to us using `updateOn: 'submit'` the value is not updated until submitted, to combat this we have to get the JS value of the option to know if it's checked or not

#### Tests
- [x] Yes

### DIFF
--- a/src/app/shared/directives/create-workplace/select-main-service/select-main-service.directive.ts
+++ b/src/app/shared/directives/create-workplace/select-main-service/select-main-service.directive.ts
@@ -191,4 +191,8 @@ export class SelectMainServiceDirective implements OnInit, OnDestroy, AfterViewI
   ngOnDestroy() {
     this.subscriptions.unsubscribe();
   }
+
+  serviceSelected(id: string): boolean {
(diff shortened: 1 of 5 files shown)


src/app/shared/directives/create-workplace/select-main-service/select-main-service.directive.ts
```diff
@@ -191,4 +191,8 @@ export class SelectMainServiceDirective implements OnInit, OnDestroy, AfterViewI
   ngOnDestroy() {
     this.subscriptions.unsubscribe();
   }
+
+  serviceSelected(id: string): boolean {
+    return !(<HTMLInputElement>document.getElementById(id)).checked;
+  }
```
duncanc19, line 197: Isn't this returning true if the service is not selected? Because it returns true to add the hidden class, so should it be __serviceNotSelected__?

aaron-russell: Very true, I'll rename!

duncanc19 (approved):

PR #3868, opened 2021-08-13
Added fix rather than defaulting to 0 and added no option

aaron-russell:
#### Work done
- Added a No option to other job roles
- Added a fix to allow bulk upload to accept other job roles

#### Tests
- [x] Yes

### DIFF
--- a/server/models/BulkImport/csv/workers.js
+++ b/server/models/BulkImport/csv/workers.js
@@ -2009,7 +2009,7 @@ class Worker {
           source: `${this._currentLine.OTHERJOBROLE} - ${this._currentLine.OTHERJRDESC}`,
           column: 'OTHERJOBROLE/OTHERJRDESC',
         });
-      } else if (listOfOtherJobs.includes('0') && listOfJobsWithoutNo.length !== 0) {
+      } else if (listOfOtherJobs.includes('0') && listOfOtherJobs.length > 1) {
         this._validationErrors.push({
           lineNumber: this._lineNumber,
           errCode: Worker.OTHER_JOB_ROLE_ERROR,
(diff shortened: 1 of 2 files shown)


server/models/BulkImport/csv/workers.js
```diff
@@ -2008,6 +2009,17 @@ class Worker {
           source: `${this._currentLine.OTHERJOBROLE} - ${this._currentLine.OTHERJRDESC}`,
           column: 'OTHERJOBROLE/OTHERJRDESC',
         });
+      } else if (listOfOtherJobs.includes('0') && listOfJobsWithoutNo.length !== 0) {
```
joannafawl, line 2012: Are these two statements doing the same thing?

aaron-russell: We are looking to see if there's a 0 within the `listOfOtherJobs` this should be just a 0 by itself, but if there are other jobs aside from the 0, then it shouldn't hit this condition, that's why there's the second part. `listOfJobsWithoutNo` removed the 'No' option to see if there are any other jobs in there... If that makes any sense?

joannafawl: If `listOfOtherJobs = ['0', '14']`, would `['0', '14'].includes('0')` not return true?

aaron-russell: Yeah, but then it would go to the second condition `listOfJobsWithoutNo = ['14']` which would then hit the error

joannafawl: Ohh yeah I understand now!!

joannafawl (approved):

PR #3889, opened 2021-08-16
Fix/is this your workplace errors

Richard-Pentecost:
#### Work done
- add check in the prefill form function to check if the workplaceInterfaceService has a selected location
- if there was no location an error would occur in the console due to trying to access a property of null, and this caused the bug in the error messages

#### Tests
- [ ] Yes
- [ x] No, they are not required for this change

### DIFF
--- a/src/app/core/test-utils/MockConfigCatClient.ts
+++ b/src/app/core/test-utils/MockConfigCatClient.ts
@@ -49,7 +49,7 @@ export const mockConfigCatClient = {
     }
     if (flagName === 'createAccountNewDesign') {
       return new Promise((resolve) => {
-        return resolve(false);
+        return resolve(true);
       });
     }
     return new Promise((resolve) => {
--- a/src/app/core/test-utils/MockConfigCatClient.ts
+++ b/src/app/core/test-utils/MockConfigCatClient.ts
@@ -49,7 +49,7 @@ export const mockConfigCatClient = {
     }
     if (flagName === 'createAccountNewDesign') {
       return new Promise((resolve) => {
-        return resolve(true);
+        return resolve(false);
       });
     }
     return new Promise((resolve) => {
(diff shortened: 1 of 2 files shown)


aaron-russell (changes requested):

src/app/shared/directives/create-workplace/is-this-your-workplace/is-this-your-workplace.directive.ts
```diff
@@ -67,7 +67,10 @@ export class IsThisYourWorkplaceDirective implements OnInit, AfterViewInit {
   }
 
   protected prefillForm(): void {
-    if (this.workplaceInterfaceService.selectedLocationAddress$.value.locationId === this.locationData.locationId) {
+    if (
+      this.workplaceInterfaceService.selectedLocationAddress$.value &&
+      this.workplaceInterfaceService.selectedLocationAddress$.value.locationId === this.locationData.locationId
+    ) {
```
aaron-russell, line 73: You can use typescript to make this easier

```suggestion
    if (
      this.workplaceInterfaceService.selectedLocationAddress$.value?.locationId === this.locationData.locationId
    ) {
```

Richard-Pentecost: refactored in line with the above

src/app/core/test-utils/MockConfigCatClient.ts
```diff
@@ -49,7 +49,7 @@ export const mockConfigCatClient = {
     }
     if (flagName === 'createAccountNewDesign') {
       return new Promise((resolve) => {
-        return resolve(false);
```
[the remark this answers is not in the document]
Richard-Pentecost: changed back to have false as default

aaron-russell (approved):

PR #3919, opened 2021-08-24
Feat/changing primary user

duncanc19:
#### Work done
- Created change-primary-user page to replace pop up
- Changed call to pop-up in user-account-edit-permissions to navigate to new page
- Refactored MockUserService to generate readUsers instead of using the same one
- Deleted old change primary pop-up component

#### Tests
- [X] Yes


src/app/features/workplace/change-primary-user/change-primary-user.component.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    const props = {
+      role: selectedUser.role,
+      isPrimary: true,
+    };
+
+    this.subscriptions.add(
+      this.userService.updateUserDetails(this.workplaceUid, selectedUser.uid, { ...selectedUser, ...props }).subscribe(
+        (data) => {
+          this.router.navigate(['/workplace', this.workplaceUid, 'user', this.currentUserUid]);
```
Richard-Pentecost, line 87: could use the syntax `this.router.navigate([<route>] { relativeTo: this.route })` - so if `this.route = /workplace/:workplaceuid/user/:currentUserId/change-primary-user` and you wanted to go to route above, you should be able to just put `['../']` in place of `[<route>]`

Richard-Pentecost (approved):

PR #3964, opened 2021-09-08
Feat/pending registration

duncanc19:
#### Work done
- Created registration-request page to display single registration
- Created updateWorkplaceId endpoint to separate updating workplace ID from approving a registration
- Created getSingleRegistration endpoint in backend
- Created getEstablishmentWithPrimaryUser query in establishment model which is used by getSingleRegistration
- Moved getAllRegistrations out of index routes file

#### Tests
- [X] Yes


server/routes/admin/registrations/getAllRegistrations.js
```diff
@@ -0,0 +1,206 @@
...
+      },
+      order: [['id', 'DESC']],
+      include: [
+        {
+          model: models.services,
+          as: 'mainService',
+          attributes: ['id', 'name'],
+        },
+      ],
+    });
```
aaron-russell, line 92: Can we move these to the models? The new standard we have tried to go to is all db code lives with the models

server/routes/admin/registrations/getSingleRegistration.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+};
+
+const getParentEstablishmentId = async (parentId) => {
+  const parentEstablishmentData = await models.establishment.findOne({
+    where: {
+      id: parentId,
+    },
+    attributes: ['NmdsID'],
+  });
```
aaron-russell, line 29: Same with this, can we move it to the models?

server/routes/admin/registrations/index.js
```diff
@@ -1,210 +1,13 @@
...
-    } else {
-      res.status(200);
-    }
-  } catch (error) {
-    res.status(503);
-  }
-});
+router.route('/').get(getAllRegistrations);
+router.route('/:establishmentUid').get(getSingleRegistration);
+router.route('/updateWorkplaceId').post(updateWorkplaceId);
```
aaron-russell, line 11: This doesn't really match how we have other endpoints set up, can we move the routes into the folder with them and require them in the file above?

aaron-russell (changes requested):

server/routes/admin/registrations/getAllRegistrations.js
duncanc19: This was just moving the getAllRegistrations code out of the route, I was going to add refactoring it to the tech debt board because the endpoint makes several calls to the database and loops through all the data again unnecessarily.
I'd say it's not worth moving it to the model for now because we can change the sequelize part to do something similar to the singleRegistration one

aaron-russell: Alright, that's fine, fair enough. Yeah, please add it to tech debt though, don't want it getting lost.

aaron-russell (approved):

PR #4015, opened 2021-09-20
Fix: Multi update

joannafawl:
#### Work done
- Use `[(ngModel)]` in form for select all checkbox instead of reactive form so that validation is updated on submit
- Reset selected staff in training service when pressing cancel on `add-training-details` page

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/add-multiple-training/training-details/training-details.component.ts
+++ b/src/app/features/add-multiple-training/training-details/training-details.component.ts
@@ -80,4 +80,9 @@ export class MultipleTrainingDetailsComponent extends AddEditTrainingDirective i
   private onError(error) {
     console.log(error);
   }
+
+  public onCancel(): void {
+    this.trainingService.resetSelectedStaff();
+    this.router.navigate(this.previousUrl, { fragment: 'training-and-qualifications' });
+  }
 }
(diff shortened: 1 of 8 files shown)


src/app/features/add-multiple-training/training-details/training-details.component.ts
```diff
@@ -80,4 +80,9 @@ export class MultipleTrainingDetailsComponent extends AddEditTrainingDirective i
   private onError(error) {
     console.log(error);
```
Richard-Pentecost, line 81: Should this onError method do anything more with this error other than logging it

Richard-Pentecost (approved):

src/app/features/add-multiple-training/training-details/training-details.component.ts
joannafawl: It's a bit of a weird one, in other parts of the code we either display server error messages or just console log the error but we don't have any server error messages. We could navigate to the `Sorry there is a problem with this service` page too if you think that's better?

Richard-Pentecost: Yeah, I had something similar in one of my tickets. The issue with just console logging it is the user doesn't necessarily know that there has been a problem. Navigating to the error page maybe a good solution. Might be worth speaking to Aaron to see what he thinks

joannafawl: I checked with Aaron and we automatically navigate to that page on error

PR #4032, opened 2021-09-23
Feat: Long-term absence

joannafawl:
#### Work done
- Add `LongTermAbsence` column to worker table in database
- Add `longTermAbsence` to worker endpoint
- Create `long-term-absence` component
- Create `longTermAbsence` endpoint to get enum values from the database

#### Tests
- [x] Yes

### DIFF
--- a/server.js
+++ b/server.js
@@ -66,6 +66,7 @@ var approvals = require('./server/routes/approvals');
 var satisfactionSurvey = require('./server/routes/satisfactionSurvey');
 var registrationSurvey = require('./server/routes/registrationSurvey');
 var cqcStatusCheck = require('./server/routes/cqcStatusCheck');
+var longTermAbsence = require('./server/routes/longTermAbsence')
 
 // admin route
 var admin = require('./server/routes/admin');
@@ -244,6 +245,7 @@ app.use('/api/serviceUsers', [refCacheMiddleware.refcache, serviceUsers]);
 app.use('/api/trainingCategories', workingTrainingCategories);
 app.use('/api/nurseSpecialism', [refCacheMiddleware.refcache, nurseSpecialism]);
 app.use('/api/availableQualifications', [refCacheMiddleware.refcache, availableQualifications]);
+app.use('/api/longTermAbsence', require('./server/routes/longTermAbsence'));
 
 // transaction endpoints
 app.use('/api/errors', errors);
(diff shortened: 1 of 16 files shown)


server.js
```diff
@@ -66,6 +66,7 @@ var approvals = require('./server/routes/approvals');
 var satisfactionSurvey = require('./server/routes/satisfactionSurvey');
 var registrationSurvey = require('./server/routes/registrationSurvey');
 var cqcStatusCheck = require('./server/routes/cqcStatusCheck');
+var longTermAbsence = require('./server/routes/longTermAbsence')
```
aaron-russell, line 69: I really don't like the use of var, but I know this file has that.

src/app/features/workers/long-term-absence/long-term-absence.component.spec.ts
```diff
@@ -0,0 +1,258 @@
...
+  it('should reset radio buttons when pressing back at work checkbox', async () => {
+    const { component, fixture, getByText } = await setup();
+
+    component.worker.longTermAbsence = 'Illness';
+    fixture.detectChanges();
+
+    const backAtWorkCheckbox = getByText('Set as back at work');
+    fireEvent.click(backAtWorkCheckbox);
+
+    expect(component.form.value.longTermAbsence).toBe(null);
```
aaron-russell, line 121: I would also check the value earlier on to make sure it turns null. If that makes sense?

src/app/features/workers/long-term-absence/long-term-absence.component.ts
```diff
@@ -0,0 +1,153 @@
...
+    this.workerService.setReturnTo(null);
+  }
+
+  private getReasonsForLongTermAbsence(): void {
+    this.subscriptions.add(
+      this.workerService.getLongTermAbsenceReasons().subscribe((reasons) => {
+        this.longTermAbsenceReasons = reasons;
+      }),
+    );
+  }
```
aaron-russell, line 63: Could this be in a resolver?

aaron-russell (changes requested):

server.js
joannafawl: Do you think it's better to change this one to `const` or keep it consistent? Or change them all to `const`?

aaron-russell: I debate this every time, and I end up doing exactly as you've done, keeping it consitent

aaron-russell (approved):

PR #4034, opened 2021-09-23
Bug/remove skip test

Richard-Pentecost:
#### Work done
- remove .only from tests that were merged in with the previous branch

#### Tests
- [ ] Yes


migrations/20210920105716-registrationNotesTable.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable(
+      'RegistrationNotes',
```
aaron-russell, line 6: Didn't we say it was just going to be notes, shared across all places? Then it would have a column for when it was added (e.g. reg)

src/app/core/model/registrations.model.ts
```diff
@@ -43,6 +43,11 @@ export interface UpdateRegistrationStatusRequest {
   inReview: boolean;
 }
 
+export interface Note {
+  createdAt: Date;
+  note: string;
+  user: { FullNameValue: string };
+}
```
aaron-russell, line 50: Missing a new line, looks like that's the norm?

src/app/core/resolvers/admin/registration-requests/single-registration/get-registration-notes.resolver.ts
```diff
@@ -0,0 +1,23 @@
...
+export class GetRegistrationNotesResolver implements Resolve<any> {
+  constructor(private router: Router, private registrationsService: RegistrationsService) {}
+
+  resolve(route: ActivatedRouteSnapshot): Observable<Note[]> {
+    const lastUrlSegmentIndex = route.url.length - 1;
+    const establishmentUid = route.url[lastUrlSegmentIndex].path;
+
+    return this.registrationsService.getRegistrationNotes(establishmentUid).pipe(
+      catchError(() => {
+        this.router.navigate(['/sfcadmin']);
```
aaron-russell, line 18: Should this go to `There's a problem with the service?`

aaron-russell (approved):

PR #4039, opened 2021-09-24
Bugfix/pending parent request missing from home tab

Richard-Pentecost:
#### Work done
- create endpoint for deleting a becomeAParent request
- add cancel becomeAParent dialog box
- change the way the links are shown so the correct links are shown at the right time

#### Tests
- [x] Yes


server/models/approvals.js
```diff
@@ -200,5 +200,15 @@ module.exports = (sequelize, DataTypes) => {
...
+    const deletedRequest = await this.destroy({
+      where: {
+        EstablishmentID: establishmentId,
+      },
+    });
+
+    return deletedRequest;
+  }
+
   return Approvals;
```
duncanc19, line 213: Could call it just deleteParentRequest and just return instead of setting it to the deletedRequest variable

server/routes/approvals/becomeAParent.js
```diff
@@ -39,16 +39,59 @@ const validateBecomeAParentRequest = async (req, res, next) => {
...
+
+};
+
 const becomeAParentEndpoint = async (req, res) => {
   const becomeAParentRequest = await models.Approvals.createBecomeAParentRequest(req.userId, req.establishment.id);
-
   res.send(becomeAParentRequest);
 };
 
+const deleteAParentEndpoint = async (req, res) => {
```
duncanc19, line 76: Maybe deleteParentRequest? It makes it sound like it's deleting a parent account without request and I'm not a big fan of endpoint being in the names

src/app/core/test-utils/MockConfigCatClient.ts
```diff
@@ -49,7 +49,7 @@ export const mockConfigCatClient = {
     }
     if (flagName === 'createAccountNewDesign') {
       return new Promise((resolve) => {
-        return resolve(false);
+        return resolve(true);
```
duncanc19, line 52: I'd forgotten about this, do you want to remove it from here and we'll talk about it in tech time like we said we'd do last time?

src/app/features/dashboard/home-tab/home-tab.component.html
```diff
@@ -158,5 +163,8 @@ <h3 class="govuk-heading-s govuk-!-margin-top-1">Download reports</h3>
         </ul>
       </div>
     </ng-container>
+    <!-- <ng-container *ngIf="!(canBecomeAParent && canLinkToParent) && !workplace.isParent">
```
duncanc19, line 166: Does this need to go back in or can you remove it?

duncanc19 (approved):

PR #4068, opened 2021-09-29
Feat/rejected registrations list

Richard-Pentecost:
#### Work done
- put pending and rejected registration links behind separate urls
- create rejected request table
- refactor backend getRegistrations endpoint
- create getRejectedRegistrationsResolver
- rename getRegistrationsResolver to getPendingRegistrationsResolver
- create individual page for a rejected request
- tests for all the above

#### Tests
- [x] Yes


server/routes/admin/approval/index.js
```diff
@@ -97,13 +97,19 @@ const _approveNewUser = async (login, workplace, nmdsId, res) => {
   }
 };
 
-const _rejectNewUser = async (user, workplace, res) => {
+const _rejectNewUser = async (user, workplace, req, res) => {
   try {
     if (user && workplace) {
       const deletedUser = await user.destroy();
 
+      const { FullNameValue } = await models.user.findByUUID(req.userUid);
```
duncanc19, line 105: Instead of doing a separate call to the model, you could add the full name to the attributes returned in findByLoginId, then it would already be available in user

src/app/features/admin/registration-requests/rejected-registration-request/rejected-registration-request.component.html
```diff
@@ -0,0 +1,70 @@
...
+          {{ registration.establishment.isRegulated ? 'Yes' : 'No' }}
+        </dd>
+      </div>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">Main service</dt>
+        <dd class="govuk-summary-list__value">
+          {{ registration.establishment.mainService }}
+        </dd>
+      </div>
+    </dl>
```
duncanc19, line 60: I think a lot of this is duplicated in the pending registration request page, because the table is the same in both isn't it?

src/app/features/admin/registration-requests/rejected-registration-request/rejected-registration-request.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.registrationsService.getRegistrationNotes(this.registration.establishment.uid).subscribe(
+      (data) => {
+        this.notes = data;
+      },
+      (error) => {
+        this.notesError = 'There was an error retrieving notes for this registration';
+      },
+    );
+  }
+}
```
duncanc19, line 88: And same here, a lot of this is in the pending registration request page so we should be able to use it in both, either by breaking the common stuff into separate components or by using a directive

server/routes/admin/approval/index.js
Richard-Pentecost: This doesn't work because the purpose of this call is get the user that is approving/rejecting not the user who is trying to register

src/app/features/admin/registration-requests/rejected-registration-request/rejected-registration-request.component.html
Richard-Pentecost: Yeah, spend ages this afternoon trying to create a directive for both pages but is actually quite complicated due to having get calls. Think this should go into a separate tech debt ticket if we think it is worth it

src/app/features/admin/registration-requests/rejected-registration-request/rejected-registration-request.component.ts
Richard-Pentecost: See above

duncanc19 (approved):

PR #4158, opened 2021-10-19
Feat/remove new from select-main-service components

duncanc19:
#### Work done
- Removed new from select-main-service components

#### Tests
- [ ] Yes


src/app/features/create-account/user/your-details/your-details.component.spec.ts
```diff
@@ -285,19 +285,19 @@ describe('YourDetailsComponent', () => {
...
       component.fixture.detectChanges();
 
       expect(backLinkSpy).toHaveBeenCalledWith({
-        url: ['registration', 'new-select-main-service'],
+        url: ['registration', 'select-main-service'],
       });
     });
 
-    it('should set the back link to new-select-main-service if return url is not null', async () => {
+    it('should set the back link to confirm-details if return url is not null', async () => {
```
aaron-russell, line 300: Is this correct? As you've changed the destination?

PR #4256, opened 2021-11-05
Feat/parent training report training tab

Zuhal-Ayob:
#### Work done
- Add parent training tab to the parent training report
- Add parent training table to the training tab
- Add the content of the table and add styling for the table

#### Tests
- [x] Yes


aaron-russell (dismissed):

server/utils/trainingAndQualificationsUtils.js
```diff
@@ -1,3 +1,6 @@
+const establishment = require("../models/establishment");
+const worker = require("../models/worker");
+
```
joannafawl, line 3: According to LGTM these aren't being used so they should be safe to delete

joannafawl (approved):

PR #4265, opened 2021-11-08
Tech debt: Added DayJS and migrated moment to DayJS on the front end

aaron-russell:
#### Work done
- Migrated to dayJS
- Removed moment from the front end

#### Tests
- [ ] Yes


duncanc19 (approved):

src/app/shared/validators/date.validator.ts
```diff
@@ -138,14 +226,26 @@ export abstract class DateValidator {
...
+
+        if (
+          this.validate(
+            `${FormatUtil.formatSingleDigit(year.value)}-${FormatUtil.formatSingleDigit(
+              month.value,
+            )}-${FormatUtil.formatSingleDigit(day.value)}`,
+            DATE_PARSE_FORMAT,
+          )
+        ) {
           return date.isAfter(min) ? null : { dateMin: { min, actual: date } };
```
duncanc19, line 249: Could you put the last part into a variable, then do the validate in an if block

`
const formattedDateString = `${FormatUtil.formatSingleDigit(year.value)}-${FormatUtil.formatSingleDigit(
            month.value,
          )}-${FormatUtil.formatSingleDigit(day.value)}`;

        if (
          this.validate(
            formattedDateString,
            DATE_PARSE_FORMAT,
          ) {
            const date = dayjs(formattedDateString, DATE_PARSE_FORMAT)
          }
        )`

duncanc19: The back ticks mess up the code formatting
I'm wondering if it'd be worth putting that formatting into a function as well because it's used quite a few times

PR #4287, opened 2021-11-15
Feat/sharing permissions questions

Zuhal-Ayob:
#### Work done
- Add radio buttons to the data sharing question page 
- changed shared with property to be boolean instead of array
- deleted shareDataValue column in establishment table
- updated areas with old data-sharing model

#### Tests
- [x] Yes

### DIFF
--- a/src/app/core/test-utils/MockEstablishmentService.ts
+++ b/src/app/core/test-utils/MockEstablishmentService.ts
@@ -8,20 +8,20 @@ import { Observable, of } from 'rxjs';
 
 @Injectable()
 export class MockEstablishmentService extends EstablishmentService {
-  private share: any = { enabled: false, with: [] };
+  private shareWith: any = { cqc: null, localAuthorities: null };
 
-  public static factory(newShare: any) {
+  public static factory(shareWith: any) {
     return (http: HttpClient) => {
       const service = new MockEstablishmentService(http);
-      if (newShare) {
-        service.setShare(newShare);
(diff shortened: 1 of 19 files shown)


src/app/core/test-utils/MockEstablishmentService.ts
```diff
@@ -102,9 +102,9 @@ export class MockEstablishmentService extends EstablishmentService {
       postcode: '',
       primaryAuthority: undefined,
       serviceUsers: [],
-      share: {
-        enabled: false,
-        with: [],
+      shareWith: {
+        cqc: null,
+        localAuthorities: null,
```
joannafawl, line 107: Could this be `shareWith: this.shareWith` to keep it consistent with the one above?

src/app/features/workplace/data-sharing/data-sharing.component.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      expect(updateDataSharingSpy).toHaveBeenCalledWith(component.establishment.uid, {
+        shareWith: {
+          cqc: null,
+          localAuthorities: false,
+        },
+      });
+    });
+  });
+
+  describe('Passing data to updateDataSharing in establishment service', async () => {
```
joannafawl, line 112: This describe block is the same name as the one above, maybe put them in the same one or rename so it's clear this one is for CQC?

joannafawl (approved):

PR #4288, opened 2021-11-15
Feat: Qualifications delete button

joannafawl:
#### Work done
- Create `QualificationResolver`
- Change heading titles and sizes for consistency on `add-edit-qualification` and `select-record-type` components
- Adapt `delete-record` component for deleting a qualification record

#### Tests
- [x] Yes


src/app/features/workers/add-edit-qualification/add-edit-qualification.component.html
```diff
@@ -54,8 +63,24 @@ <h1 class="govuk-fieldset__heading">
...
+      type="button"
+      class="govuk-button govuk-button--warning govuk-!-margin-right-8"
+      [routerLink]="[
+        '/workplace',
+        workplace.uid,
+        'training-and-qualifications-record',
+        worker.uid,
+        'qualification',
+        qualificationId,
+        'delete'
```
duncanc19, line 80: Is this route similar to the current one or is it quite different? If it's just adding to the end or changing the last one it might be better to use ../ because it's quite long?

duncanc19 (approved):

PR #4322, opened 2021-11-23
Feat/sharing permissions save and return

Richard-Pentecost:
#### Work done
- create new endpoint for updating the show permission banner in the db
- add functionality to the data-sharing component so that on 'save and return' and 'exit' the banner is removed
- change back link routing to take you back to the previous page
- write tests

#### Tests
- [x] Yes


server/models/classes/establishment.js
```diff
@@ -339,9 +343,9 @@ class Establishment extends EntityValidator {
     return this._eightWeeksFromFirstLogin;
   }
 
-  get showSharingPermissionsBanner() {
-    return this._showSharingPermissionsBanner;
-  }
+  // get showSharingPermissionsBanner() {
+  //   return this._showSharingPermissionsBanner;
+  // }
```
joannafawl, line 348: Do you still need this?

src/app/features/workplace/data-sharing/data-sharing.component.ts
```diff
@@ -63,12 +63,20 @@ export class DataSharingComponent extends Question {
...
+    const completeFunction = () => {
+      this.subscriptions.add(
+        this.establishmentService.updateDataSharing(this.establishment.uid, props).subscribe(
+          (data) => {
+            this._onSuccess(data);
+          },
+          (error) => this.onError(error),
+        ),
+      );
+    };
```
joannafawl, line 75: Could this function be defined outside of `updateEstablishment` and just called from within?

joannafawl (approved):

PR #4337, opened 2021-11-29
Feat/admin search main page

Richard-Pentecost:
#### Work done
- create links to change between different search pages
- create the search for workplace component and put initial text and form
- create search for user and search for group component with dummy text
- create tests

#### Tests
- [x] Yes


src/app/features/admin/search/search.component.scss
```diff
@@ -0,0 +1,30 @@
...
+      &:hover {
+        color: $govuk-link-hover-colour;
+      }
+    }
+
+    &:not(:first-of-type) {
+      text-align: center;
+    }
+  }
+}
```
duncanc19, line 30: Is this CSS similar to anything else we've got like the registrations page? If it's used in a few of the admin pages it might be worth moving it into a shared css file?

duncanc19 (approved):

PR #4347, opened 2021-12-02
Feat/admin search groups

duncanc19:
#### Work done
- Created search for group page in new admin section
- Created workplace-dropdown component used in search for group and to be used by search for workplace

#### Tests
- [X] Yes


src/app/features/admin/search/search-for-group/search-for-group.component.spec.ts
```diff
@@ -1,24 +1,265 @@
...
+
+      it('should collapse the Workplace details when clicking Close', async () => {
+        const { getByTestId } = await setup(true);
+
+        const searchResults = within(getByTestId('group-search-results'));
+
+        fireEvent.click(searchResults.getByText('Open'));
+        fireEvent.click(searchResults.getByText('Close'));
+
+        expect(searchResults.queryByTestId('groups-workplace-details')).toBeNull();
```
joannafawl, line 201: Would it be useful to check it's not null when you first click open?

src/app/features/admin/search/search-for-group/search-for-group.component.spec.ts
```diff
@@ -1,24 +1,265 @@
...
+        const { fixture, searchGroupsSpy, getByTestId } = await setup();
+
+        searchGroupsSpy.and.returnValue(of([]));
+
+        const searchButton = getByTestId('searchButton');
+        fireEvent.click(searchButton);
+
+        fixture.detectChanges();
+
+        expect(getByTestId('no-search-results'));
```
joannafawl, line 242: Could you test the no search results text?

src/app/features/admin/search/search-for-group/search-for-group.component.spec.ts
duncanc19: The previous test checks that the content is there when it's open so I don't think we need it here too. I could change this to check for the text instead though to make the link between the tests clearer

src/app/features/admin/search/search-for-group/search-for-group.component.spec.ts
duncanc19: Sure, I've just updated it

joannafawl (approved):

PR #4425, opened 2021-12-22
Feat/remove parent status

Richard-Pentecost:
#### Work done
- create search directive and refactor the three search components to use the directive
- alter the search establishment endpoint to return subsidiaries
- add link to the frontend
- create new dialog component
- create new endpoint to change the isParent status 
- write tests

#### Tests
- [x] Yes


src/app/features/admin/search/search-for-user/search-for-user.component.ts
```diff
@@ -48,36 +45,8 @@ export class SearchForUserComponent implements OnInit, OnDestroy {
...
-      removeUnlock: () => {
-        this.results[index].isLocked = false;
-      },
-    };
-    this.dialogService.open(AdminUnlockConfirmationDialogComponent, data);
-  }
-
-  ngOnDestroy(): void {
-    this.subscriptions.unsubscribe();
+    this.unlockUser(username, index, this.results[index]);
```
joannafawl, line 50: Could you call `unlockUser` directly in the html instead of calling it inside `unlockWorkplaceUser`?

src/app/shared/components/remove-parent-confirmation/remove-parent-confirmation.component.ts
```diff
@@ -0,0 +1,87 @@
...
+        message: 'Send request to unlock service not found. You can try again or contact us.',
+      },
+    ];
+  }
+
+  /**
+   * Function is used to close dialog window after successful confirmation
+   * @param {any} true to close dialog after response or null to close without action
+   * @return {void}
+   */
```
joannafawl, line 57: I think it would be good to get rid of some of these comments in this file

src/app/shared/components/remove-parent-confirmation/remove-parent-confirmation.component.ts
```diff
@@ -0,0 +1,87 @@
...
+   * @param {string}
+   * @return {void}
+   */
+  public removeParentStatus(data: any): void {
+    this.establishmentService.removeParentStatus(data).subscribe(
+      () => {
+        data.removeStatus();
+        this.alertService.addAlert({
+          type: 'success',
+          message: 'Parent status has been removed,',
```
joannafawl, line 73:
```suggestion
          message: 'Parent status has been removed',
```
Maybe get rid of this comma too?

src/app/features/admin/search/search-for-user/search-for-user.component.ts
Richard-Pentecost: It was originally called like you've suggested, however all search components need it but search user doesn't use the dropdown. So I've pulled it into the directive, so that it is only in one place that is common to all components.

src/app/shared/components/remove-parent-confirmation/remove-parent-confirmation.component.ts
Richard-Pentecost: Removed the comments

src/app/shared/components/remove-parent-confirmation/remove-parent-confirmation.component.ts
Richard-Pentecost: Yeah that should of been a full stop

joannafawl (approved):

PR #4430, opened 2021-12-23
Admin parent request approval buttons

popey2700:
#### Work done
- Added approve/reject buttons to parent request page
- Added appropriate pop-ups and alerts

#### Tests
- [ x] Yes


src/app/features/admin/parent-requests/parent-request-individual/parent-request-individual.component.html
```diff
@@ -45,7 +45,7 @@ <h2 class="govuk-heading-m">Workplace details</h2>
         </dd>
       </div>
       <div class="govuk-summary-list__row">
-        <dt class="govuk-summary-list__key">Username</dt>
+        <dt class="govuk-summary-list__key">Name</dt>
```
Zuhal-Ayob, line 48: I think this is a Username in the prototype

joannafawl: The full name is being brought back so we checked with Jackie and she said to change it on the front end

Zuhal-Ayob (approved):

PR #4456, opened 2022-01-11
Feat/bulk upload troubleshooting page

popey2700:
#### Work done
- Created Troubleshooting page with related content sidebar
- Created BULK_UPLOAD_HELP breadcrumb journey
- Created CMS Troubleshooting page in pages collections

#### Tests
- [x] Yes


src/app/core/breadcrumb/journey.bulk-upload.ts
```diff
@@ -56,3 +57,32 @@ export const bulkUploadJourney: JourneyRoute = {
...
+            {
+              title: 'Troubleshooting',
+              path: Path.TROUBLESHOOTING_PAGE,
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
```
duncanc19, line 88: Could you add the troubleshooting path to the children of GET_HELP in the bulkUploadJourney above? It looks like the Troubleshooting is the only new page so we shouldn't need the paths for the other pages again

duncanc19 (approved):

PR #4469, opened 2022-01-17
Feat/main bulk upload page

Zuhal-Ayob:
#### Work done
- Create migration to add `Data Change Last Updated` column
- Create a service to get `Data Changes` page data from CMS
- Add `Data Changes` page and set routing 
- Add endpoint to and resolver to update `Data Change Last Updated` in the database
- Add function to update data in the database 
- Add endpoint and resolver to get data of  `Data Change Last Updated` from the database
- Add  a function to compare get data and set data and show the flag

#### Tests
- [x] Yes


src/app/core/resolvers/data-change.resolver.ts
```diff
@@ -0,0 +1,20 @@
...
+import { DataChange } from '@core/model/data-change.model';
+import { DataChangeService } from '@core/services/data-change.service';
+import { Observable, of } from 'rxjs';
+import { catchError, take } from 'rxjs/operators';
+
+@Injectable()
+export class DataChangeResolver implements Resolve<any> {
+  constructor(private router: Router, private dataChangeService: DataChangeService) {}
+
+  resolve(route: ActivatedRouteSnapshot): Observable<null | DataChange[]> {
```
Richard-Pentecost, line 12: Is it definitely `DataChange[]`. Feel like the return would look like `{ dataChangesLastUpdate: '01 Feb 2021'}`. If this is the case, it maybe worth making a new model and referencing that

src/app/core/resolvers/data-changes-lastupdated.resolver.ts
```diff
@@ -0,0 +1,21 @@
...
+@Injectable()
+export class DataChangeLastUpdatedResolver implements Resolve<any> {
+  constructor(private dataChangeService: DataChangeService, private establishmentService: EstablishmentService) {}
+
+  resolve(route: ActivatedRouteSnapshot) {
+    const establishmentId = this.establishmentService.establishmentId;
+    return this.dataChangeService.getDataChangesLastUpdate(establishmentId).pipe(
+      take(1),
+      catchError(() => {
+        return of(null);
```
Richard-Pentecost, line 17: Check with Aaron, but I'm thinking these should error to go the `Problem with a service `page`. This is something that I've missed on some of my resolvers, but have just realised

src/app/core/resolvers/data-change.resolver.ts
```diff
@@ -0,0 +1,20 @@
...
+    return this.dataChangeService.getDataChange().pipe(
+      take(1),
+      catchError(() => {
+        return of(null);
```
Richard-Pentecost, line 16: Same as comment for `data-changes-lastupdate.resolver`

src/app/core/services/data-change.service.ts
```diff
@@ -0,0 +1,31 @@
...
+    params = params.set('fields', 'content,title,last_updated');
+
+    return this.http.get<DataChange>(`${environment.cmsUri}/items/${this.path}`, { params });
+  }
+
+  public updateBUDataChangeLastUpdated(establishmentId: string, lastUpdated: Date) {
+    return this.http.post<any>(`/api/establishment/${establishmentId}/bulkUpload/dataChange`, { lastUpdated });
+  }
+
+  public getDataChangesLastUpdate(establishmentId: string): Observable<any> {
```
Richard-Pentecost, line 28: May be worth making a model for the return and using that in place of `<any>`. This could also be used in the updateBUDataChangeLastUpdated

src/app/features/bulk-upload/data-changes/data-change.component.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.breadcrumbService.show(JourneyType.BULK_UPLOAD);
+  }
+
+  public updateLastUpdatedDataChangeDate() {
+    const lastUpdated = this.datachange.last_updated;
+
+    this.subscriptions.add(
+      this.dataChangeService
+        .updateBUDataChangeLastUpdated(this.workplaceUid, lastUpdated)
+        .subscribe((err) => console.log(err)),
```
Richard-Pentecost, line 37: Do we need to do anything with an error, ie go to `problem with a service page` or show a banner

src/app/features/bulk-upload/help-area/bulk-upload-help-main-page.component.spec.ts
```diff
@@ -80,13 +85,16 @@ describe('BulkUploadHelpMainPageComponent', () => {
...
   it('should render the step by step link with the correct href attribute', async () => {
     const { getByText } = await setup();
+    expect(getByText('Step by step bulk upload guide')).toBeTruthy();
     const link = getByText('Step by step bulk upload guide');
     expect(link.getAttribute('href')).toContain('step-by-step-guide');
   });
 
   it('should render the troubleshooting link with the correct href attribute', async () => {
     const { getByText } = await setup();
     const link = getByText('Get handy troubleshooting hints to help you fix common problems and errors');
```
Richard-Pentecost, line 95: use `queryByText`, otherwise if the link doesn't exist the tests will break

Richard-Pentecost (commented):
A couple of little things

src/app/core/resolvers/data-changes-lastupdated.resolver.ts
aaron-russell: This doesn;t stop the page from working, so wouldn't go to `Problem with the service` in this case.

src/app/features/bulk-upload/help-area/bulk-upload-help-main-page.component.spec.ts
Richard-Pentecost: Sorry Zuhal, you've changed the wrong one. Line 88 was fine with a `getByText` because it was in an `expect` statement. It's the one on line 95, should be `const link = queryByText('....')`. I would set both the test on line 86 and the test line 93 the same -`const link = queryByText('....')` then do the 2 expect statements using the link (i.e. `expect(link).toBeTruth() and expect(link.getAttribute('href').toContain(....)`

Richard-Pentecost (commented):
just a comment on the test

src/app/features/bulk-upload/help-area/bulk-upload-help-main-page.component.spec.ts
Zuhal-Ayob: Sorry, I hope it's fixed now

PR #4476, opened 2022-01-19
Tech debt/refactor permissions

duncanc19:
#### Work done
- Created new version of permissions to make it more extensible for additional future roles
- Changed permissions to be passed around the service as an array of permissions, instead of an object with permission keys and true/false values
- Changed the way admin permissions work, so admins have a set of permissions instead of automatically having permission to do everything

#### Tests
- [X] Yes


server/utils/security/permissions.js
```diff
@@ -0,0 +1,188 @@
...
+  'canEditWorker',
+  'canDeleteWorker',
+  'canTransferWorker',
+  'canEditEstablishment',
+  'canRunLocalAuthorityReport',
+  ...getAdditionalEditPermissions(estabType, establishmentInfo, isLoggedInAsParent),
+];
+
+const adminPermissions = (estabType = 'Standalone', establishmentInfo, isLoggedInAsParent) => {
+  const adminCant = ['canViewNinoDob'];
```
Zuhal-Ayob, line 73: which kind of permisson is this?

Zuhal-Ayob (approved):

server/utils/security/permissions.js
aaron-russell: It's so the admins can't view sensitive information about workers

PR #4496, opened 2022-01-21
Feat/login wizard video

popey2700:
#### Work done
- Created a video field in wizard CMS
- Added logic into HTML to determine whether to show an image or video
- Created a Pipe to allow the vimeoUrl to be used

#### Tests
- [x] Yes


src/app/shared/pipes/first-login-wizard.pipe.ts
```diff
@@ -0,0 +1,10 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { DomSanitizer } from '@angular/platform-browser';
+
+@Pipe({ name: 'sanitizeVideoUrl' })
+export class FirstLoginWizardPipe implements PipeTransform {
+  constructor(private sanitizer: DomSanitizer) {}
+  transform(rawVideoUrl) {
+    return this.sanitizer.bypassSecurityTrustResourceUrl(rawVideoUrl);
+  }
+}
```
duncanc19, line 10: Can you change the file and class names to match the pipe name?

src/app/features/first-login-wizard/first-login-wizard.component.ts
```diff
@@ -14,12 +15,14 @@ export class FirstLoginWizardComponent {
   public isLast: boolean;
   public currentIndex: number;
   public imageUrl: string;
-  constructor(private route: ActivatedRoute) {}
+  public rawVideoUrl: string;
+  constructor(private route: ActivatedRoute, private domSanitizer: DomSanitizer) {}
```
duncanc19, line 20: You're not using the DomSanitizer in the component because you moved it into the pipe, so you can remove it from the constructor and the import. Also can you add a line between the variables and the constructor?

src/app/features/first-login-wizard/first-login-wizard.component.spec.ts
```diff
@@ -75,6 +76,24 @@ describe('FirstLoginWizardComponent', () => {
...
+  });
+
+  it('should render a image if there is no video url in the cms', async () => {
+    const { fixture, getByAltText } = await setup();
+    fixture.detectChanges();
+    fixture.whenStable().then(() => {
+      expect(getByAltText(wizard.data[1].video)).toBeFalsy();
+      expect(getByAltText(wizard.data[1].image)).toBeTruthy();
+    });
+  });
```
duncanc19, line 95: You shouldn't need detectChanges and whenStable because the onInit isn't async

duncanc19 (approved):

PR #4510, opened 2022-01-25
Removed training and qualifications feature flag and old trainingReport folder

popey2700:
#### Work done
- Removed feature flag from appropriate services
- Removed old trainingReport folder

#### Tests
- [ ] Yes


src/app/shared/components/training-link-panel/training-link-panel.component.ts
```diff
@@ -42,18 +38,6 @@ export class TrainingLinkPanelComponent implements OnInit, OnDestroy, OnChanges
...
-      .then((value) => {
-        this.newTrainingAndQualificationsReport = value;
-      });
-
-    this.featureFlagsService.configCatClient
-      .getValueAsync('parentTrainingAndQualificationsReport', false)
-      .then((value) => {
-        this.parentTrainingAndQualificationsReport = value;
-      });
   }
```
duncanc19, line 41: This ngOnInit doesn't need to be async (I don't think it ever needed to be because it was using .then anyway)

duncanc19 (approved):

PR #4542, opened 2022-01-28
Tech debt/bu refactor training lambda

duncanc19:
#### Work done
- Created validateTraining lambda
- Moved trainingCSVValidator into the lambdas/bulkUpload folder
- Deleted validatorFactory as only one validator class still in server section of codebase 

#### Tests
- [ ] Yes


lambdas/bulkUpload/classes/BUDI.js
```diff
@@ -67,6 +67,10 @@ class BUDI {
   qualifications(direction, originalCode) {
     return this.convertValue(direction, originalCode, this.mappings.QUALIFICATIONS);
   }
+
+  trainingCategory(direction, originalCode) {
+    return this.convertValue(direction, originalCode, this.mappings.TRAINING_CATEGORY);
```
Zuhal-Ayob, line 72: Could we have a specific return type ?

lambdas/bulkUpload/classes/trainingCSVValidator.js
```diff
@@ -310,7 +314,7 @@ class TrainingCsvValidator {
   _validateCategory() {
     const myCategory = parseInt(this._currentLine.CATEGORY, 10);
 
-    if (Number.isNaN(myCategory) || BUDI.trainingCategory(BUDI.TO_ASC, myCategory) === null) {
+    if (Number.isNaN(myCategory) || this.BUDI.trainingCategory(this.BUDI.TO_ASC, myCategory) === null) {
```
Zuhal-Ayob, line 317: do we need that ===null ?

lambdas/bulkUpload/test/unit/classes/trainingCSVValidator.spec.js
```diff
@@ -1,22 +1,28 @@
...
-      });
+      const validator = new TrainingCsvValidator(
+        {
+          LOCALESTID: 'foo',
+          UNIQUEWORKERID: 'bar',
+          CATEGORY: 1,
+          DESCRIPTION: 'training',
+          DATECOMPLETED: '',
+          EXPIRYDATE: '',
+          ACCREDITED: '',
```
Zuhal-Ayob, line 20: i think we can remove ACCREDITED from here when we are checking it shouldn’t be there, here its checking if its empty .

Zuhal-Ayob (approved):

lambdas/bulkUpload/classes/BUDI.js
duncanc19: This was backend code where we're not using Typescript, so the code moved into the lambdas folder is also just standard JS with no types

lambdas/bulkUpload/classes/trainingCSVValidator.js
duncanc19: The === null is there to check if null's returned from trainingCategory instead of a code, I think there is a lot of refactoring we could do in this file but it was a bit out of scope of this ticket because it doesn't affect putting it in the lambda

lambdas/bulkUpload/test/unit/classes/trainingCSVValidator.spec.js
duncanc19: This one is just passing in an object to set up an instance of the validator class

PR #4550, opened 2022-02-01
Tech debt/dashboard resolvers

duncanc19:
#### Work done
- Created allUsersForEstablishment and totalStaffRecords resolvers, updated workers resolver to use in both dashboard and view-workplace components
- Refactored SharingPermissionsBanner to use field in establishment instead of setting in the establishment service
- Updated user-accounts-summary to use allUsersForEstablishment resolver

#### Tests
- [X] Yes

#### Note
- Had updated how canViewBenchmarks is set in benchmarks-tab/dashboard/view-workplace to use can, however this caused issue with admin permissions, as admins automatically get permission with can but shouldn't be able to see benchmarks in workplace which doesn't have access. This issue will be resolved when the permissions refactor ticket goes through


src/app/core/resolvers/dashboard/all-users-for-establishment.resolver.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  resolve(route: ActivatedRouteSnapshot): Observable<Array<UserDetails> | null> {
+    const workplaceUid = route.paramMap.get('establishmentuid')
+      ? route.paramMap.get('establishmentuid')
+      : this.establishmentService.establishmentId;
+
+    return this.userService.getAllUsersForEstablishment(workplaceUid).pipe(
+      catchError(() => {
+        return of(null);
+      }),
```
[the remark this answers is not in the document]
duncanc19: Yeah, I was wondering about this, there wasn't any error handling in the subscriptions and I wasn't sure if it was best to do a redirect. Some of the resolvers are for things which aren't completely essential to the service so if an error is thrown on an endpoint would you want it to stop you from accessing the service at all?

src/app/features/dashboard/dashboard.component.ts
```diff
@@ -84,44 +85,23 @@ export class DashboardComponent implements OnInit, OnDestroy {
...
-          if (total) {
-            this.totalStaffRecords = total;
-          }
-        },
-        (error) => {
-          console.error(error.error);
-        },
-      ),
-    );
+    this.totalStaffRecords = this.route.snapshot.data.totalStaffRecords;
```
Richard-Pentecost, line 88: Can this not be moved in to the ngOnInit()? Seems unnecessary to have a single line function

src/app/features/workplace/view-workplace/view-workplace.component.ts
```diff
@@ -187,7 +178,11 @@ export class ViewWorkplaceComponent implements OnInit, OnDestroy {
     }
   }
 
-  ngOnDestroy() {
+  private getTotalStaffRecords(): void {
```
Richard-Pentecost, line 181: Like above, can this just go as a one line in ngOnInit()

Richard-Pentecost (commented):
Just a couple of comments

src/app/core/resolvers/dashboard/total-staff-records.resolver.ts
```diff
@@ -0,0 +1,24 @@
...
+export class TotalStaffRecordsResolver implements Resolve<any> {
+  constructor(private workerService: WorkerService, private establishmentService: EstablishmentService) {}
+
+  resolve(route: ActivatedRouteSnapshot): Observable<Array<UserDetails> | null> {
+    const workplaceUid = route.paramMap.get('establishmentuid')
+      ? route.paramMap.get('establishmentuid')
+      : this.establishmentService.establishmentId;
+
+    return this.workerService.getTotalStaffRecords(workplaceUid).pipe(
+      catchError(() => {
```
Richard-Pentecost, line 19: Sorry misses this one. Does this need to navigate to the problem with a service page?

Richard-Pentecost (commented):
Just a comment on one of the resolvers I missed on the first pass.

aaron-russell (approved):

src/app/core/resolvers/dashboard/total-staff-records.resolver.ts
duncanc19: I think this is one where it maybe doesn't need to navigate, the resolver is only used for the flag on the Staff records tab

PR #4553, opened 2022-02-04
Feat/wdf user permissions

duncanc19:
#### Work done
- Created canManageWdfClaims property and columns
- Created None user role
- Added canManageWdfClaims permission to permissions

#### Tests
- [X] Yes

### DIFF
--- a/migrations/20220203112819-addNoneToUserRoleEnum.js
+++ b/migrations/20220203112819-addNoneToUserRoleEnum.js
@@ -1,0 +1,12 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.query('ALTER TYPE cqc."user_role" ADD VALUE IF NOT EXISTS \'None\'');
+  },
+
+  down: (queryInterface) => {
+    return queryInterface.sequelize.query(`DELETE FROM pg_enum WHERE enumlabel = 'None'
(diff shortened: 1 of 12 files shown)


server/test/unit/utils/security/permissions.spec.js
```diff
@@ -533,6 +537,42 @@ describe('permissions', () => {
...
+    describe('None user', async () => {
+      beforeEach(() => {
+        req.role = 'None';
+      });
+
+      it('should return empty array when role is None without canManageWdfClaims', async () => {
+        const returnedPermissions = await getPermissions(req);
+
+        expect(returnedPermissions).to.deep.equal([]);
+      });
```
aaron-russell, line 564: This should never be a case though?

migrations/20220203112819-addNoneToUserRoleEnum.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface) => {
+    return queryInterface.sequelize.query('ALTER TYPE cqc."user_role" ADD VALUE IF NOT EXISTS \'None\'');
```
aaron-russell, line 5: Can we add a constraint to only allow None if Manage WDF Claims is true?

aaron-russell (changes requested):

aaron-russell (approved):

PR #4581, opened 2022-02-10
Feat/wdf add user page

duncanc19:
#### Work done
- Added WDF permissions options to the Add user page
- Updated user summary table to display new user permissions types

#### Tests
- [X] Yes


server/routes/accounts/user.js
```diff
@@ -474,6 +474,10 @@ const partAddUser = async (req, res) => {
   }
 };
 
+const notPermittedToRegisterNewUser = (role) => !role || role === 'Read' || role === 'None';
+
+const newUserRoleNotValid = (role) => !role || !['Edit', 'Read', 'None'].includes(role);
```
aaron-russell, line 479: It would never get to this as it would be hit on the if above?
```suggestion
const newUserRoleNotValid = (role) => !['Edit', 'Read', 'None'].includes(role);
```

aaron-russell (approved):

PR #4608, opened 2022-02-14
Feat/missing mandatory staff training

Zuhal-Ayob:
#### Work done
- Change the training and qualification summary box for staff to add `mandatory training records need to be added`

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/workers/new-training-qualifications-record/new-training-and-qualifications-record-summary/new-training-and-qualifications-record-summary.component.scss
+++ b/src/app/features/workers/new-training-qualifications-record/new-training-and-qualifications-record-summary/new-training-and-qualifications-record-summary.component.scss
@@ -120,7 +120,7 @@ hr {
 }
 
 .asc-grid-container {
-  display: grid;
+  display: flex;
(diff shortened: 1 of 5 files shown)


src/app/features/workers/new-training-qualifications-record/new-training-and-qualifications-record-summary/new-training-and-qualifications-record-summary.component.scss
```diff
@@ -120,7 +120,7 @@ hr {
 }
 
 .asc-grid-container {
-  display: grid;
+  display: flex;
   grid-column-gap: 8px;
   grid-template-columns: 20% 80%;
 }
```
Richard-Pentecost, line 126: If you are not using `display: grid`, can you get rid of `grid-column-gap` and `grid-template-columns`? Also might be worth renaming the `.asc-grid-container` to `.asc-flex-container`

Richard-Pentecost (commented):
Just a comment about scss

PR #4667, opened 2022-03-01
Bug/staff record nav

mdaley-madetech:
#### Work done
- Adds tests for main-job-start-date component
- Passes reference from staff-records to ensure user is taken back into the 'Save and continue' journey.

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/workers/main-job-start-date/main-job-start-date.component.spec.ts
+++ b/src/app/features/workers/main-job-start-date/main-job-start-date.component.spec.ts
@@ -1,0 +1,189 @@
+import { getTestBed } from '@angular/core/testing';
+import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { FormBuilder, ReactiveFormsModule } from '@angular/forms';
+import { ActivatedRoute, Router } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+import { WorkerService } from '@core/services/worker.service';
+import { MockWorkerService } from '@core/test-utils/MockWorkerService';
+import { WorkplaceService } from '@core/services/workplace.service';
+import { MockWorkplaceService } from '@core/test-utils/MockWorkplaceService';
+import { DatePickerComponent } from '@shared/components/date-picker/date-picker.component';
+import { SubmitButtonComponent } from '@shared/components/submit-button/submit-button.component';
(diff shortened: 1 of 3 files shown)


src/app/features/workers/main-job-start-date/main-job-start-date.component.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    expect(submitSpy).toHaveBeenCalledOnceWith({ action: 'return', save: false });
+    expect(navigateSpy).toHaveBeenCalledWith(['/dashboard'], { fragment: 'workplace', queryParams: undefined });
+  });
+
+  it('renders an error message component has invalid date', async () => {
+    const { getAllByText, getByLabelText, fixture } = await setup();
+
+    userEvent.type(getByLabelText('Day'), '11');
+    userEvent.type(getByLabelText('Month'), '11');
+    userEvent.type(getByLabelText('Year'), '11');
```
duncanc19, line 180: I think it'd be worth updating the tests above to get the day, month and year like this without the expects, it saves a few lines and it's already tested in the correct default fields test

src/app/features/workers/main-job-start-date/main-job-start-date.component.ts
```diff
@@ -16,6 +16,7 @@ import { QuestionComponent } from '../question/question.component';
 })
 export class MainJobStartDateComponent extends QuestionComponent {
   private dateMin = dayjs().subtract(100, 'years');
+  public returnThis?: { url: string[] } | null = this.return;
```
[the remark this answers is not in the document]
mdaley-madetech: No, actually, this was meant to be return as a shadowed variable rather than manipulating the variable in the question component.

src/app/features/workers/main-job-start-date/main-job-start-date.component.ts
```diff
@@ -16,6 +16,7 @@ import { QuestionComponent } from '../question/question.component';
 })
 export class MainJobStartDateComponent extends QuestionComponent {
   private dateMin = dayjs().subtract(100, 'years');
+  public return: { url: string[] } | null = this.return;
```
duncanc19, line 19: I don't think you need this because there's already the return in the QuestionComponent which this component extends, it's just setting it again to the same thing

duncanc19 (approved):

PR #4671, opened 2022-03-03
FIX - Confirm staff record navigation

mdaley-madetech:
#### Work done
- Fix navigation to workplace if establishment is a parent

#### Tests
- [X] Yes

### DIFF
--- a/src/app/features/workers/staff-record/staff-record.component.spec.ts
+++ b/src/app/features/workers/staff-record/staff-record.component.spec.ts
@@ -23,9 +23,8 @@ import { WorkersModule } from '../workers.module';
 import { StaffRecordComponent } from './staff-record.component';
 
 describe('StaffRecordComponent', () => {
-  const workplace = establishmentBuilder() as Establishment;
-
   async function setup() {
+    const workplace = establishmentBuilder() as Establishment;
     const { fixture, getByText, getAllByText, queryByText, getByTestId } = await render(StaffRecordComponent, {
       imports: [SharedModule, RouterModule, RouterTestingModule, HttpClientTestingModule, WorkersModule],
       providers: [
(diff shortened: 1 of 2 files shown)


src/app/features/workers/staff-record/staff-record.component.spec.ts
```diff
@@ -248,5 +257,20 @@ describe('StaffRecordComponent', () => {
         state: { showBanner: true },
       });
     });
+
+    it('should redirect back to the child workplace when worker is updated establishment is a parent', async () => {
```
popey2700, line 261: This doesn't quite make sense, could it be: "should redirect back to the child workplace when the worker and establishment are not the same(parent in child workplace)"

src/app/features/workers/staff-record/staff-record.component.ts
```diff
@@ -99,7 +99,11 @@ export class StaffRecordComponent implements OnInit, OnDestroy {
       this.workerService.updateWorker(this.workplace.uid, this.worker.uid, props).subscribe(
         (data) => {
           this.workerService.setState({ ...this.worker, ...data });
-          this.router.navigate(['/dashboard'], { fragment: 'staff-records', state: { showBanner: true } });
+          const url =
```
popey2700, line 102: Could we consider moving this navigation into a function e.g "returnToHomeTab" so it can be more readable with the concept of navigating to either dashboard or child workplace home

src/app/features/workers/staff-record/staff-record.component.ts
```diff
@@ -108,6 +108,12 @@ export class StaffRecordComponent implements OnInit, OnDestroy {
     );
   }
 
+  public returnToHomeTab() {
+    const isEstablishmentCurrentWorkplace = this.establishmentService.establishmentId === this.workplace.uid;
```
duncanc19, line 112: Is it more descriptive to say isLoggedInWorkplace?

mdaley-madetech: Pushed.

duncanc19 (approved):

PR #4691, opened 2022-03-08
Remove the indentation in Yaml file for disbursement

Zuhal-Ayob:
#### Work done
-

#### Tests
- [ ] Yes

### DIFF
--- a/server/config/preproduction.yaml
+++ b/server/config/preproduction.yaml
@@ -40,7 +40,7 @@ bulkupload:
   lambda:
     stage: sfcpreprod
 
- disbursement:
+disbursement:
     region: eu-west-2
     bucketname: wdf-disbursement-preprod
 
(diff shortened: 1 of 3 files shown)


server/config/preproduction.yaml
```diff
@@ -40,7 +40,7 @@ bulkupload:
   lambda:
     stage: sfcpreprod
 
- disbursement:
+disbursement:
     region: eu-west-2
```
mdaley-madetech, line 44: Should these be one indention to the left? 

It looks like there are 2 tabs rather than 1.

server/config/production.yaml
```diff
@@ -41,7 +41,7 @@ bulkupload:
   lambda:
     stage: sfcprod
 
- disbursement:
+disbursement:
     region: eu-west-2
```
mdaley-madetech, line 45: Same as above

mdaley-madetech (approved):

PR #4715, opened 2022-03-16
Feat/cache quals

Zuhal-Ayob:
#### Work done
- Create a connection with the live disbursement database.
- Get qualification table data and add it to the Redis.
- Create environment variable
- Add environment variables to the circle ci 

#### Tests
- [ ] Yes


.circleci/config.yml
```diff
@@ -578,12 +592,13 @@ workflows:
           appname: sfcbenchmarks
           env_prefix: SFC_BENCHMARKS
           requires:
-            - prod_deploy
```
mdaley-madetech, line 581: It looks like you might have reverted the change for running benchmarks after deploy to live?

mdaley-madetech (commented):
Just take a look at the CircleCI config and the linked ticket as I think it has reverted the updates to benchmarks branch.

PR #4738, opened 2022-03-21
Feat/wdf tab

Richard-Pentecost:
#### Work done
- create new wdf tab component folder and files
- copy over code from wdf overview component, and alter for the tab

#### Tests
- [x] Yes


src/app/features/dashboard/dashboard.component.spec.ts
```diff
@@ -106,94 +112,135 @@ describe('DashboardComponent', () => {
...
+    });
+
+    it('should display the WDF tab', async () => {
+      const { component, fixture, getByTestId } = await setup();
+
+      component.canViewEstablishment = true;
+      fixture.detectChanges();
+
+      expect(getByTestId('tab_wdf')).toBeTruthy();
+    });
```
duncanc19, line 154: Because permissions are getting set a lot for these tests, would it be better to add a param to the setup function and pass in the extra permission to avoid setting it in the component and calling detectChanges each time? 

I'm also wondering with these tests do we also want to test that they don't display when they don't have the permission?

PR #4792, opened 2022-04-05
Revert "Feat/cache quals"

Zuhal-Ayob:
Reverts NMDSdevopsServiceAdm/SopraSteria-SFC#4719


package.json
```diff
@@ -98,7 +98,6 @@
     "pg-hstore": "^2.3.4",
     "pug": "^3.0.1",
     "rate-limit-redis": "^2.1.0",
-    "redis": "^4.0.4",
```
mdaley-madetech, line 101: Not sure if this will cause any issues as some of the work that Richard did required redis.

server/config/config.js
```diff
@@ -124,54 +124,6 @@ const config = convict({
...
-    },
-    port: {
-      doc: 'Database port',
-      format: 'port',
-      default: 1433,
-      env: 'DISBURSEMENT_DB_PORT',
-    },
-  },
-
-  redis: {
```
mdaley-madetech, line 161: Again, worth checking that redis configs are still available for the other code that requires it.

server/models/establishment.js
```diff
@@ -1911,6 +1911,7 @@ module.exports = function (sequelize, DataTypes) {
         where: {
           id: establishmentId,
         },
+        logging: true,
```
mdaley-madetech, line 1914: Not sure if we need this going back in.

mdaley-madetech (commented):
Before merging, please have a check that the redis reverts won't have any impact on the other code that requires the package and config.

server/config/config.js
mdaley-madetech: Test will be checked after revert.

package.json
mdaley-madetech: Test will be checked after revert

mdaley-madetech (approved):

PR #4806, opened 2022-04-05
Feat/toggleSearch

mdaley-madetech:
#### Work done
- Makes search bar conditional based on whether the pagination threshold has been met. 

#### Tests
- [X] Yes

### DIFF
--- a/src/app/shared/components/staff-summary/staff-summary.component.html
+++ b/src/app/shared/components/staff-summary/staff-summary.component.html
@@ -1,5 +1,5 @@
 <div class="govuk-grid-row govuk-util__vertical-align-center">
-  <div class="govuk-grid-column-three-quarters">
+  <div *ngIf="workerCount > itemsPerPage" class="govuk-grid-column-three-quarters">
     <app-search-input
       ref="trainingRecordSearch"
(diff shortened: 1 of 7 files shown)


src/app/shared/components/staff-summary/staff-summary.component.html
```diff
@@ -1,13 +1,15 @@
 <div class="govuk-grid-row govuk-util__vertical-align-center">
-  <div class="govuk-grid-column-three-quarters">
+  <div *ngIf="workerCount > itemsPerPage" class="govuk-grid-column-three-quarters">
     <app-search-input
       ref="trainingRecordSearch"
       accessibleLabel="for staff records"
       (emitInput)="handleSearch($event)"
     ></app-search-input>
   </div>
```
duncanc19, line 8: Won't this mean the search disappears when you do a search and less than a page is returned? I think we might need to do something similar to in the view my workplaces one where we need to differentiate between the first one which is the total of all records and then the count for the search

mdaley-madetech: Updated.

duncanc19 (approved):

PR #4913, opened 2022-05-18
Feat/workplace details remove total staff no

Jonopono123:
#### Work done
-Removed total staff no page from workplace details workflow

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/workplace/data-sharing/data-sharing.component.ts
+++ b/src/app/features/workplace/data-sharing/data-sharing.component.ts
@@ -39,7 +39,7 @@ export class DataSharingComponent extends Question {
       });
     }
 
-    this.previousRoute = ['/workplace', `${this.establishment.uid}`, 'service-users'];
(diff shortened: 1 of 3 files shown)


src/app/features/workplace/data-sharing/data-sharing.component.ts
```diff
@@ -80,7 +80,7 @@ export class DataSharingComponent extends Question {
   }
 
   protected onSuccess(): void {
-    this.nextRoute = ['/workplace', `${this.establishment.uid}`, 'total-staff'];
+    this.nextRoute = ['/workplace', `${this.establishment.uid}`, 'vacancies'];
```
duncanc19, line 83: I know it was already like that but I think it makes sense to remove the back ticks and ${} from the establishment uid, there's no need for it

duncanc19 (approved):

PR #4916, opened 2022-05-19
Feat/workplace details completion banner

popey2700:
#### Work done
- Removed success page for once workplace details are saved
- Added a success alert for once workplace details are added
- Changed navigation to the workplace dashboard tab once workplace details are confirmed

#### Tests
- [x] Yes


src/app/features/workplace/check-answers/check-answers.component.html
```diff
@@ -12,7 +12,8 @@ <h1 class="govuk-heading-l" data-testid="establishmentHeading">
           class="govuk-button"
           role="button"
           draggable="false"
-          [routerLink]="['/workplace', establishment.uid, 'success']"
+          (click)="showConfirmWorkplaceDetailsAlert($event)"
+          style="text-decoration: none"
```
duncanc19, line 16: We normally try to avoid adding styles directly to the html, is it to remove underlining? You could maybe use a button tag instead

src/app/features/workplace/check-answers/check-answers.component.ts
```diff
@@ -30,4 +35,14 @@ export class CheckAnswersComponent implements OnInit, OnDestroy {
   ngOnDestroy(): void {
     this.subscriptions.unsubscribe();
   }
+
+  public showConfirmWorkplaceDetailsAlert(event: Event): void {
+    event.preventDefault();
```
duncanc19, line 40: I think because you've changed it to a button you can remove the event and prevent default, because there's no event related to routing now it's a button tag, it'll just be a click event

duncanc19 (approved):

PR #4936, opened 2022-05-25
Removed unnecessary accordion fields and amended voucher code text

Jonopono123:
#### Work done
Removed unnecessary accordion sections and amended voucher code. Also removed tests that are no longer needed.

#### Tests
- [ ] Yes

### DIFF
--- a/src/app/features/benefits-bundle/benefits-bundle.component.spec.ts
+++ b/src/app/features/benefits-bundle/benefits-bundle.component.spec.ts
@@ -193,7 +193,7 @@ describe('BenefitsBundleComponent', () => {
     });
   });
 
-  fdescribe('Links', () => {
+  describe('Links', () => {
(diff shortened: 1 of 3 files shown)


src/app/features/benefits-bundle/benefits-bundle.component.spec.ts
```diff
@@ -213,15 +185,15 @@ describe('BenefitsBundleComponent', () => {
...
+        fireEvent.click(getByText('10% off all publications in the Skills for Care bookshop'));
         fixture.detectChanges();
 
         expect(getByText('Open all')).toBeTruthy();
       });
     });
   });
 
-  describe('Links', () => {
+  fdescribe('Links', () => {
```
duncanc19, line 196: There's an fdescribe here

PR #5025, opened 2022-06-23
Feat/highlight users where only one added

Jonopono123:
#### Work done
- Added notification flag to users link in header
- Aligned Add Users button with header on Users page

#### Tests
- [x] Yes


src/app/core/components/header/header.component.ts
```diff
@@ -29,6 +31,7 @@ export class HeaderComponent implements OnInit, OnDestroy {
   async ngOnInit(): Promise<void> {
```
Richard-Pentecost, line 31: Can probably get rid of the async

Richard-Pentecost (changes requested):

Richard-Pentecost (approved):

PR #5035, opened 2022-06-28
Feat/staff recruitment navigation

popey2700:
#### Work done
- Created boolean in establishment service to show which flow the user is in
- Implemented dynamic navigation in the new staff recruitment pages
- Updated staff recruitment summary pages when there are no values

#### Tests
- [x] Yes


src/app/core/services/establishment.service.ts
```diff
@@ -168,6 +170,27 @@ export class EstablishmentService {
...
+  public get inStaffRecruitmentFlow() {
+    if (this._inStaffRecruitmentFlow) {
+      return this._inStaffRecruitmentFlow;
+    }
+
+    const inStaffRecruitmentFlow = localStorage.getItem('inStaffRecruitmentFlow');
+
+    if (inStaffRecruitmentFlow) {
+      this._inStaffRecruitmentFlow = JSON.parse(inStaffRecruitmentFlow);
+    } else if (isDevMode() && !this._inStaffRecruitmentFlow) {
```
Zuhal-Ayob, line 182: Hey Alex, what is the benefit of using isDevMode() here ?

popey2700: Hi Zuhal, it's just an Angular function that will return true when we're working locally. This error will therefore only be thrown when in local environments and is useful for debugging

Zuhal-Ayob (approved):

src/app/core/services/establishment.service.ts
Zuhal-Ayob: Thanks Alex, yeah I know it returns a boolean, yeah it's used in many places, ah okay it helps with using the local storage then .

PR #5054, opened 2022-07-06
Feat/add resend user set up

Zuhal-Ayob:
#### Work done
- Add `resendActivationLinkAdmin` function to resend the email for admin user.
- Add `isAdminManager` authorization for resend email link

#### Tests
- [x] Yes

### DIFF
--- a/src/app/core/services/user.service.ts
+++ b/src/app/core/services/user.service.ts
@@ -155,6 +155,12 @@ export class UserService {
     });
   }
 
+  public resendActivationLinkAdmin(useruid: string) {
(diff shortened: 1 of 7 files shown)


src/app/core/services/user.service.ts
```diff
@@ -155,6 +155,12 @@ export class UserService {
     });
   }
 
+  public resendActivationLinkAdmin(useruid: string) {
```
Richard-Pentecost, line 158: Can we put this in the admin-user.service. Other endpoints that are being used in the admin-user section are in the admin-user service, so I think it makes more sense to be in there

Richard-Pentecost (changes requested):

PR #5079, opened 2022-07-15
Added progress bar to services-capacity component

Jonopono123:
#### Work done
Added progress bar to services-capacity.

#### Tests
- [ ] Yes


Richard-Pentecost (changes requested):
The code looks good, but we need some tests around the component. There isn't a test file at m, so don't worry about testing the functionality that you haven't affected, but testing around whether the progress bar shows is not should be in

Richard-Pentecost (approved):

PR #5128, opened 2022-07-28
Bug/admin Download Most Recent Upload

popey2700:
#### Work done
- Changed how type of file is working in last-bulk-upload 

#### Tests
- [ ] Yes

### DIFF
--- a/src/app/features/bulk-upload/last-bulk-upload/last-bulk-upload.component.ts
+++ b/src/app/features/bulk-upload/last-bulk-upload/last-bulk-upload.component.ts
@@ -42,16 +42,16 @@ export class LastBulkUploadComponent implements OnInit {
     return encodeURI(url);
   }
 
-  public downloadFile(event: Event, key: string): void {
+  public downloadFile(event: Event, key: string, fileType: string): void {
     event.preventDefault();
 
     let type: BulkUploadFileType = null;
 
-    if (key.includes('staff')) {
+    if (fileType === 'Worker') {
       type = this.sanitise ? BulkUploadFileType.WorkerSanitise : BulkUploadFileType.Worker;
-    } else if (key.includes('workplace')) {
+    } else if (fileType === 'Establishment') {
       type = BulkUploadFileType.Establishment;
-    } else if (key.includes('training')) {
+    } else if (fileType === 'Training') {
       type = BulkUploadFileType.Training;
     }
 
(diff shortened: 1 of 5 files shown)


src/app/features/bulk-upload/last-bulk-upload/last-bulk-upload.component.ts
```diff
@@ -42,16 +42,16 @@ export class LastBulkUploadComponent implements OnInit {
     return encodeURI(url);
   }
 
-  public downloadFile(event: Event, key: string): void {
+  public downloadFile(event: Event, key: string, fileType: string): void {
```
Richard-Pentecost, line 45: The filetype is actually of type BulkUploadFileType. Then instead of having that long if else statement, you could have `let type: BulkUploadFileType = fileType`. Then just do simple ternary on type, something like `type === BulkUploadFileType.Worker && this.sanitise && type = BulkUploadFileType.WorkerSanitise`

Richard-Pentecost (approved):

src/app/features/bulk-upload/last-bulk-upload/last-bulk-upload.component.ts
```diff
@@ -42,16 +42,16 @@ export class LastBulkUploadComponent implements OnInit {
...
 
-    if (key.includes('staff')) {
+    if (fileType === 'Worker') {
       type = this.sanitise ? BulkUploadFileType.WorkerSanitise : BulkUploadFileType.Worker;
-    } else if (key.includes('workplace')) {
+    } else if (fileType === 'Establishment') {
       type = BulkUploadFileType.Establishment;
-    } else if (key.includes('training')) {
+    } else if (fileType === 'Training') {
       type = BulkUploadFileType.Training;
```
Richard-Pentecost, line 55: If you don't want to do the refactor above, the if else statements would be better if they use the constants. So `if (fileType === BulkUploadFileType.Worker)` etc.

src/app/features/bulk-upload/last-bulk-upload/last-bulk-upload.component.ts
Richard-Pentecost: Could also do `if (type === BulkUploadFileType.Worker && this.sanitise) type = BulkUploadFileType.WorkerSanitise` if it's clearer

PR #5146, opened 2022-08-02
Feat/staff benefits pensions

Jonopono123:
#### Work done
- Added pensions page to staff recruitment workflow
- Added necessary property to back end

#### Tests
- [x] Yes


src/app/core/model/establishment.model.ts
```diff
@@ -183,6 +183,12 @@ export enum StaffBenefitEnum {
   DONT_KNOW = "Don't know",
 }
 
+export enum PensionsEnum {
+  YES = 'Yes',
+  NO = 'No',
+  DONT_KNOW = "Don't know",
+}
+
```
popey2700, line 191: Rather than create a new Enum here we can just use the StaffBenefitEnum as they're the same. You'll need to change the pension component to use the StaffBenefitEnum

src/app/features/workplace/pensions/pensions.component.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+    // it('should navigate to the next page when submitting from the flow', async () => {
+    //   const { fixture, getByText, routerSpy } = await setup(false);
+
+    //   const button = getByText('Save and continue');
+    //   fireEvent.click(button);
+    //   fixture.detectChanges();
+
+    //   expect(routerSpy).toHaveBeenCalledWith(['/workplace', 'mocked-uid', 'pensions']);
+    // });
+
```
popey2700, line 147: Can we include these tests

popey2700 (approved):

PR #5170, opened 2022-08-15
Fix/inactive workplaces query

Zuhal-Ayob:
#### Work done
- Rewrite the archive inactive workplaces query so when a parent workplace has an active child it shouldn’t be deleted.

#### Tests
- [ ] Yes


src/app/features/workers/nursing-category/nursing-category.component.spec.ts
```diff
@@ -63,7 +63,6 @@ describe('NursingCategoryComponent', () => {
       const { getByText } = await setup();
 
       expect(getByText('Save and return')).toBeTruthy();
-      expect(getByText('Cancel')).toBeTruthy();
```
popey2700, line 66: Why are we deleting this expect statement?

popey2700 (approved):

PR #5218, opened 2022-08-30
Feat/non cqc workplace name

Jonopono123:
#### Work done
Added progress bar to workplace name on non CQC path of registration and add-workplace.

#### Tests
- [x] Yes


src/app/shared/directives/create-workplace/name-of-workplace/name-of-workplace.directive.ts
```diff
@@ -32,6 +36,9 @@ export class NameOfWorkplaceDirective implements OnInit, AfterViewInit {
 
   ngOnInit(): void {
     this.flow = this.route.snapshot.parent.url[0].path;
```
Richard-Pentecost, line 38: Think you can remove this line here as you are setting flow in the components

Richard-Pentecost (approved):

src/app/shared/directives/create-workplace/name-of-workplace/name-of-workplace.directive.ts
Jonopono123: Done

PR #5350, opened 2022-10-13
Feat/salary

gab-bernotaite:
#### Work done
- Added progress bar
- Added 'Employment details' header
- Text changes
- Changed routing code

#### Tests
- [x] Yes


migrations/20221012110537-updateJobRolesList.js
```diff
@@ -0,0 +1,23 @@
...
+  down: (queryInterface, Sequelize) => {
+    /*
+      Add reverting commands here.
+      Return a promise to correctly handle asynchronicity.
+
+      Example:
+      return queryInterface.dropTable('users');
+    */
+  }
+};
```
Richard-Pentecost, line 23: This file shouldn't be in the pr.

src/app/features/workers/question/question.component.ts
```diff
@@ -61,7 +61,7 @@ export class QuestionComponent implements OnInit, OnDestroy, AfterViewInit {
         this.worker = worker;
 
         if (!this.initiated) {
-          this._init();
+          this.init();
```
Richard-Pentecost, line 64: I think we need this to be this_init(). this.init() is called form within the this._init() function

Richard-Pentecost (changes requested):

Richard-Pentecost (approved):

PR #5649, opened 2023-02-06
Feat/table pagination component

Richard-Pentecost:
#### Work done
- create new table wrapper component with pagination
- add to staff record and t and q tab
- add to view training page
- create new backend for retrieving paginated training for both mandatory and non-mandatory training

#### Tests
- [x] Yes


server/models/establishment.js
```diff
@@ -1833,6 +1838,7 @@ module.exports = function (sequelize, DataTypes) {
           ],
         },
       ],
+      logging: console.log,
```
Zuhal-Ayob, line 1841: do we need this logging?

Zuhal-Ayob (approved):

PR #5704, opened 2023-02-27
change fetching training for establishment to take job Ids when getti…

Richard-Pentecost:
…ng mandatory training

#### Work done
-

#### Tests
- [x] Yes


server/models/workerTraining.js
```diff
@@ -204,11 +210,13 @@ module.exports = function (sequelize, DataTypes) {
     }[sortBy] || [['worker', 'NameOrIdValue', 'ASC']];
 
     const response = await this.findAll({
+      logging: console.log,
```
popey2700, line 213: Do we want this left in?

popey2700 (approved):

PR #5727, opened 2023-03-09
Fix/benchmarks header

Zuhal-Ayob:
#### Work done
- Fix benchmark header and messages

#### Tests
- [ ] Yes

### DIFF
--- a/server/utils/security/permissions.js
+++ b/server/utils/security/permissions.js
@@ -58,6 +58,7 @@ const readPermissions = (establishmentAndUserInfo) => [
   'canViewListOfUsers',
   'canDownloadWdfReport',
   'canViewNinoDob',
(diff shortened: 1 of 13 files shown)


server/utils/security/permissions.js
```diff
@@ -132,6 +136,19 @@ const getAdditionalEditPermissions = (estabType, establishmentAndUserInfo, isLog
   return additionalPermissions.filter((item) => item !== undefined);
 };
 
+const getAdditionalReadPermissions = (establishmentAndUserInfo) => {
+  console.log({ establishmentAndUserInfo });
```
Jonopono123, line 140: Could do with removing this console log

Jonopono123 (approved):

PR #5750, opened 2023-03-23
Spike/e2e test setup

Richard-Pentecost:
#### Work done
- add some e2e tests
- setup page_object folder for testing login page
- add programmatic login for cypress tests
- make changes to docker-compose and dockerfile

#### Tests
- [x] Yes


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
popey2700, line 1: Assuming this is just from boiler plate generation - do we still need this ?

cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
popey2700, line 1: Seems to again be more boilerplate code - do we need to this?

cypress/integration/registration/createanaccount.spec.js
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable no-undef */
+/// <reference types="cypress" />
+
+xdescribe('Create account', () => {
```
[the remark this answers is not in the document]
Richard-Pentecost: Yeah at the moment we need this, because the test isn't working properly.

cypress/fixtures/example.json
Richard-Pentecost: It is, while we don't need atm. It is useful to know that we can add data into it, so think it is worth keeping for the time being

server.js
```diff
@@ -317,6 +317,7 @@ const startApp = () => {
   console.log('Listening on port: ' + app.get('port'));
 };
 
+console.log('AppConfig.ready', AppConfig.ready);
```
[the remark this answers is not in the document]
Richard-Pentecost: I'l remove

popey2700 (approved):

PR #5771, opened 2023-04-12
Feat/home page blue banner

Richard-Pentecost:
#### Work done
- add blue border to the top of the header
- move notifications link to the header

#### Tests
- [x] Yes

### DIFF
--- a/src/app/features/new-dashboard/home-tab/home-tab.component.ts
+++ b/src/app/features/new-dashboard/home-tab/home-tab.component.ts
@@ -116,10 +116,6 @@ export class NewHomeTabComponent implements OnInit, OnDestroy {
       this.permissionsService.can(workplaceUid, 'canBecomeAParent') && !this.linkToParentRequestedStatus;
   }
 
-  public pushLinkToHistoryState(urlLink: string): void {
-    history.pushState({}, '', urlLink);
-  }
-
   /**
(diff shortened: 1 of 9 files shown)


src/app/features/new-dashboard/home-tab/home-tab.component.ts
```diff
@@ -116,10 +116,6 @@ export class NewHomeTabComponent implements OnInit, OnDestroy {
       this.permissionsService.can(workplaceUid, 'canBecomeAParent') && !this.linkToParentRequestedStatus;
   }
 
-  public pushLinkToHistoryState(urlLink: string): void {
-    history.pushState({}, '', urlLink);
-  }
-
```
Jonopono123, line 122: I seem to recall Sabrina adding this last week, is it no longer necessary?

Richard-Pentecost: Yeah, it is causing issues with backlinks. I'm not sure why the problem with the links not changing colour occurred. I've not seen it, but we shouldn't need to be manually adding to history. I've added the govuk-link class to try and make sure it works properly but will have to speak to Billy and Rich when they are back

Jonopono123 (approved):

RobertM905 (approved):
Looks good

PR #5851, opened 2023-06-02
Set benchmarks card text dynamically

Jonopono123:
#### Work done
-Added additional dynamic text to benchmarks car don home page

#### Tests
- [x] Yes


Zuhal-Ayob (approved):

src/app/features/new-dashboard/home-tab/home-tab.component.ts
```diff
@@ -115,19 +116,32 @@ export class NewHomeTabComponent implements OnInit, OnDestroy {
...
+    const comparisonDataAvailable = this.meta.staff && this.meta.workplaces;
+
+    if(!comparisonDataAvailable) {
+      this.benchmarksHeader = 'See how you compare against other workplaces';
+      this.benchmarksMessage = `Benchmarks can show how you're doing when it comes to pay, recruitment and retention.`;
+    } else {
+      if(!this.workplace.isRegulated) {
+        this.benchmarksHeader = 'See how you compare against other workplaces'
+        this.benchmarksMessage = `There are ${this.meta?.workplaces ? this.meta.workplaces : 0} workplaces providing adult social care in ${this.meta?.localAuthority}.`;
+      } else if([1, 2, 8].filter(x => x === this.workplace.mainService.id).length > 0) {
```
RobertM905: main service id is not the thing to check against for this. (I did the same thing before). 
I merged a change into test branch that adds reportingID to the workplace.MainService object so we can use that if we can. (possible refactor in the future to add a single flag on the establishment to avoid these "Magic numbers" everywhere.)

For now use `[1, 2, 8].includes(this.workplace.mainService.reportingID)`

src/app/features/new-dashboard/home-tab/home-tab.component.ts
```diff
@@ -115,19 +116,34 @@ export class NewHomeTabComponent implements OnInit, OnDestroy {
...
+    if (!comparisonDataAvailable) {
+      this.benchmarksHeader = 'See how you compare against other workplaces';
+      this.benchmarksMessage = `Benchmarks can show how you're doing when it comes to pay, recruitment and retention.`;
+    } else {
+      if (!this.workplace.isRegulated) {
+        this.benchmarksHeader = 'See how you compare against other workplaces';
+        this.benchmarksMessage = `There are ${
+          this.meta?.workplaces ? this.meta.workplaces : 0
+        } workplaces providing adult social care in ${this.meta?.localAuthority}.`;
+      } else if ([1, 2, 8].filter((x) => x === this.workplace.mainService.id).length > 0) {
```
RobertM905, line 135: main service id is not the thing to check against for this. (I did the same thing before).
I merged a change into test branch that adds reportingID to the workplace.MainService object so we can use that if we can. (possible refactor in the future to add a single flag on the establishment to avoid these "Magic numbers" everywhere.)

For now use [1, 2, 8].includes(this.workplace.mainService.reportingID)

RobertM905 (changes requested):

src/app/features/new-dashboard/home-tab/home-tab.component.ts
Jonopono123: Done, this now uses reportingID

RobertM905 (approved):

PR #5952, opened 2023-08-17
Remove benchmarks resolver and route from the service

Zuhal-Ayob:


src/app/core/components/standAloneAccount/standAloneAccount.component.spec.ts
```diff
@@ -112,19 +112,19 @@ describe('StandAloneAccountComponent', () => {
...
-    });
-
-    it('should not run postBenchmarkTabUsage when called with a tab slug that is not benchmarks', async () => {
-      const { component, benchmarksSpy } = await setup();
-
-      component.tabClickEvent({ tabSlug: 'home' });
-      expect(benchmarksSpy).not.toHaveBeenCalled();
-    });
-  });
+  // xdescribe('tabClickEvent', () => {
```
Jonopono123, line 115: If this uses `xdescribe` does it need to be commented out?

Jonopono123 (approved):

src/app/core/components/standAloneAccount/standAloneAccount.component.spec.ts
Zuhal-Ayob: Hey Jon, got you, yes, because the function wasnt any more so it was erroring

PR #5965, opened 2023-09-13
feat/1325 link to parent pending

ssrome:
#### Work done
- Created a model for link to parent table
- Add new route to link to parent for requested details
- Add functionality for link to parent pending to navigate to page instead of dialog

#### Tests
- [x] Yes


server/routes/establishments/linkToParent.js
```diff
@@ -5,6 +5,7 @@ const Establishment = require('../../models/classes/establishment');
 const { v4: uuidv4 } = require('uuid');
 uuidv4();
 const linkSubToParent = require('../../data/linkToParent');
+const models = require('../../models');
```
magicmilo, line 8: Just looking at this and seeing as we have it everywhere.
I think we really should be only importing the models we need from now on if possible.

magicmilo (approved):
Really nice

PR #5971, opened 2023-09-29
Fix/green banner

ssrome:
#### Work done
- 

#### Tests
- [ ] Yes


src/app/features/new-dashboard/become-a-parent/become-a-parent.component.ts
```diff
@@ -74,7 +74,7 @@ export class BecomeAParentComponent implements OnInit, OnDestroy {
         if (data) {
           this.router.navigate(['/dashboard'], {
             state: {
-              successAlertMessage: `You’ve sent a request to become a parent workplace`,
+              alertMessage: `You’ve sent a request to become a parent workplace`,
```
sbt92, line 77: You've apostrophe is different between L77 and L93. I am aware this is the same as before but should this be changed for consistency given we're changing the code here anyhow?

Zuhal-Ayob (approved):

PR #5992, opened 2023-11-23
view new benchmarks on subs when logged in as parent

Jonopono123:
#### Work done
- added ability to view new benchmarks on subs when logged in as parent

#### Tests
- [ ] Yes


sbt92 (approved):
LGTM

PR #5993, opened 2023-11-23
Fix/view sub benchmarks

Jonopono123:
#### Work done
- Follow up fix to add reportingID to condition for viewing data area on subs

#### Tests
- [x] Yes


sbt92 (approved):
LGTM

PR #6015, opened 2023-12-14
Fix/delete training records

ssrome:
#### Work done
- Add function in add-edit-training to route to delete record as previous way caused issues
- Update karma.conf to run ChromeHeadlessNoSandbox

#### Tests
- [x] Yes


magicmilo (approved):
Good stuff

PR #6029, opened 2024-01-08
Feat/local dev setup

magicmilo:
#### Work done
- Add Makefile (Run 'make install' then 'make run') from root
- Moved localhost server config
- Fixed 'mocha all' tests path

#### Tests
- [ ] Yes


Jonopono123 (approved):

Makefile
```diff
@@ -0,0 +1,16 @@
+.DEFAULT_GOAL := all
+.PHONY: all install run
+
+all: install run
+
+install:
+	export DB_HOST=127.0.0.1
```
sbt92, line 7: May make more sense to add this in the localhost.yaml

PR #6056, opened 2024-01-17
Disable send in blue whitelist when parameter = off

sbt92:
#### Work done
- Disable whitelist when parameter = off

#### Tests
- [ ] Yes


magicmilo:
This is as the parameter can never be empty. So we now allow it to be 'off'. default is changeme

magicmilo (approved):

PR #6087, opened 2024-02-09
Feat/1379 missing cqc workplaces

ssrome:
#### Work done
- [Work done for ticket](https://trello.com/c/Puzen889/1379-missing-cqc-workplaces)
- Create new `missingCqcLocations` api
- Add missing CQC locations message to parent-home-tab and view-my-workplaces 
- Add getMissingCqcLocations resolver

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/approvals.js
+++ b/backend/server/models/approvals.js
@@ -162,7 +162,7 @@ module.exports = (sequelize, DataTypes) => {
         'Data',
         'Reviewer',
         'InReview',
-        'createdAt',
+        'updatedAt',
       ],
       include: [
         {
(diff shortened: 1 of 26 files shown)


backend/server/models/approvals.js
```diff
@@ -162,7 +162,7 @@ module.exports = (sequelize, DataTypes) => {
         'Data',
         'Reviewer',
         'InReview',
-        'createdAt',
```
magicmilo, line 165: Is there a migration needed to remove this column?

ssrome: It was on the query twice so just replaced one with a column that already exists, which I required

magicmilo (approved):

PR #6106, opened 2024-02-28
NHSBSA-API changes

Zuhal-Ayob:
#### Work done
- Add two routes for NHSBSA API (one for authentication and another for data retrieval )
- Add sequelize query to retrieve data from database
- Add function to retrieve WDF eligibility for each workplace
- Add authentication with jwt token
- Add ratelimiting

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/establishment.js
+++ b/backend/server/models/establishment.js
@@ -2380,16 +2380,6 @@ module.exports = function (sequelize, DataTypes) {
 
   };
 
-  Establishment.getWorkplaceId= async function (workplaceId) {
-    return await this.findOne({
-      attributes: ['nmdsId'],
-
-      where: {
-        archived: false,
-        nmdsId: workplaceId,
-      },
-    });
-  };
 
   return Establishment;
 };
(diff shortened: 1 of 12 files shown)


backend/server/routes/nhsBsaApi/workplaceData.js
```diff
@@ -0,0 +1,129 @@
...
+    }),
+  );
+  return subsidiaries;
+};
+
+const parentList = async (parentId) => {
+  const where = {
+    id: parentId,
+  };
+  const subs = await models.establishment.nhsBsaApiData(where);
```
duncanc19, line 91: This variable should be the parent workplace right?

backend/server/models/establishment.js
```diff
@@ -2322,5 +2322,64 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
+
+  const attributes = [
```
duncanc19, line 2326: Could we rename this to something like nhsBsaAttributes?

duncanc19 (approved):

PR #6115, opened 2024-03-11
Training tab view page in t and q

jpmaHub:
#### Work done
- Added subsidiary route to the training tab in Training and qualification tab
- Hide the banner when viewing the staff and training in Training and qualifications tab

#### Tests
- [ ] Yes
- [ x] No, they are not required for this change

### DIFF
--- a/frontend/src/app/shared/components/training-and-qualifications-categories/training-and-qualifications-categories.component.ts
+++ b/frontend/src/app/shared/components/training-and-qualifications-categories/training-and-qualifications-categories.component.ts
@@ -1,7 +1,9 @@
 import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angular/core';
 import { Establishment, SortTrainingAndQualsOptionsCat } from '@core/model/establishment.model';
+import { BreadcrumbService } from '@core/services/breadcrumb.service';
 import { EstablishmentService } from '@core/services/establishment.service';
 import { TrainingStatusService } from '@core/services/trainingStatus.service';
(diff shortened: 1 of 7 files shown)


frontend/src/app/shared/components/training-and-qualifications-categories/training-and-qualifications-categories.component.ts
```diff
@@ -23,20 +25,29 @@ export class TrainingAndQualificationsCategoriesComponent implements OnInit, OnD
...
   constructor(
     protected trainingStatusService: TrainingStatusService,
     private establishmentService: EstablishmentService,
+    private parentSubsidiaryViewService: ParentSubsidiaryViewService,
+    private breadcrumbService: BreadcrumbService,
   ) {}
 
   ngOnInit(): void {
+    if (this.parentSubsidiaryViewService.getViewingSubAsParent()) {
+      this.link = '/subsidiary/workplace';
```
magicmilo, line 39: We are trying to avoid doing this on any existing components, does the router service not catch these links? I believe Jon is working on a fix for the router that should ensure we do not have to do this anywhere.

magicmilo (changes requested):

frontend/src/app/shared/components/training-and-qualifications-categories/training-and-qualifications-categories.component.ts
jpmaHub: no, it did not.

PR #6117, opened 2024-03-12
changed sub router service to override navigateByUrl rather than Navi…

Jonopono123:
#### Work done
- Changed the subsidiary-router-service to override `router.navigateByUrl()` rather than `router.navigate()`, this should resolve the issues we're seeing when using the `[routerLink]` directive

#### Tests
- [x] Yes


magicmilo (approved):
Brilliant

PR #6136, opened 2024-03-19
Fix: Minor fixes for View sub as parent feature

duncanc19:
#### Work done
- Removed added code which was no longer needed
- Simplified about-the-data-link component

#### Tests
- [ ] Yes


Jonopono123 (approved):
Looks good

PR #6142, opened 2024-03-26
Fix/approve parent link banner

ssrome:
#### Work done
- Remove send alert when parent approves or rejects link to parent request so the green banner doesn't appear

#### Tests
- [ ] Yes


duncanc19 (commented):
I thought the issue was that the banner was flashing up and then disappearing, and we expect the banner to stay there until the user clears it. Is the banner not supposed to appear at all?

duncanc19 (approved):

ssrome:
Billy and Rich confirmed green banner shouldn't appear

PR #6147, opened 2024-03-26
Fix benchmark view in sub to show new and old benchmark data

jpmaHub:
#### Work done
- Added canSeeNewData flag
- Remove benchmark tab components from app module and add it to the benchmark module

#### Tests
- [ ] Yes


Jonopono123 (changes requested):
There's a couple of failing tests here that need a quick fix unfortunately

Jonopono123 (approved):

PR #6149, opened 2024-03-26
Fix/news article

Zuhal-Ayob:
#### Work done
- Change the URL for the article news.
- Add navigation for the Return to home button

#### Tests
- [x] Yes


magicmilo (approved):

frontend/src/app/features/articles/article/article.component.spec.ts
```diff
@@ -66,4 +66,13 @@ describe('ArticleComponent', () => {
...
   });
+
+  it('should navigate back to home tab', async () => {
+    const { fixture, getByTestId } = await setup();
+    const button = getByTestId('returnButton');
+    fireEvent.click(button);
+    fixture.detectChanges();
+
+    expect(button.getAttribute('href')).toEqual('/dashboard');
+  });
```
duncanc19, line 77: I think you can simplify the test a bit. You could use getByText on 'Return to home' and because you're checking the href I don't think you need to click it

duncanc19 (approved):

PR #6151, opened 2024-03-27
Add unit tests

jpmaHub:
#### Work done
- Add test units 

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/new-dashboard/benchmarks-tab/benchmarks-tab.component.spec.ts
+++ b/frontend/src/app/features/new-dashboard/benchmarks-tab/benchmarks-tab.component.spec.ts
@@ -99,7 +99,6 @@ describe('NewBenchmarksTabComponent', () => {
       queryByTestId,
       getByText,
       pdfService,
-      fixture,
     };
   };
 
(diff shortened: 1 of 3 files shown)


frontend/src/app/features/new-dashboard/benchmarks-tab/benchmarks-tab.component.spec.ts
```diff
@@ -115,4 +117,11 @@ describe('NewBenchmarksTabComponent', () => {
     expect(downloadFunctionSpy).toHaveBeenCalled();
     expect(pdfServiceSpy).toHaveBeenCalled();
   });
+
+  it('should not render the banner when in subsidiary view of benchmark tab', async () => {
+    const { component, fixture, queryByTestId } = await setup();
+    component.showBanner = false
+    fixture.detectChanges();
+    expect(queryByTestId("showBannerTestId")).toBeFalsy();
+  });
```
duncanc19, line 126: I think if it's straightforward it's better to try to set variables directly, so in this case showBanner is an input, so we can add it to the componentProperties in the setup function and pass it a boolean if we have tests where we want it to be true and others where we want it false

duncanc19 (approved):

PR #6157, opened 2024-03-27
Fix/update banner when going to home using breadcrumbs

duncanc19:
#### Work done
- Updated subsidiary components to use tabs service for keeping track of tabs
- Removed showSelectedTab from parent subsidiary view service

#### Tests
- [ ] Yes


Jonopono123 (approved):
Nice one

PR #6164, opened 2024-04-02
Allow workplace users tab to load the users

jpmaHub:
#### Work done
- Added user resolver to the subsidiary workplace-users routing
- Added div to add spacing between the banner and the content title
- Added route to return to the workplace users tab in return URL method
- Refactored the sub view workplace-users tab and users components to use the shared user component 
- Exit sub view when visiting to the users route link 

#### Tests
- [x] Yes


frontend/src/app/features/workplace/users/users.component.ts
```diff
@@ -2,64 +2,33 @@ import { Component, OnInit } from '@angular/core';
...
   selector: 'app-users',
   templateUrl: './users.component.html',
 })
 export class UsersComponent implements OnInit {
   public workplace: Establishment;
-  public showSecondUserBanner: boolean;
   public users: UserDetails[];
   public canAddUser: boolean;
   public canViewUser: boolean;
   public userPermissionsTypes: UserPermissionsType[];
```
duncanc19, line 18: Can some of these variables be removed now most of the functionality is in the user-accounts-summary component?

frontend/src/app/shared/services/subsidiary-router-service.ts
```diff
@@ -56,7 +56,11 @@ export class SubsidiaryRouterService extends Router {
 
     if (exitSubsidiaryViewPages.includes(commands[0])) {
       this.parentSubsidiaryViewService.clearViewingSubAsParent();
-    } else {
+    }
+    else if ((commands[0] === 'workplace' && commands[2] === 'users')){
+      this.parentSubsidiaryViewService.clearViewingSubAsParent();
+    }
```
duncanc19, line 62: This could get a bit long if there are other instances where we need to check specific sections of the url. I think this is fine at the moment but if more of these come up we could look into whether we can check if commands includes the string (I know we had it set up like that before and I guess there's a reason we can't) or maybe change the exitSubsidiaryViewPages array to have objects with the command sections as well.

duncanc19: Actually thinking about it, I think this might throw an error in the case where the url has fewer than three commands because it'll throw an array out of bounds

jpmaHub: yes

jpmaHub: but you get undefined not an actual error if you trying to access the command with less than three @duncanc19

duncanc19: Ok, I think we still don't want to be checking on a random string. I think it might be better just to have a function like isUsersUrl and have a check for the length of commands before you check the content of commands

duncanc19 (approved):

PR #6168, opened 2024-04-02
Fix/navigation after deleting workplace

duncanc19:
#### Work done
- Updated dashboard header to navigate back to view-all-workplaces when parent has deleted a sub or sub deleted by admin from parent view
- Updated navigation to admin workplace search page when stand alone or sub viewed directly deleted

#### Tests
- [X] Yes


Jonopono123 (approved):
Good stuff

PR #6181, opened 2024-04-16
added new job roles migration

Jonopono123:
#### Work done
- Added db migration for new job roles

#### Tests
- [ ] Yes


duncanc19 (approved):
Looks good! Doesn't matter but just to let you know, you could add them all in one go by having each row in brackets separated by commas instead of separate queries

PR #6182, opened 2024-04-16
Update url on view your benchmarks button

ssrome:
#### Work done
- Update url for view your benchmarks button in about the data page for benchmarks v1

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/shared/components/benchmarks-tab/about-the-data/about-the-data.component.spec.ts
+++ b/frontend/src/app/shared/components/benchmarks-tab/about-the-data/about-the-data.component.spec.ts
@@ -97,7 +97,7 @@ describe('BenchmarksAboutTheDataComponent', () => {
     expect(lineItemCount).toBe(2);
   });
 
-  it('should navigate to the benchmarks page', async () => {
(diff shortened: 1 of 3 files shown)


duncanc19 (commented):
Looks good! Can we remove the fragment and maybe url from the component because we're not using them now? With these changes do we still need the returnTo in the benchmarks service?

frontend/src/app/shared/components/benchmarks-tab/about-the-data/about-the-data.component.spec.ts
```diff
@@ -46,38 +43,68 @@ describe('BenchmarksAboutTheDataComponent', () => {
...
   it('should have the right text with no data', async () => {
+    const { component, fixture } = await setup();
     component.meta = null;
     fixture.detectChanges();
+
     const lineItemCount = fixture.debugElement.queryAll(By.css('li')).length;
     expect(lineItemCount).toBe(2);
   });
+
+  it('should navigate to the benchmarks page', async () => {
```
duncanc19, line 100:
```suggestion
  it('should navigate to the benchmarks page when you click View your benchmarks button', async () => {
```

duncanc19 (approved):

PR #6185, opened 2024-04-17
Add fragment to the route to navigate to the right flow in the benchmark breadcrumbs

jpmaHub:
#### Work done
- Included fragment to the get route instead of in the referrer object. This will apply the fragment to both home and benchmark routes. 
#### Tests
- [ ] Yes

### DIFF
--- a/frontend/src/app/core/breadcrumb/journey.workplaces.ts
+++ b/frontend/src/app/core/breadcrumb/journey.workplaces.ts
@@ -86,10 +86,6 @@ export const benchmarksTabJourney: JourneyRoute = {
         {
           title: 'About the data',
           path: Path.ABOUT_DATA,
-          referrer: {
-            path: Path.DASHBOARD,
-            fragment: 'benchmarks',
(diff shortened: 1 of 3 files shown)


frontend/src/app/core/breadcrumb/journey.workplaces.ts
```diff
@@ -76,46 +76,21 @@ export const trainingAndQualificationsTabJourney: JourneyRoute = {
...
-            path: Path.DASHBOARD,
-            fragment: 'benchmarks',
-          },
-        },
-      ],
-    },
-  ],
-};
 
+export function benchmarksTabJourney(isOldTab: boolean = false): JourneyRoute {
```
Jonopono123, line 81: Could you possibly add a couple of tests for this? Should just be a case of checking that the correct object is returned based on the value of the `isOldTab` parameter

Jonopono123 (approved):
Looks great, nice work!

PR #6187, opened 2024-04-18
fix/staff-basic-records-with-mandatory-data

ssrome:
#### Work done
- Change workplace to use establishment resolver in staff basic records
- This fixes issues with staff-basic-records not loading staff records correctly in sub view and records with only mandatory data added links not going to correct place. Clicking add more details should go to date of birth.

#### Tests
- [ ] Yes


Jonopono123 (commented):
Looks good!

Jonopono123 (approved):

PR #6190, opened 2024-04-22
The parent news page now display the first featured article that appe…

jpmaHub:
#### Work done
- Display the featured news article that appears first in the list

#### Tests
- [x] Yes


frontend/src/app/shared/directives/new-home-tab/new-home-tab.directive.ts
```diff
@@ -128,6 +130,12 @@ export class NewHomeTabDirective implements OnInit, OnDestroy, OnChanges {
 
     this.newHomeDesignParentFlag = this.featureFlagsService.newHomeDesignParentFlag;
 
+    this.subscriptions.add(
+      this.route.url.subscribe(() => {
+        this.article = this.route.snapshot.data.articleList?.data[0];
+      }),
+    );
```
duncanc19, line 137: Do we need to subscribe to the url here? I think getting data from snapshot means it should already be resolved so we should just be able to set it directly

PR #6201, opened 2024-04-29
Feat/changes to training categories

duncanc19:
#### Work done
- Renamed training categories
- Added new training categories
- Added new training categories to Bulk Upload
- Simplified sequencing of training categories to make it easier to add new ones (sort to end by setting seq to 1, then alphabetical)
- Added database migrations to script to run in different environments

#### Tests
- [ ] Yes


backend/migrations/20240415084724-simplifySequencingForTrainingCategories.js
Jonopono123, line 1: Not going to make you rewrite all this but going forward we should look to use the sequelize functionality rather than raw sql

Jonopono123 (approved):

PR #6204, opened 2024-04-29
Fix alert banners in home page 

jpmaHub:
#### Work done
- Added alert service to link to parent notification messages

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/new-dashboard/home-tab/home-tab.component.spec.ts
+++ b/frontend/src/app/features/new-dashboard/home-tab/home-tab.component.spec.ts
@@ -43,15 +43,11 @@ const MockWindow = {
 };
 
 describe('NewHomeTabComponent', () => {
-  type AlertType = 'success' | 'warning' | 'pending';
-  const type = 'success' as AlertType
-
   const setup = async (
     checkCqcDetails = false,
     establishment = Establishment,
     comparisonDataAvailable = true,
     noOfWorkplaces = 9,
-    message = ''
   ) => {
     const { fixture, getByText, queryByText, getByTestId, queryByTestId, getByRole, getByLabelText } = await render(
       NewHomeTabComponent,
@@ -116,9 +112,6 @@ describe('NewHomeTabComponent', () => {
     );
 
     const component = fixture.componentInstance;
-    const alert = (message) => {
-      return {type: type, message: message}
-    }
 
     const alertService = TestBed.inject(AlertService);
     const alertServiceSpy = spyOn(alertService, 'addAlert').and.callThrough();
@@ -145,8 +138,6 @@ describe('NewHomeTabComponent', () => {
       getByRole,
       getByLabelText,
       routerSpy,
-      alertService,
-      alert,
     };
   };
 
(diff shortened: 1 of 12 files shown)


frontend/src/app/shared/directives/new-home-tab/new-home-tab.directive.ts
```diff
@@ -175,7 +175,9 @@ export class NewHomeTabDirective implements OnInit, OnDestroy, OnChanges {
     this.setBenchmarksCard();
     this.subscriptions.add();
 
-    this.alertMessage = history.state?.alertMessage;
+    this.alertService.alert$.subscribe((alert) => {
+      this.alertMessage = history.state?.alertMessage === undefined ? alert.message : history.state?.alertMessage;
+    }),
```
duncanc19, line 180: Do we need this extra logic? I thought the addAlert function just creates the banner on the next page so you wouldn't need to call addAlert again

frontend/src/app/features/workplace/view-my-workplaces/view-my-workplaces.component.ts
```diff
@@ -59,7 +59,9 @@ export class ViewMyWorkplacesComponent implements OnInit, OnDestroy {
     this.setupServerErrorsMap();
     this.setSearchIfPrevious();
 
-    this.alertMessage = history.state?.alertMessage;
+    this.alertService.alert$.subscribe((alert) => {
+      this.alertMessage = alert.message;
+    }),
```
duncanc19, line 64: I'm a bit confused by this, when we call addAlert, the alert is set and rendered in the alert component. Previously we were adding the alert here by checking the state and calling addAlert if there was a message, but now it should just be handled by the other component which is now setting the alert. 

I think we should be able to remove this and the sendAlert function if we're not passing any alerts through state anymore.

frontend/src/app/shared/directives/new-home-tab/new-home-tab.directive.ts
duncanc19: Same with this one as comment above

frontend/src/app/features/new-dashboard/home-tab/home-tab.component.spec.ts
```diff
@@ -381,17 +393,16 @@ describe('NewHomeTabComponent', () => {
         });
 
         it('should show an alert banner after requesting to link to a parent', async () => {
-          const { component, fixture, alertServiceSpy } = await setup();
+          const message = `You've sent a link request`;
+
+          const { component, fixture, alertServiceSpy, alertService, alert } = await setup();
 
+          spyOnProperty(alertService, 'alert$', 'get').and.returnValue(of(alert(message)));
```
duncanc19, line 400: Could maybe move this line into the setup function as it's in a few places, then we could maybe just have something like this so we can remove a few more lines:
`spyOnProperty(alertService, 'alert$', 'get').and.returnValue(of({ type:  'success', message: message }));`

duncanc19 (approved):

PR #6205, opened 2024-04-30
Feature/about parents updated

ssrome:
#### Work done
- Add 'What you can do as a parent workplace page' with CMS content
- PreviousRouteService created to capture previous url
- Add MockPreviousRouteService
- Fix following issues on parent snagging list:
  - When you ‘become a parent’ the ‘what you can do as a parent workplace’ link in the green banner generates a 404 error.
  - After you ‘become a parent’, when you navigate to the ‘your other workplaces’ page, there is a ‘what you can do as a parent workplace’ link that doesn’t do anything when you click on it.

#### Tests
- [x] Yes


frontend/src/app/features/workplace/view-my-workplaces/view-my-workplaces.component.ts
```diff
@@ -44,7 +46,10 @@ export class ViewMyWorkplacesComponent implements OnInit, OnDestroy {
     private route: ActivatedRoute,
     private router: Router,
     private alertService: AlertService,
-  ) {}
+    private previousRouteService: PreviousRouteService,
+  ) {
+    this.isParent = establishmentService.primaryWorkplace.isParent;
+  }
```
duncanc19, line 52: Is there a case where they'd be on this page if the workplace isn't a parent?

frontend/src/app/shared/components/new-workplace-summary/workplace-summary.component.ts
```diff
@@ -44,6 +45,8 @@ export class NewWorkplaceSummaryComponent implements OnInit, OnDestroy {
     private cqcStatusChangeService: CqcStatusChangeService,
     private tabsService: TabsService,
   ) {
+    this.isParent = establishmentService.primaryWorkplace.isParent;
```
duncanc19, line 48: Because we're taking workplace as an input and the rest of the component uses that data, should we use isParent from workplace instead?

frontend/src/app/features/pages/about-parents/about-parents.component.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  ngOnInit(): void {
+    this.workplace = this.establishmentService.primaryWorkplace;
+    this.pages = this.route.snapshot.data.pages?.data[0];
+    this.setPreviousPageText();
+    this.breadcrumbService.show(this.showJourneyType(), this.workplace?.name);
+  }
+
+  private setPreviousPageText() {
+    if(this.previousRouteService.getPreviousPage() != "about parents") {
```
duncanc19, line 39: I think this/the getPreviousPage function could maybe be simplified a bit. With this function we could set a local variable and check it before updating previousPageText. For getPreviousPage, I'm wondering if we get much benefit from removing the dashes because we're still just checking against a string here

frontend/src/app/features/workplace/view-my-workplaces/view-my-workplaces.component.ts
ssrome: This has now been removed

frontend/src/app/shared/components/new-workplace-summary/workplace-summary.component.ts
ssrome: This has now been updated to use from workplace

frontend/src/app/features/pages/about-parents/about-parents.component.ts
ssrome: Updated the getPreviousPage function so it won't remove the dashes

duncanc19 (approved):

PR #6221, opened 2024-05-10
Removed setting primary workplace in sub view and update benefits bundle to use establishment

duncanc19:
The way have been setting state in the establishment class for subsidiary workplaces is different from the original way of having primary workplace which is the logged in workplace, and establishment, which is set either as the primary workplace or the subsidiary which is being viewed. This created problems with setting permissions and meant that original navigating logic in the question sections was being ignored and then handled by the subsidiary router service. 

#### Work done
- Removed setting primaryWorkplace from parent subsidiary service
- Changed primaryWorkplaceResolver to set establishment as well so that sections outside of question areas can access it from parent view
- Updated benefits bundle area to use establishment instead of primary workplace to display correct name in sub view

#### Tests
- [ ] Yes


Jonopono123 (approved):
Good stuff, glad to see we're changing instances of `primaryWorkplace` to `establishment` as well, should make things a bit more modular

PR #6222, opened 2024-05-10
Feat/add nhs bsa api spec endpoint

duncanc19:
#### Work done
- Added NHS BSA API spec to the codebase
- Created get endpoint to retrieve the API spec

#### Tests
- [X] Yes

### DIFF
--- a/backend/server/test/unit/routes/cqcStatusCheck/index.spec.js
+++ b/backend/server/test/unit/routes/cqcStatusCheck/index.spec.js
@@ -31,7 +31,7 @@ const locationIDs = [
   { locationID: '1-6114402008', mainService: 'Community healthcare services' },
 ];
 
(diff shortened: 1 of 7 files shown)


backend/server/test/unit/routes/cqcStatusCheck/index.spec.js
Jonopono123, line 1: Is there a reason we're skipping this one?

duncanc19: These tests are failing because you now need to pass a secret key to the CQC API. We don't have a way of storing that API key outside of Github/making it accessible in the build pipeline at the moment so it'll need a bit of work and I'd say it's better to have the rest of the tests passing for now.

PR #6243, opened 2024-05-21
Feat/home page workplaces dropdown

duncanc19:
#### Work done
- Created navigateToWorkplaceDropdown component which shows when in parent sub view
- Moved back to parent link logic into dropdown component and have it render when in sub view and more than 30 workplaces
- Added boolean param to getChildWorkplaces endpoint so you can return with or without pending workplaces

#### Tests
- [X] Yes


backend/server/routes/establishments/childWorkplaces.js
```diff
@@ -45,6 +46,10 @@ const formatChildWorkplaces = (childWorkplaces) => {
   });
 };
 
+const convertToBoolean = (getPendingWorkplaces) => {
+  return getPendingWorkplaces === 'false' ? false : true;
```
Jonopono123, line 50: Should this default to true in the event of a typo?

Jonopono123 (commented):
Just one minor question

backend/server/routes/establishments/childWorkplaces.js
Jonopono123: In that case I have no issue with it, makes sense why we do it this way. I'll approve the PR!

Jonopono123 (approved):

PR #6258, opened 2024-06-04
added employed from inside uk question

Jonopono123:
#### Work done
- Added employed from inside/outside uk question

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/classes/worker/workerProperties.js
+++ b/backend/server/models/classes/worker/workerProperties.js
@@ -1,5 +1,6 @@
 // encapsulates all properties of a worker, by returning a PropertyManager
 const Manager = require('../properties/manager');
+const { EmployedFromInsideUkProperty } = require('./properties/employedFromInsideUk');
 
 // individual properties
 const contractProperty = require('./properties/contractProperty').WorkerContractProperty;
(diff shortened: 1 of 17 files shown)


backend/server/models/classes/worker/workerProperties.js
```diff
@@ -1,5 +1,6 @@
 // encapsulates all properties of a worker, by returning a PropertyManager
 const Manager = require('../properties/manager');
+const { EmployedFromInsideUkProperty } = require('./properties/employedFromInsideUk');
```
duncanc19, line 3: Think we can remove this line

frontend/src/app/features/workers/employed-from-inside-uk/employed-from-inside-uk.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (!employedFromInsideUk) {
+      return null;
+    }
+
+    return {
+      employedFromInsideUk,
+    };
+  }
+
+  private determineConditionalRouting(): string[] {
```
duncanc19, line 60: I _think_ because this page will always go to main-job-start-date inside the flow, we can just have the line on 45 and we don't need this function. I think this function is needed either when routing is dependent on the response or when there's a follow on page which should navigated to even when on the staff record (outside flow)

frontend/src/app/features/workers/employed-from-outside-uk/employed-from-outside-uk.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (!employedFromOutsideUk) {
+      return null;
+    }
+
+    return {
+      employedFromOutsideUk,
+    };
+  }
+
+  private determineConditionalRouting(): string[] {
```
duncanc19, line 60: Did you have a check to see if this can be removed? For me I think it must be because it's already getting set to main-job-start-date in the init

ssrome (approved):

PR #6259, opened 2024-06-05
Feat/1442 international recruitment edit page

ssrome:
#### Work done
-  Display the health and visa and employed from outside or inside UK questions in the employment details section within the staff record page
- Added International recruitment service to hold the logic for the correct routing for the health and visa and employed from outside or inside UK questions or to show on staff record. Service also used to transform the employed from outside or inside UK values

Employment details

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/core/services/international-recruitment.service.spec.ts
+++ b/frontend/src/app/core/services/international-recruitment.service.spec.ts
@@ -23,23 +23,23 @@ describe('InternationalRecruitmentService', () => {
 
   describe('employed from outside or inside UK values', () => {
     it('should convert the yes value', () => {
-      worker.employedFromOutsideUk = 'Yes';
-      expect(service.convertEmployedFromOutsideUkValue(worker)).toBe('Outside the UK');
+      const employedFromOutsideUk = 'Yes';
+      expect(service.convertEmployedFromOutsideUkValue(employedFromOutsideUk)).toBe('Outside the UK');
     });
 
     it('should convert the no value', () => {
-      worker.employedFromOutsideUk = 'No';
-      expect(service.convertEmployedFromOutsideUkValue(worker)).toBe('Inside the UK');
+      const employedFromOutsideUk = 'No';
+      expect(service.convertEmployedFromOutsideUkValue(employedFromOutsideUk)).toBe('Inside the UK');
     });
 
     it("should convert the don't know value", () => {
-      worker.employedFromOutsideUk = "Don't know";
-      expect(service.convertEmployedFromOutsideUkValue(worker)).toBe('Not known');
+      const employedFromOutsideUk = "Don't know";
+      expect(service.convertEmployedFromOutsideUkValue(employedFromOutsideUk)).toBe('I do not know');
     });
 
     it('should convert the null value', () => {
-      worker.employedFromOutsideUk = null;
-      expect(service.convertEmployedFromOutsideUkValue(worker)).toBe(null);
+      const employedFromOutsideUk = null;
+      expect(service.convertEmployedFromOutsideUkValue(employedFromOutsideUk)).toBe(null);
     });
   });
 
--- a/frontend/src/app/core/services/international-recruitment.service.spec.ts
+++ b/frontend/src/app/core/services/international-recruitment.service.spec.ts
@@ -1,0 +1,114 @@
+import { TestBed } from '@angular/core/testing';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { InternationalRecruitmentService } from './international-recruitment.service';
+import { workerBuilder } from '@core/test-utils/MockWorkerService';
+import { Worker } from '@core/model/worker.model';
+
+describe('InternationalRecruitmentService', () => {
+  let service: InternationalRecruitmentService;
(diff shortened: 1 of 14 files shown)


frontend/src/app/core/services/international-recruitment.service.spec.ts
```diff
@@ -0,0 +1,114 @@
+import { TestBed } from '@angular/core/testing';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { InternationalRecruitmentService } from './international-recruitment.service';
+import { workerBuilder } from '@core/test-utils/MockWorkerService';
+import { Worker } from '@core/model/worker.model';
+
+describe('InternationalRecruitmentService', () => {
+  let service: InternationalRecruitmentService;
+  let worker = workerBuilder() as Worker;
```
duncanc19, line 10: With this, the same worker object is going to be used in each test. That works fine as it stands because you're always updating the same fields, but if you updated another field in one of the first tests it could have a knock on effect. Maybe we could have `let worker` here and then set it to workerBuilder in the beforeEach, which would make sure the tests are isolated?

frontend/src/app/core/services/international-recruitment.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor() {}
+
+  public convertEmployedFromOutsideUkValue(employedFromOutsideUk) {
+    return employedFromOutsideUk === 'Yes'
+      ? 'Outside the UK'
+      : employedFromOutsideUk === 'No'
+      ? 'Inside the UK'
+      : employedFromOutsideUk === "Don't know"
+      ? 'I do not know'
+      : null;
```
duncanc19, line 15: I find nested ternaries a bit confusing, maybe it could be a case statement or I quite like just having a look up object, but is maybe just personal preference. This is the kind of thing I mean:
`const mappings = { 'Yes':  'Outside the UK' }
return mappings[employedFromOutsideUk]`

frontend/src/app/features/workers/employed-from-outside-uk/employed-from-outside-uk.component.ts
```diff
@@ -7,16 +7,17 @@ import { EstablishmentService } from '@core/services/establishment.service';
...
   templateUrl: './employed-from-outside-uk.component.html',
 })
 export class EmployedFromOutsideUkComponent extends QuestionComponent {
   public answersAvailable = [
-    { tag: 'Outside the UK', value: 'Yes' },
-    { tag: 'Inside the UK', value: 'No' },
-    { tag: 'I do not know', value: `Don't know` },
+    { tag: this.internationalRecruitmentService.convertEmployedFromOutsideUkValue('Yes'), value: 'Yes' },
+    { tag: this.internationalRecruitmentService.convertEmployedFromOutsideUkValue('No'), value: 'No' },
+    { tag: this.internationalRecruitmentService.convertEmployedFromOutsideUkValue(`Don't know`), value: `Don't know` },
```
duncanc19, line 20: Or as an alternative we could just have a mappings object in the service rather than a conversion function? Where each answer has databaseValue and serviceValue or something?

frontend/src/app/shared/components/staff-record-summary/employment/employment.component.spec.ts
```diff
@@ -13,18 +13,21 @@ import { SharedModule } from '@shared/shared.module';
...
       imports: [SharedModule, RouterModule, RouterTestingModule, HttpClientTestingModule],
       declarations: [SummaryRecordChangeComponent],
       providers: [
+        InternationalRecruitmentService,
         {
           provide: PermissionsService,
           useFactory: MockPermissionsService.factory(['canEditWorker']),
           deps: [HttpClient],
         },
+        //{ provide: InternationalRecruitmentService },
```
duncanc19, line 30: Commented out code

frontend/src/app/core/services/international-recruitment.service.spec.ts
ssrome: The variable assignment has been moved into the beforeEach

frontend/src/app/shared/components/staff-record-summary/employment/employment.component.spec.ts
ssrome: Now removed

frontend/src/app/core/services/international-recruitment.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    this._employedFromOutsideUkMappings['No'].questionValues,
+    this._employedFromOutsideUkMappings[`Don't know`].questionValues,
+  ];
+
+  public getEmployedFromOutsideUkAnswers() {
+    return this._employedFromOutsideUkAnswers;
+  }
+
+  public getEmployedFromOutsideUkStaffRecordValue(employedFromOutsideUkDbValue) {
+    return this._employedFromOutsideUkMappings[employedFromOutsideUkDbValue]?.staffRecordValue;
```
duncanc19, line 49: Rather than the null section in the mappings object, could we just have `|| null` here?

ssrome: Added suggestion

PR #6274, opened 2024-06-13
Feat/1440 international recruitment existing worker new page question 1

ssrome:
#### Work done
- Add new page to show health and care visa question for all existing workers
- Add new route to only get workers that should be asked the health and care visa question
- Add function in internationalrecruitmentservice for answers that navigate to the new employed-from-outside-or-inside-uk page

#### Tests
- [x] Yes


frontend/src/app/features/workplace/health-and-care-visa-existing-workers/health-and-care-visa-existing-workers.component.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (updatedWorkerHealthAndCareVisa.healthAndCareVisa != this.healthCareAndVisaAnswers[answer].value) {
+      this.workersHealthAndCareVisaAnswersToSave.push({
+        uid: updatedWorkerHealthAndCareVisa.uid,
+        healthAndCareVisa: this.healthCareAndVisaAnswers[answer].value,
+      });
+
+      this.updatedWorkers.push({
+        ...updatedWorkerHealthAndCareVisa,
+        healthAndCareVisa: this.healthCareAndVisaAnswers[answer].value,
+      });
```
duncanc19, line 125: Some of this logic might be simplified if we have validation which required the question to be answered for all workers

frontend/src/app/features/workplace/health-and-care-visa-existing-workers/health-and-care-visa-existing-workers.component.ts
```diff
@@ -0,0 +1,182 @@
...
+      this.establishmentService.updateWorkers(this.workplaceUid, this.workersHealthAndCareVisaAnswersToSave).subscribe(
+        (response) => this.navigateToHome(),
+        (error) => this.onSubmitError(error),
+      );
+    }
+  }
+
+  public updateHasWorkersWithHealthAndCareVisa(updatedWorkers) {
+    this.hasWorkersWithHealthAndCareVisa = updatedWorkers.some((worker) => worker.healthAndCareVisa === 'Yes');
+  }
```
duncanc19, line 155: Not a big issue but this is passing in updated workers which is already accessible across the class

duncanc19 (approved):

PR #6291, opened 2024-07-11
Spike/1451 investigate dependabot setup config

ssrome:
#### Work done
- Change dependabot config to target main branch with pr limit of 5
- Grouped patch updates for dependabot to create less PRs
- Bump express from 4.18.2 to 4.19.2 in /frontend
- Bump follow-redirects from 1.15.3 to 1.15.6 in /frontend
- Bump webpack-dev-middleware from 5.3.3 to 5.3.4 in /frontend

#### Tests
- [ ] Yes


frontend/package-lock.json
duncanc19, line 1: This PR isn't supposed to include any version updates is it? I think we might need to revert the changes in the package json lock because we don't want a mismatch between the lock file and the package.json

duncanc19 (approved):

PR #6304, opened 2024-07-26
1471-training-changes-to-the-main-add-a-training-record-page-after-selecting-category

ssrome:
#### Work done
- Show text of category chosen in previous page of add training record
- Add change link to go back to page to change the chosen category
- Redirects back to category page if there's no category id found in training service for add, or params for edit. This will prevent the page being incorrect, on page refresh

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/core/services/training.service.ts
+++ b/frontend/src/app/core/services/training.service.ts
@@ -43,12 +43,6 @@ export class TrainingService {
     );
   }
 
-  getCategoryById(categoryId): Observable<TrainingCategory[]> {
-    return this.http
-      .get<TrainingCategoryResponse>(`${environment.appRunnerEndpoint}/api/trainingCategories/${categoryId}`)
-      .pipe(map((res) => res.trainingCategories));
-  }
-
   public deleteCategoryById(establishmentId, categoryId) {
     return this.http.delete(
       `${environment.appRunnerEndpoint}/api/establishment/${establishmentId}/mandatoryTraining/${categoryId}`,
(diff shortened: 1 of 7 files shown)


frontend/src/app/core/services/training.service.ts
```diff
@@ -15,7 +15,7 @@ export class TrainingService {
   public selectedStaff: Worker[] = [];
   public addMultipleTrainingInProgress$ = new BehaviorSubject<boolean>(false);
   private _trainingOrQualificationPreviouslySelected: string = null;
-  private _trainingCategorySelectedForTrainingRecord: {};
+  private _trainingCategorySelectedForTrainingRecord: any = null;
```
duncanc19, line 18: Could we have a type with id and category and then use it for the param/return types for the getters and setters below?

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.spec.ts
```diff
@@ -359,6 +393,35 @@ describe('AddEditTrainingComponent', () => {
...
+      fixture.detectChanges();
+
+      userEvent.type(getByLabelText('Training name'), 'Some training');
+      userEvent.click(getByLabelText('No'));
+
+      const trainingServiceSpy = spyOn(trainingService, 'clearTrainingCategorySelectedForTrainingRecord');
+      fireEvent.click(getByText('Save record'));
+      fixture.detectChanges();
+
+      fixture.detectChanges();
```
duncanc19, line 420: I think we can remove all the detectChanges except the top one because we're not doing any checks on the rendered page at the end

ssrome: These have now been removed

duncanc19 (approved):

PR #6312, opened 2024-08-09
1463 training add multiple records details

ssrome:
#### Work done
For [Training - add multiple records details](https://trello.com/c/xJGifYFX/1463-training-add-multiple-records-details)
- Show selected category on page and add change link to go back to select category page
- Removed code for selecting a category using dropdown

#### Tests
- [x] Yes


frontend/src/app/shared/directives/add-edit-training/add-edit-training.directive.ts
```diff
@@ -74,12 +75,11 @@ export class AddEditTrainingDirective implements OnInit, AfterViewInit {
...
   public checkForCategoryId(): void {
-    const selectedCategoryId = this.trainingService.getTrainingCategorySelectedForTrainingRecord();
-
+    const selectedCategory = this.trainingService.getTrainingCategorySelectedForTrainingRecord();
     if (this.route.snapshot.queryParamMap.get('trainingCategory')) {
       this.trainingCategory = JSON.parse(this.route.snapshot.queryParamMap.get('trainingCategory'));
-    } else if (selectedCategoryId) {
-      this.trainingCategory = selectedCategoryId;
+    } else if (selectedCategory) {
+      this.trainingCategory = { id: selectedCategory.id, category: selectedCategory.category };
```
duncanc19, line 82: Is there a bit of a mismatch in how trainingCategory is getting set? If it's set through the query params it'll just be the id but here it's an object. It looks like it could cause an error in the onSubmit on 239

frontend/src/app/core/services/training.service.ts
```diff
@@ -105,4 +106,16 @@ export class TrainingService {
   public clearTrainingCategorySelectedForTrainingRecord(): void {
     this._trainingCategorySelectedForTrainingRecord = null;
   }
+
+  public setIsSelectStaffChange(value: boolean): void {
```
duncanc19, line 110: I wasn't too clear on what this means from the name but I'm struggling to come up with something better, how about something along the lines of 
`updatingSelectedStaffForMultipleTraining`?

frontend/src/app/shared/directives/add-edit-training/add-edit-training.directive.ts
ssrome: On line 28, trainingCategory is set as `public trainingCategory: { id: number; category: string };`  so the change on line 82 was to bring it line with that

frontend/src/app/core/services/training.service.ts
ssrome: This has now been updated

frontend/src/app/shared/directives/add-edit-training/add-edit-training.directive.ts
duncanc19: Yeah, so to me it looks like it'll work for this case but if trainingCategory is set on line 80 when it gets it from the query params, it'll be set to an id and not an object. I would've thought it breaks in that case on submit because it'll be a number and not an object with id

duncanc19 (approved):

PR #6317, opened 2024-08-20
1466 job roles changes to the main record summay page

ssrome:
#### Work done
- Move main job role in basic record and add a change link to navigate to the new main job role page
- Moved jobs resolver in wdf routing to stop causing issues viewing that route

staff summary

#### Tests
- [x] Yes


Jonopono123 (approved):
LGTM

PR #6328, opened 2024-08-21
Feat/1468 main job page

kapppa-joe:
#### Work done
- Change main-job-role component to be able to add new staff record
- Add progress bar and error validation to main-job-role
- Set up routing for create-staff-record pages
- Change staff-detail to continue to main-job-role page when adding a new staff record

#### Tests
- [x] Yes


frontend/src/app/core/services/worker.service.ts
```diff
@@ -44,6 +45,7 @@ export class WorkerService {
   public worker$ = this._worker$.asObservable();
   public getRoute$: BehaviorSubject<any> = new BehaviorSubject<any>(null);
   public createStaffResponse = null;
+  private _newWorkerMandatoryInfo: { nameOrId: string; contract: Contracts } = null;
```
duncanc19, line 48: This object type is being used in a couple of places here and in one of the components as well, could we make it an interface to avoid duplicating it?

kapppa-joe: Thanks! I added a `NewWorkerMandatoryInfo` interface for this and refer to it in all usage now

duncanc19 (approved):

PR #6332, opened 2024-08-23
Fix/1495 accordion validation in main job role page

kapppa-joe:
#### Work done
- pass in `hasError` and `errorMessage` to the accordion component in main job role page
- correct an id at grouped-radio-button-accordion, so that the "error summary link focus on the first radio" works for other pages as well
- amend typo: accordi**a**n --> groupedAccordi**o**n. "grouped" is added as we are already using testid=accordion somewhere

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/workers/main-job-role/main-job-role.component.spec.ts
+++ b/frontend/src/app/features/workers/main-job-role/main-job-role.component.spec.ts
@@ -22,7 +22,7 @@ import { Worker } from '@core/model/worker.model';
 import userEvent from '@testing-library/user-event';
 import { AlertService } from '@core/services/alert.service';
 
-fdescribe('MainJobRoleComponent', () => {
+describe('MainJobRoleComponent', () => {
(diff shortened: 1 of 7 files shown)


frontend/src/app/features/workers/main-job-role/main-job-role.component.spec.ts
```diff
@@ -32,71 +32,74 @@ describe('MainJobRoleComponent', () => {
...
+            useValue: {
+              parent: {
+                snapshot: {
+                  url: [{ path }],
+                  data: {
+                    establishment: { uid: 'mocked-uid' },
+                    primaryWorkplace: {},
+                  },
+                },
+              },
```
kapppa-joe, line 70: Joe:
this part looks like a big change, but actually it is almost the same as before. just that the indentation & line break change caused github diff view to show this in an odd way

ssrome (approved):

PR #6333, opened 2024-08-23
Feat/1485 l2 cc new question in staff record flow

ssrome:
#### Work done
- Add migrations for new properties
- Add property to model
- Created page and routing for level 2 care certificate
- Site wide change to increase thickness of grey line on radio conditional input
level 2 care certification question
Year after error

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
+++ b/backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
@@ -2,15 +2,13 @@
 const ChangePropertyPrototype = require('../../properties/changePrototype').ChangePropertyPrototype;
 
 const LEVEL_2_CARE_CERTIFICATE_TYPE = ['Yes, completed', 'Yes, started', 'No'];
-exports.WorkerLevel2CareCertificateProperty = class WorkerLevel2CareCertificateProperty extends (
-  ChangePropertyPrototype
-) {
+exports.Level2CareCertificateProperty = class Level2CareCertificateProperty extends ChangePropertyPrototype {
   constructor() {
     super('Level2CareCertificate');
   }
 
   static clone() {
-    return new WorkerLevel2CareCertificateProperty();
+    return new Level2CareCertificateProperty();
   }
 
   // concrete implementations
--- a/backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
+++ b/backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
@@ -2,13 +2,16 @@
 const ChangePropertyPrototype = require('../../properties/changePrototype').ChangePropertyPrototype;
 
 const LEVEL_2_CARE_CERTIFICATE_TYPE = ['Yes, completed', 'Yes, started', 'No'];
-exports.Level2CareCertificateProperty = class Level2CareCertificateProperty extends ChangePropertyPrototype {
+exports.WorkerLevel2CareCertificateProperty = class WorkerLevel2CareCertificateProperty extends (
+  ChangePropertyPrototype
+) {
   constructor() {
     super('Level2CareCertificate');
+    this._allowNull = true;
   }
 
   static clone() {
-    return new Level2CareCertificateProperty();
+    return new WorkerLevel2CareCertificateProperty();
   }
 
   // concrete implementations
--- a/backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
+++ b/backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
@@ -1,0 +1,60 @@
+// the level 2 care certificate property is an enumeration
+const ChangePropertyPrototype = require('../../properties/changePrototype').ChangePropertyPrototype;
+
+const LEVEL_2_CARE_CERTIFICATE_TYPE = ['Yes, completed', 'Yes, started', 'No'];
+exports.WorkerLevel2CareCertificateProperty = class WorkerLevel2CareCertificateProperty extends (
+  ChangePropertyPrototype
+) {
+  constructor() {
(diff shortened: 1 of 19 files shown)


backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
```diff
@@ -0,0 +1,92 @@
...
+    const yearLevel2CareCertificateIntroduced = 2024;
+    if (document.level2CareCertificate) {
+      if (LEVEL_2_CARE_CERTIFICATE_TYPE.includes(document.level2CareCertificate.value)) {
+        if (document.level2CareCertificate.value === 'Yes, completed') {
+          const thisYear = new Date().getFullYear();
+          if (
+            document.level2CareCertificate.year &&
+            Number.isInteger(document.level2CareCertificate.year) &&
+            document.level2CareCertificate.year >= yearLevel2CareCertificateIntroduced &&
+            document.level2CareCertificate.year < thisYear + 1
```
duncanc19, line 28: Very minor but for me it reads better to have this as less than or equals to make it clearer it's checking whether it's up to this year. Could also maybe have this and the yearLevel2CareCertificateIntroduced variable in a helper function with a descriptive name like `completedInYearSinceLevel2CareCertIntroduced` or something just to make this function a bit shorter

backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
```diff
@@ -0,0 +1,92 @@
...
+  savePropertyToSequelize() {
+    return {
+      Level2CareCertificateValue: this.property.value,
+      Level2CareCertificateYear: this.property.value === 'Yes, completed' ? this.property.year : null,
+    };
+  }
+
+  isEqual(currentValue, newValue) {
+    // Level 2 Care Certificate is a simple (enum'd) string
+    return currentValue && newValue && currentValue === newValue;
```
duncanc19, line 70: I'm not sure if this will work as it's an object with value and year on it?

frontend/src/app/features/workers/level-2-adult-social-care-certificate/level-2-adult-social-care-certificate.component.ts
```diff
@@ -0,0 +1,97 @@
...
+
+@Component({
+  selector: 'app-level-2-adult-social-care-certificate',
+  templateUrl: './level-2-adult-social-care-certificate.component.html',
+})
+export class Level2AdultSocialCareCertificateComponent extends QuestionComponent {
+  public answersAvailable = [
+    { value: 'Yes, completed', tag: 'Yes, completed' },
+    { value: 'Yes, started', tag: 'Yes, started' },
+    { value: 'No', tag: 'No' },
```
duncanc19, line 19: As these all have the value and tag as exactly the same, does it make sense to just have it as an array? I think the object is useful for mapping if there's a difference between the database and how it's displayed in the service but I'm not sure it adds much in this case

ssrome: This has been updated to a simple array

backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
ssrome: Moved part of it into a function, as suggested

backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  isEqual(currentValue, newValue) {
+    // not a simple (enum'd) string compare; if "Yes, completed", also need to compare the year (just an integer)
+    let yearEqual = false;
+    if (currentValue && newValue && currentValue.value === 'Yes, completed') {
+      if (currentValue.year && newValue.year && currentValue.year === newValue.year) yearEqual = true;
+    } else {
+      yearEqual = true;
+    }
```
duncanc19, line 80: I think this will always set yearEqual to true, I think we maybe want to do a check for a mismatch, so set `yearEqual` to false in the case where either `currentValue.year` or `newValue.year` exists and they are not equal, otherwise it's true

duncanc19: And yearEqual will always be false in the case that the value isn't 'Yes, completed', which means the function would always return false then

backend/server/models/classes/worker/properties/level2CareCertificateProperty.js
```diff
@@ -0,0 +1,100 @@
...
+          }
+        } else {
+          this.property = {
+            value: document.level2CareCertificate.value,
+          };
+        }
+      }
+    } else {
+      this.property = null;
+    }
```
kapppa-joe, line 55: the long if-else branch here might be not so easy to follow, would it be good if we handle each case of "value" separately and do early return instead?
for example something like:

```javascript
    const isValidYear = (year) => {
      return Number.isInteger(year) && completedInYearSinceLevel2CareCertIntroduced(year);
    };

    if (['Yes, started', 'No'].includes(document.level2CareCertificate.value)) {
      this.property = { value: document.level2CareCertificate.value, year: null };
      return;
    }

    if (document.level2CareCertificate.value === 'Yes, completed') {
      if (isValidYear(document.level2CareCertificate.year)) {
        this.property = {
          value: document.level2CareCertificate.value,
          year: document.level2CareCertificate.year,
        };
      } else {
        this.property = { value: document.level2CareCertificate.value, year: null };
      }
      return;
    }

    // if none of above match
    this.property = null;
```

duncanc19 (approved):

PR #6349, opened 2024-09-05
Feat/1487 quals groups

Jonopono123:
#### Work done
- Modified the casing and wording on existing quals
- Added new quals
- Removed unused quals group
- Added custom getter to workerAvailableQualification model (Side note: this really needs renaming at some point, we should be able to refer to models predictably but this doesn't match our schema in the slightest)
- Added BUDI mappings for new quals
- Fixed typo in getAvailableQualifications function name

#### Tests
- [ ] Yes


backend/migrations/20240903124901-addNewQuals.js
```diff
@@ -0,0 +1,68 @@
...
+        ],
+        "Any other qualification": [
+          { id: 148, code: 154, title: "Functional skills in english", level: '1' },
+          { id: 149, code: 155, title: "Functional skills in english", level: '2' },
+          { id: 150, code: 156, title: "Functional skills in maths", level: '1' },
+          { id: 151, code: 157, title: "Functional skills in maths", level: '2' },
+        ]
+      }
+
+      for (const key of Object.keys(newQuals)) {
```
duncanc19, line 33: Fairly minor as it's a migration, but could we change `key` to something descriptive like `qualificationType` and then instead of row have `qualification`?

backend/server/models/workerAvailableQualifications.js
```diff
@@ -22,7 +22,22 @@ module.exports = function(sequelize, DataTypes) {
     title: {
       type: DataTypes.TEXT,
       allowNull: false,
-      field: '"Title"'
+      field: '"Title"',
+      get() {
```
duncanc19, line 26: Pretty nifty having the conversion on the model! Seems like something we might want tests for though, could we have it as a function in a utils file and import it here to make it easy to test?

backend/migrations/20240903124901-addNewQuals.js
```diff
@@ -0,0 +1,68 @@
...
+          { id: 144, code: 150, title: "Oral health", level: '2' },
+        ],
+        "Degree": [
+          { id: 145, code: 151, title: "Occupational therapy degree", level: '6' },
+          { id: 146, code: 152, title: "Physiotherapy degree", level: '6' },
+          { id: 147, code: 153, title: "Any nursing degree", level: '6' },
+        ],
+        "Any other qualification": [
+          { id: 148, code: 154, title: "Functional skills in english", level: '1' },
+          { id: 149, code: 155, title: "Functional skills in english", level: '2' },
```
duncanc19, line 27: English should be capitalised, shouldn't it?

Jonopono123: I can double check with Billy but the spreadsheet has it lower case

backend/server/test/unit/utils/qualificationsUtils.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    describe('level is provided', () => {
+      it('should return the title with the level in brackets', () => {
+        const title = 'This is a qual';
+        const level = '4';
+
+        const returnedValue = formatQualificationTitle(title, level);
+
+        expect(returnedValue).to.deep.equal('This is a qual (level 4)');
+      });
+    })
```
duncanc19, line 26: I think it could do with a test for when the title ends in a bracket to check that it adds the level inside of the brackets

duncanc19 (approved):

ssrome (approved):

PR #6359, opened 2024-09-16
Feat/1506 training certs retrieve a list of all certificates for a training record

ssrome:
#### Work done
- Add training certificates to training record response
- Update training model to include training certificate
- Add certifications-table component
- Add the certifications table to edit training details
Screenshot 2024-09-17 at 10 59 32
Screenshot 2024-09-17 at 10 59 24

#### Tests
- [x] Yes


backend/server/models/classes/training.js
```diff
@@ -603,6 +614,15 @@ class Training extends EntityValidator {
             model: models.workerTrainingCategories,
             as: 'category',
           },
+          {
+            model: models.trainingCertificates,
+            as: 'trainingCertificates',
+            attributes: ['uid', 'filename', 'uploadDate'],
+          },
+        ],
+        order: [
```
duncanc19, line 623: I'm not sure if I'm understanding this right, but isn't one training record returned, which includes an array of training certificates attached to it, so don't we want to sort the certificates and not outside of the include? Does it work if we put the order inside the trainingCertificates object above?

frontend/src/app/shared/components/certifications-table/certifications-table.component.scss
```diff
@@ -0,0 +1,3 @@
+.filename-table-header {
+  width: 420px;
+}
```
duncanc19, line 3: It might be better setting the width as a percentage to make it more dynamic when you adjust the page size

backend/server/models/classes/training.js
ssrome: It didn't sort when I moved the order array into the training certs object. Asked Billy and Rich about sort and it was most recent then alphabetical

frontend/src/app/shared/components/certifications-table/certifications-table.component.scss
ssrome: Change now made

backend/server/models/classes/training.js
duncanc19: Ok, thanks for checking, I reckon we should be able to get rid of the `as: 'trainingCertificates'` bits in the order section still because that's usually just there for referencing the fields from that table in other parts of the query?

ssrome: Has now been removed

duncanc19 (approved):

PR #6361, opened 2024-09-18
Feat/1512 get certificate records in training and quals backend

duncanc19:
#### Work done
- Updated fetch on Training model to return training certificates linked to training records
- Refactored fetch function and added tests

#### Tests
- [X] Yes


Jonopono123 (approved):
LGTM

PR #6364, opened 2024-09-23
Feat/1514 download single and multiple training certificates

duncanc19:
#### Work done
- Created new API endpoint to return a presigned URL for downloading existing certificates on a training record
- Connected Download and Download all buttons on frontend to new endpoint 
- Updated certificate metadata table to store key as well to make retrieval from S3 easier 

#### Tests
- [X] Yes


frontend/src/app/shared/components/certifications-table/certifications-table.component.html
```diff
@@ -9,6 +9,7 @@
             <a href="#" class="govuk-link govuk-link--no-visited-state">
               <img alt="" src="/assets/images/icon-download.svg" class="govuk-util__vertical-align-bottom" /><span
                 class="govuk-!-margin-left-1"
+                (click)="handleDownloadCertificate($event, null)"
```
kapppa-joe, line 12: could we attach this (click) function to the <a> tag rather than the <span> tag?  
right now if user clicked the download icon, the download wouldn't work and they will be directly to dashboard

kapppa-joe (approved):

PR #6365, opened 2024-09-23
Feat/1518 upload cert with new training

kapppa-joe:
#### Work done
- Amend frontend to allow adding new training together with certificate upload
- Add a measure to prevent submit button from being triggered more than once 

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/core/services/worker.service.ts
+++ b/frontend/src/app/core/services/worker.service.ts
@@ -10,7 +10,12 @@ import {
   QualificationsResponse,
   QualificationType,
 } from '@core/model/qualification.model';
-import { MultipleTrainingResponse, TrainingRecordRequest, TrainingResponse } from '@core/model/training.model';
+import {
+  CreateTrainingRecordResponse,
+  MultipleTrainingResponse,
+  TrainingRecordRequest,
+  TrainingResponse,
+} from '@core/model/training.model';
 import { TrainingAndQualificationRecords } from '@core/model/trainingAndQualifications.model';
 import { URLStructure } from '@core/model/url.model';
 import { Worker, WorkerEditResponse, WorkersResponse } from '@core/model/worker.model';
@@ -222,7 +227,7 @@ export class WorkerService {
   }
 
   createTrainingRecord(workplaceUid: string, workerId: string, record: TrainingRecordRequest) {
-    return this.http.post<TrainingRecordRequest>(
+    return this.http.post<CreateTrainingRecordResponse>(
       `${environment.appRunnerEndpoint}/api/establishment/${workplaceUid}/worker/${workerId}/training`,
       record,
     );
@@ -234,7 +239,7 @@ export class WorkerService {
     trainingRecordId: string,
     record: TrainingRecordRequest,
   ) {
-    return this.http.put<TrainingRecordRequest>(
+    return this.http.put<any>(
       `${environment.appRunnerEndpoint}/api/establishment/${workplaceUid}/worker/${workerId}/training/${trainingRecordId}`,
       record,
     );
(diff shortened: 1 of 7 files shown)


frontend/src/app/core/services/worker.service.ts
```diff
@@ -234,7 +239,7 @@ export class WorkerService {
     trainingRecordId: string,
     record: TrainingRecordRequest,
   ) {
-    return this.http.put<TrainingRecordRequest>(
+    return this.http.put<any>(
```
duncanc19, line 242: Is the original interface still valid for this?

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.spec.ts
```diff
@@ -533,6 +555,85 @@ describe('AddEditTrainingComponent', () => {
...
+        userEvent.click(getByLabelText('Yes'));
+        const completedDate = getByTestId('completedDate');
+        userEvent.type(within(completedDate).getByLabelText('Day'), '10');
+        userEvent.type(within(completedDate).getByLabelText('Month'), '4');
+        userEvent.type(within(completedDate).getByLabelText('Year'), '2020');
+        const expiresDate = getByTestId('expiresDate');
+        userEvent.type(within(expiresDate).getByLabelText('Day'), '10');
+        userEvent.type(within(expiresDate).getByLabelText('Month'), '4');
+        userEvent.type(within(expiresDate).getByLabelText('Year'), '2027');
+        userEvent.type(getByLabelText('Notes'), 'Some notes added to this training');
```
duncanc19, line 588: I'd be tempted to get rid of some of these inputs (maybe expiry and completed?) to make the test a bit shorter, as the important part is the two calls to the service being made

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
```diff
@@ -189,8 +191,8 @@ export class AddEditTrainingComponent extends AddEditTrainingDirective implement
     if (this.trainingRecordId) {
       trainingRecordId = this.trainingRecordId;
     } else {
-      // TODO: this is the case of adding new training with certificate
-      // extract trainingRecordId from trainingRecordResponse
+      const { uid } = trainingRecordResponse as CreateTrainingRecordResponse;
+      trainingRecordId = uid;
```
duncanc19, line 195: I think you might be able to make this all a one liner by using an or but will need to confirm it works -  
`const trainingRecordId = this.trainingRecordId || trainingRecordResponse.uid`

frontend/src/app/shared/directives/add-edit-training/add-edit-training.directive.ts
```diff
@@ -49,6 +49,7 @@ export class AddEditTrainingDirective implements OnInit, AfterViewInit {
   public showChangeLink: boolean = false;
   public multipleTrainingDetails: boolean;
   public trainingCertificates: TrainingCertificate[];
+  public disableSubmitButton: boolean = false;
```
duncanc19, line 52: Having the verb first maybe makes it sound like a function and a bit harder to tell what corresponds to true and false, could name it `submitButtonDisabled`?

frontend/src/app/core/services/worker.service.ts
kapppa-joe: The original interface here probably wasn't quite right. The type in <> should denote the response object type, not the request object type. 
When I checked in the local apps, the response we get here seems to change depending on what fields are filled in, so I put this as <any> for the moment

duncanc19 (approved):

PR #6368, opened 2024-09-25
Feat/1511 training certs remove certificate from a training record

ssrome:
#### Work done
- Add endpoint to delete training certificates from database and S3 bucket
- Add call to delete training certificates on save for training record

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/trainingCertificates.js
+++ b/backend/server/models/trainingCertificates.js
@@ -87,5 +87,13 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
+  TrainingCertificates.countCertificatesToBeDeleted = async function (uids) {
+    return await this.count({
+      where: {
+        uid: uids,
+      },
+    });
+  };
+
   return TrainingCertificates;
 };
--- a/backend/server/models/trainingCertificates.js
+++ b/backend/server/models/trainingCertificates.js
@@ -79,5 +79,13 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
+  TrainingCertificates.deleteCertificate = async function (key) {
+    return await this.destroy({
+      where: {
+        key: key,
+      },
+    });
+  };
+
   return TrainingCertificates;
 };
--- a/backend/server/models/trainingCertificates.js
+++ b/backend/server/models/trainingCertificates.js
@@ -79,10 +79,10 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
-  TrainingCertificates.deleteCertificate = async function (key) {
+  TrainingCertificates.deleteCertificate = async function (uid) {
     return await this.destroy({
       where: {
-        key: key,
+        uid: uid,
       },
     });
   };
--- a/backend/server/models/trainingCertificates.js
+++ b/backend/server/models/trainingCertificates.js
@@ -79,10 +79,10 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
-  TrainingCertificates.deleteCertificate = async function (uid) {
+  TrainingCertificates.deleteCertificate = async function (uids) {
     return await this.destroy({
       where: {
-        uid: uid,
+        uid: uids,
       },
     });
   };
--- a/backend/server/routes/establishments/workerCertificate/s3.js
+++ b/backend/server/routes/establishments/workerCertificate/s3.js
@@ -1,4 +1,10 @@
-const { PutObjectCommand, GetObjectCommand, S3Client, HeadObjectCommand } = require('@aws-sdk/client-s3');
+const {
+  PutObjectCommand,
+  GetObjectCommand,
+  S3Client,
+  HeadObjectCommand,
+  DeleteObjectsCommand,
+} = require('@aws-sdk/client-s3');
 const { getSignedUrl } = require('@aws-sdk/s3-request-presigner');
 
 const config = require('../../../config/config');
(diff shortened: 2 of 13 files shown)


backend/server/models/trainingCertificates.js
```diff
@@ -79,5 +79,13 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
+  TrainingCertificates.deleteCertificate = async function (key) {
+    return await this.destroy({
+      where: {
+        key: key,
+      },
+    });
```
kapppa-joe, line 87: Could we use the certificate uid instead of key as the where condition here?

backend/server/routes/establishments/workerCertificate/s3.js
```diff
@@ -34,6 +40,21 @@ const getSignedUrlForDownload = ({ bucket, key, options }) => {
   return getSignedUrl(s3Client, getCommand, options);
 };
 
+async function deleteCertificatesFromS3({ bucket, objects }) {
+  const s3client = getS3Client();
```
duncanc19, line 44: I think it's meant to be better to create an S3 client once and then reuse it across all S3 calls, so you can just use the S3 client in the global scope here. It means you don't have to create a client for every single call so it's more efficient and I think there are other benefits like the SDK managing things like connection pooling.

backend/server/routes/establishments/workerCertificate/trainingCertificate.js
```diff
@@ -124,11 +124,70 @@ const getPresignedUrlForCertificateDownload = async (req, res) => {
...
+    try {
+      deleteFromDb = await models.trainingCertificates.deleteCertificate(fileKey);
+    } catch (error) {
+      console.log(error);
+      errors.push({ key: fileKey, error });
+    }
+
+    if (deleteFromDb) {
+      filesToDeleteFromS3.push({ Key: fileKey });
+    } else if (deleteFromDb !== null) {
```
duncanc19, line 152: I'm a bit confused by this, is it for the case it's undefined? Otherwise I'm not sure when it'd enter this block

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
```diff
@@ -226,6 +242,14 @@ export class AddEditTrainingComponent extends AddEditTrainingDirective implement
     return { uid: certificate.uid, filename: certificate.filename };
   }
 
+  private deleteTrainingCertificate(files) {
+    this.trainingService
+      .deleteCertificates(this.establishmentUid, this.workerId, this.trainingRecordId, files)
+      .subscribe((res) => {
+        console.log(res);
```
duncanc19, line 249: Console log

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
```diff
@@ -25,6 +25,8 @@ export class AddEditTrainingComponent extends AddEditTrainingDirective implement
   public establishmentUid: string;
   public workerId: string;
   private _filesToUpload: File[];
+  public uploadFilesErrors: string[] | null;
```
duncanc19, line 28: I think this might be a remnant from pulling in from the feature branch, this got renamed to `certificateErrors`

frontend/src/app/core/services/training.service.ts
```diff
@@ -244,4 +245,16 @@ export class TrainingService {
       confirmUploadRequestBody,
     );
   }
+
+  public deleteCertificates(
+    workplaceUid: string,
+    workerUid: string,
+    trainingUid: string,
+    filesToDelete: Array<any>,
```
kapppa-joe, line 253:
```suggestion
    filesToDelete: TrainingCertificate[],
```
from the usage, seems that `filesToDelete` have this type

backend/server/routes/establishments/workerCertificate/trainingCertificate.js
ssrome: On delete the database was returning either 1 or 0. The 0 was when the record wasn't there so I wanted some way to capture that

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
ssrome: Now removed

backend/server/routes/establishments/workerCertificate/s3.js
ssrome: Now removed

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
ssrome: Removed as `certificateErrors` was also there

backend/server/routes/establishments/workerCertificate/trainingCertificate.js
duncanc19: Ah ok, might be better to check on 1 and 0 and could potentially have one liners to make it clear what it's doing? Something along the lines of 
```
    const successfullyDeletedFromTable = (res) => res === 1;
const failedToDeleteFromTable = (res) => res === 0;
```

backend/server/routes/establishments/workerCertificate/trainingCertificate.js
```diff
@@ -124,11 +124,70 @@ const getPresignedUrlForCertificateDownload = async (req, res) => {
...
+  for (const file of filesToDelete) {
+    let deleteFromDb = null;
+    let fileKey = makeFileKey(id, workerId, trainingUid, file.uid);
+
+    try {
+      deleteFromDb = await models.trainingCertificates.deleteCertificate(fileKey);
+    } catch (error) {
+      console.log(error);
+      errors.push({ key: fileKey, error });
+    }
```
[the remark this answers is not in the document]
duncanc19: Sounds good!

duncanc19 (approved):

PR #6373, opened 2024-10-02
Feat/1510 training certs remove certificates when deleting a training record

ssrome:
#### Work done
- Added backend functionality to delete the training certificates from the database and S3 on deletion of a training record

#### Tests
- [x] Yes


frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
```diff
@@ -242,10 +242,10 @@ export class AddEditTrainingComponent extends AddEditTrainingDirective implement
     return { uid: certificate.uid, filename: certificate.filename };
   }
 
-  private deleteTrainingCertificate(files) {
+  private deleteTrainingCertificate(files: any) {
```
duncanc19, line 245: I think this type can be `TrainingCertificate[]` because it's coming from `filesToRemove` which is just items pushed from `trainingCertificates`

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
```diff
@@ -242,10 +242,10 @@ export class AddEditTrainingComponent extends AddEditTrainingDirective implement
...
     this.trainingService
       .deleteCertificates(this.establishmentUid, this.workerId, this.trainingRecordId, files)
-      .subscribe((res) => {});
+      .subscribe(() => {});
```
duncanc19, line 248: I've realised we've got a few subscribe calls in this component which aren't being added to the subscriptions, so the subscriptions will hang when the component is destroyed. Would you mind just adding them in this PR please?

duncanc19 (approved):

frontend/src/app/features/training-and-qualifications/add-edit-training/add-edit-training.component.ts
ssrome: Now updated

PR #6376, opened 2024-10-03
Feat/1520 qualification accordion

kapppa-joe:
#### Work done
- Add a select-qualification-type component
- Store the selected qualification at QualificationService
- Patch a minor bug at GroupedRadioButtonAccordionComponent

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/training-and-qualifications/add-edit-qualification/select-qualification-type/select-qualification-type.component.ts
+++ b/frontend/src/app/features/training-and-qualifications/add-edit-qualification/select-qualification-type/select-qualification-type.component.ts
@@ -20,10 +20,6 @@ export class SelectQualificationTypeComponent implements OnInit {
   public submitted: boolean = false;
   public error = false;
   public formErrorsMap: Array<ErrorDetails>;
-
-  public title: string;
-  public section: string;
-  public submitButtonText: string;
   public preFilledId: number;
 
   public worker: Worker;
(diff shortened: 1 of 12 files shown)


frontend/src/app/features/training-and-qualifications/add-edit-qualification/select-qualification-type/select-qualification-type.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  constructor(
+    protected formBuilder: FormBuilder,
+    protected qualificationService: QualificationService,
+    protected router: Router,
+    protected backLinkService: BackLinkService,
+    protected workerService: WorkerService,
+    protected route: ActivatedRoute,
+    protected errorSummaryService: ErrorSummaryService,
+  ) {}
```
duncanc19, line 62: I think it's a bit unusual to have the ngOnInit above the constructor, or at least I don't think we do it in this codebase much!

duncanc19: Also, I think these classes can be private, protected is usually used for either directives or components which extend a directive

frontend/src/app/features/training-and-qualifications/add-edit-qualification/select-qualification-type/select-qualification-type.component.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  protected setTitle(): void {
+    this.title = 'Select the type of qualification you want to add';
+  }
```
duncanc19, line 66: With a few of these setter functions, I'm wondering if we get much benefit from them as they're not dynamic, the text could just go directly in the html template

Jonopono123 (approved):

frontend/src/app/features/training-and-qualifications/add-edit-qualification/select-qualification-type/select-qualification-type.component.ts
kapppa-joe: Thanks!  Not sure why I put it like that. Now ngOnInit is below the constructor.
Also changed the protected vars to private

duncanc19 (approved):

PR #6377, opened 2024-10-03
Feat/1509 remove missing training

Jonopono123:
#### Work done
- Removed missing mandatory training rows from mandatory training table
- Added new message and links in place of table if no mandatory training exists

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/training-and-qualifications/new-training-qualifications-record/new-training-and-qualifications-record.component.html
+++ b/frontend/src/app/features/training-and-qualifications/new-training-qualifications-record/new-training-and-qualifications-record.component.html
@@ -206,30 +206,7 @@
     </li>
   </ul>
 </div>
-
-<div *ngIf="currentFragment === fragmentsObject.allRecords">
-  <div class="govuk-grid-column-full">
-    <app-new-training
-      [trainingCategories]="mandatoryTraining"
-      [trainingType]="'Mandatory training'"
-      [isMandatoryTraining]="true"
-      [canEditWorker]="canEditWorker"
-      [setReturnRoute]="setReturnRoute"
-      [missingMandatoryTraining]="missingMandatoryTraining.length > 0"
-    >
-    </app-new-training>
-    <app-new-training
-      [trainingCategories]="nonMandatoryTraining"
-      [trainingType]="'Non-mandatory training'"
-      [canEditWorker]="canEditWorker"
-      [setReturnRoute]="setReturnRoute"
-    >
-    </app-new-training>
-    <app-new-qualifications [canEditWorker]="canEditWorker" [qualificationsByGroup]="qualificationsByGroup">
-    </app-new-qualifications>
-  </div>
-</div>
-<div class="govuk-grid-column-full" *ngIf="currentFragment === fragmentsObject.mandatoryTraining">
+<div class="govuk-grid-column-full">
   <app-new-training
     *ngIf="currentFragment === fragmentsObject.allRecords || currentFragment === fragmentsObject.mandatoryTraining"
     [trainingCategories]="mandatoryTraining"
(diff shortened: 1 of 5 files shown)


frontend/src/app/features/training-and-qualifications/new-training-qualifications-record/new-training-and-qualifications-record.component.html
```diff
@@ -207,14 +207,37 @@ <h1 class="govuk-heading-l govuk-!-margin-bottom-3">Training and qualifications<
...
+    <app-new-training
+      [trainingCategories]="nonMandatoryTraining"
+      [trainingType]="'Non-mandatory training'"
+      [canEditWorker]="canEditWorker"
+      [setReturnRoute]="setReturnRoute"
+    >
+    </app-new-training>
+    <app-new-qualifications [canEditWorker]="canEditWorker" [qualificationsByGroup]="qualificationsByGroup">
+    </app-new-qualifications>
+  </div>
```
duncanc19, line 230: It looks like you've reintroduced some duplication which got removed on the feature branch, each app-new-training now just has a check for being allRecords or whichever section so you don't have to have it twice

frontend/src/app/features/training-and-qualifications/new-training-qualifications-record/new-training-and-qualifications-record.component.ts
```diff
@@ -146,30 +148,38 @@ export class NewTrainingAndQualificationsRecordComponent implements OnInit, OnDe
...
 
-    this.getLastUpdatedDate([this.qualificationsByGroup?.lastUpdated, trainingRecords?.lastUpdated]);
+    this.getLastUpdatedDate([this.qualificationsByGroup?.lastUpdated, this.trainingRecords?.lastUpdated]);
   }
 
-  private createBlankMissingMandatoryTrainings(mandatoryTraining: TrainingRecordCategory[]) {
+  private setMandatoryTraining() {
+    this.mandatoryTraining = this.trainingRecords.mandatory;
+    this.sortTrainingAlphabetically(this.mandatoryTraining);
+    this.createBlankMissingMandatoryTrainings(this.mandatoryTraining);
```
duncanc19, line 167: I think it's a bit unclear that this is setting `missingMandatoryTraining`, maybe we could do something like rename `createBlankMissingMandatoryTrainings` to `setMissingMandatoryTraining` and call it from the `setTraining` function like we do with this function (`setMandatoryTraining`)

frontend/src/app/features/training-and-qualifications/new-training-qualifications-record/new-training/new-training.component.spec.ts
```diff
@@ -273,6 +261,36 @@ describe('NewTrainingComponent', async () => {
...
       expect(noMandatoryTrainingLink).toBeFalsy();
     });
+
+    it('should display a no mandatory training for job role message when mandatory training is not required for the job role', async () => {
+      const { component, fixture } = await setup();
+      component.trainingCategories = [];
+      component.isMandatoryTraining = true;
+      component.workplaceUid = '123';
+      component.missingMandatoryTraining = false;
+      fixture.detectChanges();
```
kapppa-joe, line 271:
```suggestion
      component.ngOnChanges();
      fixture.detectChanges();
```
For now, this test and the test below seems not passing
Probably we can call ngOnChanges() here to make them pass, because manually editing the variables of the component doesn't seem to trigger the onChange hook

Another way would be to amend the setup() method and add parameters for inputs like `trainingCategories`, `isMandatoryTraining` and `missingMandatoryTraining`. That takes more work but probably reflect the actual data flow better

duncanc19: I'd say go for the setup option because it better reflects how the component is used, also I'm not sure why I hadn't thought of passing in an override object for setup functions in tests when I'd suggested the same idea for our mock services, we should definitely start doing this!

duncanc19 (approved):

PR #6387, opened 2024-10-16
Feat/1526 quals certs backend

Jonopono123:
#### Work done
- Added qualifications certificates endpoint
- Refactored common certificate functionality into new `WorkerCertificateService` class
- Refactored delete training functionality to remove duplicate code

#### Tests
- [x] Yes


backend/server/routes/establishments/qualification/index.js
```diff
@@ -197,5 +198,6 @@ router.route('/available').get(hasPermission('canViewWorker'), availableQualific
 router.route('/:qualificationUid').get(hasPermission('canViewWorker'), viewQualification);
 router.route('/:qualificationUid').put(hasPermission('canEditWorker'), updateQualification);
 router.route('/:qualificationUid').delete(hasPermission('canEditWorker'), deleteQualification);
+router.use('/:trainingUid/certificate', QualificationCertificateRoute);
```
kapppa-joe, line 201:
```suggestion
router.use('/:qualificationUid/certificate', QualificationCertificateRoute);
```

backend/server/routes/establishments/workerCertificate/qualificationCertificate.js
```diff
@@ -0,0 +1,90 @@
...
+
+  return {
+    files: formatFilesArg(req.body),
+    params: {
+      establishmentUid: req.params.id,
+      workerUid: req.params.workerId,
+      recordUid: req.params.qualificationUid,
+    }
+  };
+};
```
kapppa-joe, line 28: I would suggest to avoid doing this format request process, and prefer to just extract the necessary args and then pass into the service method.

Some reasons for this :
- body.filesToDownload / body.filesToDelete probably isn't neccessary. In current implementation, any request body to backend is having the shape { files: [] }
- having a nested object as input args makes it not easy to see what is being passed in and what is info is needed at service level

backend/server/routes/establishments/workerCertificate/workerCertificateService.js
```diff
@@ -0,0 +1,206 @@
...
+    for (const file of files) {
+      const signedUrl = await s3.getSignedUrlForDownload({
+        bucket: certificateBucket,
+        key: this.makeFileKey(establishmentUid, workerUid, recordUid, file.uid),
+        options: { expiresIn: downloadSignedUrlExpire },
+      });
+      responsePayload.push({ signedUrl, filename: file.filename });
+    }
+
+    return { files: responsePayload };
```
kapppa-joe, line 141:
```suggestion
    return responsePayload;
```
the endpoint controller have `res.status(200).json({ files: responsePayload })` so we don't need to wrap it here

backend/server/models/qualificationCertificates.js
```diff
@@ -0,0 +1,99 @@
...
+  };
+
+  QualificationCertificates.addCertificate = function ({ qualificationRecordId, workerFk, filename, fileId, key }) {
+    const timeNow = dayjs().format();
+
+    return this.create({
+      uid: fileId,
+      workerFk: workerFk,
+      workerQualificationsFk: qualificationRecordId,
+      filename: filename,
```
kapppa-joe, line 76:
```suggestion
  QualificationCertificates.addCertificate = function ({ recordId, workerFk, filename, fileId, key }) {
    const timeNow = dayjs().format();

    return this.create({
      uid: fileId,
      workerFk: workerFk,
      workerQualificationsFk: recordId,
      filename: filename,
```

To match the param name that being passed in

backend/server/models/trainingCertificates.js
```diff
@@ -66,13 +66,15 @@ module.exports = function (sequelize, DataTypes) {
     });
   };
 
-  TrainingCertificates.addCertificate = function ({ trainingRecordId, workerFk, filename, fileId, key }) {
+  TrainingCertificates.addCertificate = function ({ recordId, workerFk, filename, fileId, key }) {
     const timeNow = dayjs().format();
 
+    console.log(recordId);
+
```
kapppa-joe, line 73:
```suggestion

```

backend/server/routes/establishments/workerCertificate/qualificationCertificate.js
```diff
@@ -0,0 +1,90 @@
...
+
+const requestUploadUrl = async (req, res) => {
+  const certificateService = initialiseCertificateService();
+
+  const request = formatRequest(req);
+
+  try {
+    const responsePayload = await certificateService.requestUploadUrl(request);
+    return res.status(200).json({ files: responsePayload });
```
kapppa-joe, line 37: I think maybe better to use different method names at controller level (line 30) and service level (line 36)
Having the same method name in two related piece makes it difficult to trace the flow and impact readability / maintainability. 
For example, if I am new to this code base and I try search for the name `requestUploadUrl`, it would be not easy to tell which one is the method actually being called
Same for the other method names here as well

backend/server/routes/establishments/workerCertificate/qualificationCertificate.js
Jonopono123: The main reason I added this was to change `qualificationUid`/`trainingUid` to `recordUid` as well as changing the `id` param to `establishmentUid`. I figured that `req` is essentially a nested input already so it wouldn't be too much different in principle to what we were already doing.

kapppa-joe (approved):

PR #6389, opened 2024-10-17
Feat/1542 funding stand alone main funding page changes

ssrome:
#### Work done
- Update the wdf overview page for standalone and parents
- Create wdf summary panel that uses the styling from the summary section
- Update wdf journey

#### Tests
- [x] Yes


frontend/src/app/core/test-utils/MockReportService.ts
```diff
@@ -1,10 +1,26 @@
...
 
 @Injectable()
 export class MockReportService extends ReportService {
+  private _wdf = {
+    overall: true,
+    workplace: true,
+    staff: true,
+  };
+
+  public static factory(wdfObject: any) {
```
duncanc19, line 16: Shall we go with the overrides object way, so if we want to mock something else later we can just pass it in with the object? 
`overrides: any = {}`
It also makes it a bit clearer in the tests because then you have to specify what you're setting, e.g. `MockReportService.factory({ _wdf: 'data' })`

frontend/src/app/features/wdf/wdf-data-change/wdf-overview/wdf-overview.component.spec.ts
```diff
@@ -5,27 +5,56 @@ import { BreadcrumbService } from '@core/services/breadcrumb.service';
...
-        { provide: BreadcrumbService, useClass: MockBreadcrumbService },
-        { provide: EstablishmentService, useClass: MockEstablishmentService },
-        { provide: ReportService, useClass: MockReportService },
-      ],
-    });
+  const currentYear = new Date().getFullYear();
+
+  const defaulWdfEligibilityStatus = { overall: true, workplace: true, staff: true };
+
+  const setup = async (eligibilityStatus = {}, isParent = false, parentOverallWdfEligibility = false) => {
```
duncanc19, line 21: And we could use the overrides object here too

frontend/src/app/features/wdf/wdf-data-change/wdf-overview/wdf-overview.component.spec.ts
```diff
@@ -44,142 +73,157 @@ describe('WdfOverviewComponent', () => {
...
-      const { component, fixture, getByText } = await setup();
-      const viewWorkplacesLink = 'View your workplaces';
-      const viewWorkplacesSentence = 'and keep your data up to date to save time next year';
+    it('should not display the funding requirements inset text when requirements are met', async () => {
+      const wdfStatus = { overall: true, workplace: true, staff: true };
 
-      component.isParent = true;
-      component.parentOverallWdfEligibility = true;
-      fixture.detectChanges();
+      const { queryByTestId } = await setup(wdfStatus, true, true);
```
duncanc19, line 198: Here's an example of why the overrides object is a bit nicer, it's quite hard to tell what these booleans represent, and it means you don't have to set every field for the cases when you want to change the third param but want the default for the first two

frontend/src/app/shared/components/wdf-summary-panel/wdf-summary-panel.component.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor(private router: Router) {}
+
+  ngOnInit(): void {
+    this.getSections();
+  }
+
+  ngOnChanges(): void {
+    this.getSections();
+  }
```
duncanc19, line 28: Just want to confirm we need this, what cases are there where the data will have changed without having navigated from another page and the onInit being run?

duncanc19: I guess it could happen with the WDF field confirmations on the workplace page?

ssrome: This component was rendering before all the data had been passed. Had to put this so it would show correctly once   the component it's sitting in had finished rendering

frontend/src/app/shared/components/wdf-summary-panel/wdf-summary-panel.component.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+  let isWorkplaceAParent = false;
+  let isOverallWdfEligibile = false;
+
+  const setup = async (
+    workplaceWdfEligibilityStatus = isWorkplaceWdfEligibile,
+    staffWdfEligibilityStatus = isStaffWdfEligibile,
+    parentOverallWdfEligibility = isParentOverallWdfEligibile,
+    isParent = isWorkplaceAParent,
+    overallWdfEligibility = isOverallWdfEligibile,
+  ) => {
```
duncanc19, line 30: This does something similar to the overrides object approach, but I think there's maybe a bit more risk here because there could be side effects when these variables are updated in tests, so if you update one field in one test and then the next test doesn't update that field, I'd expect it to have whatever was set in the previous test rather than the initialisation which could cause a bit of confusion if it causes later tests to fail

frontend/src/app/shared/components/wdf-summary-panel/wdf-summary-panel.component.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+      imports: [RouterTestingModule, HttpClientTestingModule, BrowserModule, SharedModule],
+      providers: [],
+      componentProperties: {
+        workplaceWdfEligibilityStatus: workplaceWdfEligibilityStatus,
+        staffWdfEligibilityStatus: staffWdfEligibilityStatus,
+        wdfStartDate: `1 April ${currentYear}`,
+        wdfEndDate: `31 March ${currentYear + 1}`,
+        isParent: isParent,
+        parentOverallWdfEligibility: parentOverallWdfEligibility,
+        overallWdfEligibility: overallWdfEligibility,
```
duncanc19, line 41: Just to explain properly, in this case all of the overrides are in this `componentProperties` object, so at the end you can use the spread operator for overrides, and it will replace any fields that you pass in in the object:
```
componentProperties: {
        workplaceWdfEligibilityStatus: workplaceWdfEligibilityStatus,
        staffWdfEligibilityStatus: staffWdfEligibilityStatus,
        wdfStartDate: `1 April ${currentYear}`,
        wdfEndDate: `31 March ${currentYear + 1}`,
        isParent: isParent,
        parentOverallWdfEligibility: parentOverallWdfEligibility,
        overallWdfEligibility: overallWdfEligibility,
        ...overrides,
```

duncanc19 (approved):

PR #6391, opened 2024-10-18
Feat/1538 funding find out more new page

ssrome:
#### Work done
- Add new wdf learn more page with CMS content to be shown
- Add additional filter condition to only get published pages
- Update wdf overview to link to new page

#### Tests
- [x] Yes


frontend/src/app/features/wdf/wdf-data-change/wdf-learn-more/wdf-learn-more.component.ts
duncanc19, line 1: It's tricky because there are so many references to WDF in the codebase, but I think this page/component name shouldn't have wdf in it because it's supposed to explain all the funds which are available to users

frontend/src/app/features/wdf/wdf-data-change/wdf-learn-more/wdf-learn-more.component.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  ngOnInit(): void {
+    this.breadcrumbService.show(JourneyType.WDF);
+    this.workplace = this.establishmentService.establishment;
+    this.pages = this.route.snapshot.data.pages?.data[0];
+  }
+
+  public viewWdfMainPage(): void {
+    this.router.navigate(['/wdf']);
+  }
```
duncanc19, line 32: Do we get any benefit out of having this as a function rather than just a router link?

ssrome: I've changed it. Think it was just a case of normally having to write a function for a button

duncanc19 (approved):

PR #6403, opened 2024-11-04
Feat/1549 bulk upload retain existing qualifications

kapppa-joe:
#### Work done
- Add a helper class to manage qualification changes in bulk upload
- On UPDATE, modify existing qualification records instead of always deleting and creating again
- On DELETE, delete any certificates record first before deleting the qualification record, so that we don't trigger fk violation errors.
- When downloading or deleting certificates, use the key from database instead of calling `makeFileKey` all the time

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/classes/helpers/bulkUploadQualificationHelper.js
+++ b/backend/server/models/classes/helpers/bulkUploadQualificationHelper.js
@@ -17,13 +17,13 @@ class BulkUploadQualificationHelper {
   async processQualificationsEntities(qualificationsEntities) {
     const promisesToReturn = [];
 
-    const allExistingQualifications = await models.workerQualifications.findAll({
+    const allQualificationRecords = await models.workerQualifications.findAll({
       where: {
         workerFk: this.workerId,
       },
     });
 
-    const existingQualificationFks = allExistingQualifications.map((record) => record.qualificationFk);
+    const existingQualificationFks = allQualificationRecords.map((record) => record.qualificationFk);
     const bulkUploadQualificationFks = compact(qualificationsEntities.map((entity) => entity?._qualification?.id));
 
     for (const bulkUploadEntity of qualificationsEntities) {
@@ -32,7 +32,7 @@ class BulkUploadQualificationHelper {
 
       if (qualificationExists) {
         console.log('modify current qualification: ', bulkUploadEntity._qualification.title);
-        const recordToUpdate = allExistingQualifications.find(
+        const recordToUpdate = allQualificationRecords.find(
           (record) => record.qualificationFk === currentQualificationId,
         );
         promisesToReturn.push(this.updateQualification(recordToUpdate, bulkUploadEntity));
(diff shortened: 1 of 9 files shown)


backend/server/models/classes/helpers/bulkUploadQualificationHelper.js
```diff
@@ -0,0 +1,86 @@
...
+    entityFromBulkUpload.workerId = this.workerId;
+    entityFromBulkUpload.workerUid = this.workerUid;
+    entityFromBulkUpload.establishmentId = this.establishmentId;
+
+    return entityFromBulkUpload.save(this.savedBy, this.bulkUploaded, 0, this.externalTransaction);
+  }
+
+  updateQualification(existingRecord, entityFromBulkUpload) {
+    const fieldsToUpdate = {
+      source: this.bulkUploaded ? 'Bulk' : 'Online',
```
duncanc19, line 62: It feels odd having checks like this on a helper specifically for BU. I think it makes sense to just say 'Bulk' here.

backend/server/models/classes/helpers/bulkUploadQualificationHelper.js
```diff
@@ -0,0 +1,86 @@
+const { compact } = require('lodash');
+
+const models = require('../../index');
+const WorkerCertificateService = require('../../../routes/establishments/workerCertificate/workerCertificateService');
+
+class BulkUploadQualificationHelper {
+  constructor({ workerId, workerUid, establishmentId, savedBy, bulkUploaded = true, externalTransaction }) {
```
duncanc19, line 7: I can see the value in having the `bulkUploaded` field to make it clear what's being passed into `save` below but is it better to take away the destructuring and just set it to true on line 12?

backend/server/models/classes/helpers/bulkUploadQualificationHelper.js
```diff
@@ -0,0 +1,86 @@
...
+
+    for (const bulkUploadEntity of qualificationsEntities) {
+      const currentQualificationId = bulkUploadEntity?._qualification?.id;
+      const qualificationExists = existingQualificationFks.includes(currentQualificationId);
+
+      if (qualificationExists) {
+        const recordToUpdate = allQualificationRecords.find(
+          (record) => record.qualificationFk === currentQualificationId,
+        );
+        promisesToReturn.push(this.updateQualification(recordToUpdate, bulkUploadEntity));
```
duncanc19, line 37: I've been trying to think of how to simplify this because it's quite a big function which is doing a lot. If we create a function for finding an existing qualification, I think we could separate it out a bit more and remove the existingQualificationFks and includes check:
```
const existingQualification = findExistingQualification(...);
if (existingQualification) {
  promisesToReturn.push(this.updateQualification(recordToUpdate, bulkUploadEntity))
} else ...
```

backend/server/routes/establishments/workerCertificate/workerCertificateService.js
```diff
@@ -62,7 +83,31 @@ class WorkerCertificateService {
...
+      attributes: ['id', 'workerFk'],
+    });
+
+    if (!record) {
+      throw new HttpError(`Failed to find related ${this.recordType} record`, 400);
+    }
+    return record;
+  }
+
+  verifyParentRecordExists = this.findParentRecord;
```
duncanc19, line 110: Why is it parent record? Is this being used as a verification step to check the certificate record exists before download and deletion?

duncanc19 (approved):

PR #6404, opened 2024-11-04
Feat/1544 delete qualification certificate on workplace deletion

Jonopono123:
#### Work done
- Refactored Training cert deletion work to bring it inline with quals changes
- Added Quals cert deletion functionality 

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/classes/worker.js
+++ b/backend/server/models/classes/worker.js
@@ -28,6 +28,7 @@ const JSON_DOCUMENT_TYPE = require('./worker/workerProperties').JSON_DOCUMENT;
 const SEQUELIZE_DOCUMENT_TYPE = require('./worker/workerProperties').SEQUELIZE_DOCUMENT;
 
 const TrainingCertificateRoute = require('../../routes/establishments/workerCertificate/trainingCertificate');
+const WorkerCertificateService = require('../../routes/establishments/workerCertificate/workerCertificateService');
 
 // WDF Calculator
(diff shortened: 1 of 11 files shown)


backend/server/routes/establishments/training/index.js
```diff
@@ -189,6 +189,9 @@ const deleteTrainingRecord = async (req, res) => {
     }
   } catch (err) {
     console.error(err);
+    if (err.statusCode && err.message) {
+      return res.status(err.statusCode).send(err.message);
+    }
```
duncanc19, line 194: I think we shouldn't be sending the actual error message back to the user, it potentially gives users an insight into the inner workings of our codebase which we don't really want, and we've got the logging for debugging purposes. Maybe we could just return an internal server error here?

backend/server/models/classes/worker.js
```diff
@@ -554,6 +555,7 @@ class Worker extends EntityValidator {
       // there is no change audit on qualifications; simply delete all that is there and recreate
       if (this._qualificationsEntities && this._qualificationsEntities.length > 0) {
         // delete all existing training records for this worker
+        await this.deleteAllQualificationCertificatesAssociatedWithWorker(externalTransaction);
```
kapppa-joe, line 558: Probably we don't need to add this line for now, because the code in this if block is to do with the bulk upload logic of deleting all existing quals and recreating them again. In the bulk upload change ticket, we will be removing this if block soon

duncanc19 (approved):

PR #6412, opened 2024-11-11
Bulk upload - Transfer staff record

kapppa-joe:
#### Work done
- Amend bulk upload worker file logic to handle a "TRANSFERSTAFFRECORD" column, which should indicate the new workplace that a worker should move into. The feature is meant to replace CHGSUB
- Add validation for this column to screen out cases that could cause issue (e.g. same unique worker id)
- Change bulkupload lambda runtime ver to nodejs18.x (as serverless no longer allows nodejs12.x)

#### Tests
- [x] Yes


backend/server/models/classes/worker.js
```diff
@@ -547,7 +566,7 @@ class Worker extends EntityValidator {
           currentTrainingRecord.workerId = this._id;
           currentTrainingRecord.workerUid = this._uid;
           currentTrainingRecord.establishmentId = this._establishmentId;
-          newTrainingPromises.push(currentTrainingRecord.save(savedBy, bulkUploaded, 0, externalTransaction));
+          newTrainingPromises.push(currentTrainingRecord.save(savedBy, bulkUploaded, externalTransaction));
```
kapppa-joe, line 569: I took out the `0,` here, as Training class save() method actually only have 3 input arguments. The transaction object should be the 3rd one

PR #6414, opened 2024-11-11
Feat/1540 funding staff record tab changes

ssrome:
#### Work done
- Update the staff tab as part of the funding redesigns, which includes adding pagination and search
- Moved part of the code into a directive as some of the functionality is the same as the main staff summary
- Renamed the requirements state component and updated it

#### Tests
- [x] Yes


frontend/src/app/features/wdf/wdf-data-change/wdf-data/wdf-data.component.html
```diff
@@ -39,8 +46,9 @@ <h1 class="govuk-heading-l govuk-!-margin-bottom-4">
     *ngIf="viewWDFData && workers && workplace"
     [workplace]="workplace"
     [workers]="workers"
-    [canEditWorker]="canEditWorker"
     [standAloneAccount]="standAloneAccount"
+    [workerCount]="workerCount"
+    [wdfView]="true"
```
duncanc19, line 51: Looks like in the `wdf-staff-summary` component `wdfView` is getting set to true and not using this input, I think it makes sense not to have it as an input as it's a WDF component so should just be able to remove this

frontend/src/app/features/wdf/wdf-data-change/wdf-data/wdf-data.component.ts
```diff
@@ -40,6 +43,10 @@ export class WdfDataComponent implements OnInit {
   public standAloneAccount = false;
   private subscriptions: Subscription = new Subscription();
   public viewWDFData = false;
```
duncanc19, line 45: Out of scope for this ticket, but just putting here as a reminder to myself/the team, when the tabs get updated if we're still using this way of toggling tabs this definitely needs to be renamed, this is being used to display the workplace tab

duncanc19 (approved):

PR #6416, opened 2024-11-18
Feat/1552 job role groups to vacancies in workplace flow first page

ssrome:
#### Work done
- Add 'do you have vacancies' page
- Update urls in workplace and wdf workplace to point to new page
- Update workplace submit button to take an extra input

#### Tests
- [x] Yes


frontend/src/app/features/workplace/do-you-have-vacancies/do-you-have-vacancies.component.spec.ts
```diff
@@ -0,0 +1,393 @@
...
+
+    const { component } = await setup(overrides);
+
+    expect(component.previousRoute).toEqual(['/workplace', `${component.establishment.uid}`, 'service-users']);
+  });
+
+  describe('prefill form', () => {
+    it('should not prefill the form', async () => {
+      const { component, fixture } = await setup();
+      component.establishment.vacancies = null;
```
duncanc19, line 117: I think you could pass in vacancies with the overrides and then it looks like it's the third param in `MockEstablishmentService.factory`, then you wouldn't have to set things in the component and detect changes in these tests

duncanc19 (approved):

PR #6417, opened 2024-11-19
fix/routing-on-vacancy-first-page

ssrome:
#### Work done
- Fix routing issue when selecting yes for vacancies outside flow so it goes to the second page

#### Tests
- [x] Yes


frontend/src/app/features/workplace/do-you-have-vacancies/do-you-have-vacancies.component.spec.ts
```diff
@@ -345,7 +345,19 @@ describe('DoYouHaveVacanciesComponent', () => {
       expect(getByText('Cancel')).toBeTruthy();
     });
 
-    it('should navigate to the summary page when submitting', async () => {
+    it("should navigate to the select vacancy job roles page when submitting 'Yes'", async () => {
```
duncanc19, line 348: This has got the same test name as the one changed above and it sits within the same describe block. Does setting returnUrl to false make it the add workplace details flow and true means it's coming from the summary page? Could we separate out any summary version tests into a separate describe block to make it a bit clearer?

duncanc19 (approved):

PR #6422, opened 2024-11-20
Feat/1560 job role groups to starters in workplace flow first page

ssrome:
#### Work done
- Create new page for do you have starters
- Move code into directive so page 1 of vacancies and starters can use shared code and html
- Update workplace summary links to go to new page 

#### Tests
- [x] Yes


frontend/src/app/features/workplace/do-you-have-starters/do-you-have-starters.component.spec.ts
```diff
@@ -0,0 +1,429 @@
...
+      fixture,
+      getByText,
+      getAllByText,
+      getByLabelText,
+      getByTestId,
+      queryByText,
+      queryAllByText,
+      queryByTestId,
+      getByRole,
+    } = await render(DoYouHaveStartersComponent, {
```
duncanc19, line 27: Jon mentioned recently and I'm not sure why we hadn't thought of it, you can just set this to a variable and then de-structure it which saves us having to duplicate all of these and just use whatever we need from the testing functions, e.g.
```
const setupTools = await render(...)
```
And at the bottom of the setup function
```
return { ...setupTools, component, establishmentService, establishmentServiceSpy, routerSpy }
```

frontend/src/app/features/workplace/do-you-have-starters/do-you-have-starters.component.spec.ts
```diff
@@ -0,0 +1,429 @@
...
+
+      component.init();
+
+      const form = component.form;
+
+      expect(form.value).toEqual({ startersLeaversVacanciesKnown: 'With Jobs' });
+    });
+
+    it("should preselect 'Yes' if hasStarters is true return to page if the database has a different value", async () => {
+      const overrides = { returnUrl: false, starters: jobOptionsEnum.NONE };
```
duncanc19, line 179: I find this test name a bit confusing, is it supposed to go to the previous page if there's a mismatch between local storage and the database? I can't see that in the test code

frontend/src/app/shared/directives/do-you-have-starters-leavers-vacancies/do-you-have-starters-leavers-vacancies.directive.ts
```diff
@@ -0,0 +1,125 @@
...
+    this.hasStartersLeaversOrVacancies = localStorage.getItem(this.localStorageKey) === 'true' ? true : false;
+  }
+
+  protected setToLocalStorage(): void {}
+
+  protected getDataToPrefill(): void {}
+
+  protected prefillForm(): void {
+    this.getDataToPrefill();
+    this.getFromLocalStorage();
```
duncanc19, line 73: I find this a little bit hard to understand, it's not super clear that `hasStartersLeaversOrVacancies` is getting set by `getFromLocalStorage`. I wonder if it might be clearer to have these two functions return data and set them to the variables here, and I wonder if we could make it clearer what they represent. Is `dataToPrefill` the current state in the database? Is it from local storage when you go back from one of the following pages?

duncanc19 (approved):

PR #6444, opened 2024-12-06
pause the feature for optional other job roles field

kapppa-joe:
#### Work done
- undo every change in backend related to optional job role name
- remove the functionality of showing a optional job role checkbox in job role selection page

#### Tests
- [ ] Yes


duncanc19 (commented):
Looks good to me but it seems a shame to reverse some of the cleaning up changes which came in with the original work, we don't know if this will be coming back in so I think it makes sense to keep the clean up

backend/server/models/classes/establishment/properties/jobHelper.js
duncanc19, line 1: Linting changes in this file

backend/server/models/classes/establishment/properties/vacanciesProperty.js
```diff
@@ -37,6 +37,7 @@ exports.VacanciesProperty = class VacanciesProperty extends ChangePropertyProtot
 
   restorePropertyFromSequelize(document) {
     if (document.VacanciesValue && document.VacanciesValue === 'With Jobs' && document.jobs) {
+      //console.log("WA DEBUG - all establishment jobs: ", document.jobs)
       // we're only interested in Vacancy jobs
```
duncanc19, line 41: This getting added back in

duncanc19 (approved):

PR #6454, opened 2024-12-12
Feat/1575 all or selected job roles

duncanc19:
#### Work done
- Created '_Which job roles need this training?_' page for new mandatory training flow
- Updated previous page '_Select the training category that you want to make mandatory_' to filter out categories which already have mandatory training

#### Tests
- [X] Yes

### DIFF
--- a/frontend/src/app/features/training-and-qualifications/add-mandatory-training/all-or-selected-job-roles/all-or-selected-job-roles.component.spec.ts
+++ b/frontend/src/app/features/training-and-qualifications/add-mandatory-training/all-or-selected-job-roles/all-or-selected-job-roles.component.spec.ts
@@ -19,6 +19,14 @@ import { AllOrSelectedJobRolesComponent } from './all-or-selected-job-roles.comp
 describe('AllOrSelectedJobRolesComponent', () => {
   async function setup(overrides: any = {}) {
     const establishment = establishmentBuilder();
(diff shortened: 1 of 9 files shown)


frontend/src/app/features/training-and-qualifications/add-mandatory-training/all-or-selected-job-roles/all-or-selected-job-roles.component.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+              overrides.selectedTraining !== undefined
+                ? overrides.selectedTraining
+                : {
+                    trainingCategory: {
+                      category: 'Activity provision, wellbeing',
+                      id: 1,
+                      seq: 0,
+                      trainingCategoryGroup: 'Care skills and knowledge',
+                    },
+                  },
```
kapppa-joe, line 40: Might be just personal preference but this part looks a bit too deeply nested to me. 
Would it be good if we to move the selectedTraining part outside?
something like:
```
const selectedTraining = overrides.selectedTraining ?? {trainingCategory: ...}
...
useValue: {selectTraining, resetState: () => {}}
```
outside of

kapppa-joe (approved):
Looks good! Gave a minor suggestion but I think it's fine to go as is

PR #6465, opened 2025-01-07
Fix/Login page - should set focus on username field when invalid username/password error message is clicked

kapppa-joe:
#### Work done
- when error message of invalid username/password is clicked, set focus on the username input box

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/login/login.component.html
+++ b/frontend/src/app/features/login/login.component.html
@@ -7,7 +7,7 @@
 >
 </app-error-summary>
 
-<form #formEl novalidate (ngSubmit)="onSubmit()" [formGroup]="form" id="server-error">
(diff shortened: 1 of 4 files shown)


frontend/src/app/features/login/login.component.html
```diff
@@ -38,6 +38,7 @@ <h1 class="govuk-fieldset__heading" data-cy="signin-heading">Sign in</h1>
           {{ getFormErrorMessage('username', 'atSignInUsername') }}
         </ng-container>
       </span>
+      <span id="server-error" *ngIf="serverError"></span>
```
kapppa-joe, line 41: This empty element act as a starting point for ErrorSummary component to locate the input box to focus on. 
In order for the logic to work, an element with id "server-error" must be under the same `<div>` with the input box.

duncanc19 (approved):

PR #6469, opened 2025-01-08
Feat/1589 data quality log in message

duncanc19:
#### Work done
- Created new page to ask users to keep vacancies and turnover info up to date
- Added logic to navigate to that page if user has not viewed it in over six months
- Created new column on user table and endpoint to update it to track the last time users have visited the page 

#### Tests
- [X] Yes


backend/server/routes/accounts/user.js
```diff
@@ -911,6 +911,27 @@ const swapEstablishment = async (req, res) => {
...
+  } catch (error) {
+    return res.status(500).send('Failed to update last viewed date');
+  }
+};
+
+const isCorrectlyFormattedUid = (userUid) => {
+  const uuidRegex = /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/;
+  return uuidRegex.test(userUid.toUpperCase());
+};
+
```
kapppa-joe, line 934: not a big issue but it might good to use the validate method in 'uuid' package rather than making our own version

PR #6482, opened 2025-01-23
Feat/1583 parent home page message

Jonopono123:
#### Work done
- Added flags to summary panel and other workplaces screen when subs need attention

#### Tests
- [x] Yes


backend/server/test/unit/routes/establishments/childWorkplaces.spec.js
```diff
@@ -155,6 +156,30 @@ describe('server/routes/establishments/childWorkplaces', () => {
...
+      };
+      req = httpMocks.createRequest(request);
+      res = httpMocks.createResponse();
+    });
+
+    afterEach(() => {
+      sinon.restore();
+    });
+
+  })
```
duncanc19, line 181: This is an empty describe, it looks like we might want to add a test to the above describe because that's the function you've changed

frontend/src/app/features/workplace/workplace-info-panel/workplace-info-panel.component.ts
```diff
@@ -35,6 +35,7 @@ export class WorkplaceInfoPanelComponent implements OnInit, OnDestroy {
   public ownershipChangeRequestCreatedByLoggegInUser: boolean;
   public ownershipChangeRequester: any;
   public newHomeDesignParentFlag: boolean;
+  public moveWorkplace: boolean = false;s
```
duncanc19, line 38: It looks like this isn't being used

frontend/src/app/features/workplace/workplace-info-panel/workplace-info-panel.component.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { WindowRef } from '@core/services/window.ref';
+import { MockWorkplaceService } from '../../../core/test-utils/MockWorkplaceService';
+import { DataPermissions, Workplace } from '../../../core/model/my-workplaces.model';
+import { Establishment } from '@core/model/establishment.model';
+import { bool, build, fake, sequence } from '@jackfranklin/test-data-bot';
+import { workplaceTabJourney } from '../../../core/breadcrumb/journey.workplaces';
+import { HttpClient } from '@angular/common/http';
+
+const workplaceBuilder = build('Workplace', {
```
duncanc19, line 27: You could maybe import this from the MockUserService

frontend/src/app/features/workplace/workplace-info-panel/workplace-info-panel.component.ts
Jonopono123: This is required as it's used in the template on lines 22 and 36

Jonopono123: I have however removed the rogue 's' from the end of the line

duncanc19: It looks like you've removed all of it anyway, but I think `moveWorkplace` in the html template is the id of the ng-template which is referenced in the if else, I don't think it was linked to that variable

Jonopono123: I've added it back as the code fails to compile without it due to it not existing

duncanc19 (approved):

PR #6495, opened 2025-02-06
Feat/1602 helpful downloads

Jonopono123:
#### Work done
- Added helpful downloads page to draw content from cms
- Fixed issue with incorrect tab being selected on refresh

#### Tests
- [x] Yes


frontend/src/app/core/resolvers/page.resolver.ts
```diff
@@ -2,6 +2,7 @@ import { Injectable } from '@angular/core';
 import { ActivatedRouteSnapshot, Resolve, Router } from '@angular/router';
 import { Page } from '@core/model/page.model';
 import { PagesService } from '@core/services/pages.service';
+import { slice } from 'lodash';
```
duncanc19, line 5: Looks like this import has been added but isn't needed

duncanc19 (approved):

PR #6502, opened 2025-02-11
Fix/1613 create username password layout

Jonopono123:
#### Work done
- Fixed position of password field and hint text

#### Tests
- [ ] Yes


frontend/src/app/features/create-account/user/username-password/username-password.component.html
```diff
@@ -37,18 +37,18 @@ <h1 class="govuk-fieldset__heading">Create your username and password</h1>
         <div [formGroupName]="'passwordGroup'">
           <div class="govuk-form-group" [class.govuk-form-group--error]="submitted && getPassword.errors">
             <label class="govuk-label" for="createPasswordInput"> Password </label>
-            <span id="createPasswordInput-hint" class="govuk-hint">
-              Must be at least 8 characters long and have uppercase letters, lowercase letters, numbers and special
+            <div id="createPasswordInput-hint" class="govuk-hint">
+              Must be at least 8 characters long and have uppercase letters,<br>lowercase letters, numbers and special
```
duncanc19, line 41: It's a minor thing but does _lowercase_ get moved onto the second line if you put some padding on the right of this div instead? With a break tag, there could be issues for users with different screen sizes, if it wraps before _lowercase_ then it could look like this
```
Must be at least 8 characters long and have uppercase 
letters,
lowercase letters, numbers and special
```

duncanc19 (approved):

PR #6505, opened 2025-02-12
Update angular

duncanc19:
#### Work done
- Updated Angular version from 15 to 17

#### Tests
- [ ] Yes


frontend/src/app/core/services/previous-route.service.ts
```diff
@@ -15,7 +15,7 @@ export class PreviousRouteService {
     this.previousUrl = null;
 
     this.router.events
-      .pipe(filter((event: RouterEvent) => event instanceof NavigationEnd))
+      .pipe(filter((event): event is NavigationEnd => event instanceof NavigationEnd))
```
kapppa-joe, line 18: probably don't need an is guard here

```suggestion
      .pipe(filter((event) => event instanceof NavigationEnd))
```

frontend/src/app/core/test-utils/MockRouter.ts
```diff
@@ -15,3 +15,20 @@ export class MockRouter extends Router {
...
+
+  const mockRouterStateSnapshot: Partial<RouterStateSnapshot> = {
+    url,
+    root: mockActivatedRouteSnapshot as ActivatedRouteSnapshot,
+  };
+
+  const mockRouterState: Partial<RouterState> = {
+    snapshot: mockRouterStateSnapshot as RouterStateSnapshot,
+  };
+
```
kapppa-joe, line 32: maybe we can condense this part a bit, so that it's easier to see how the mockRouterState would look like
```suggestion
  const mockRouterState: Partial<RouterState> = {
    snapshot: { url, root: { url: [] } } as RouterStateSnapshot,
  };
```

kapppa-joe (approved):
Gave a few suggestions but overall looks good. Tried in local machine and confirm all frontend tests are still passing

PR #6510, opened 2025-02-14
fix/order-of-q-and-a-search-results

ssrome:
#### Work done
- Update Q and A search results to place matching titles above matching content

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/help/questions-and-answers/questions-and-answers.component.ts
+++ b/frontend/src/app/features/help/questions-and-answers/questions-and-answers.component.ts
@@ -91,11 +91,20 @@ export class QuestionsAndAnswersComponent implements OnInit {
   public filterTitleAndContent(searchTerm): any[] {
     const searchTermLowerCase = searchTerm.toLowerCase();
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/features/help/questions-and-answers/questions-and-answers.component.ts
```diff
@@ -91,11 +91,20 @@ export class QuestionsAndAnswersComponent implements OnInit {
   public filterTitleAndContent(searchTerm): any[] {
     const searchTermLowerCase = searchTerm.toLowerCase();
 
-    return this.qAndASlugContentAndTitles.filter(
+    let filteredTitles = [];
+    let filteredContent = [];
```
duncanc19, line 95: The filter function returns an empty array if it has no matches so it should be fine to remove these and have `const filteredTitles` and `const filteredContent` below

ssrome: Updated with suggestion

duncanc19 (approved):

PR #6516, opened 2025-02-21
Fix/1619 63634 parent seeing tq links they shouldnt

ssrome:
#### Work done
- Change permission to check the workplace instead of the parent
- Add test file for view-subsidiary-training-and-qualifications component

#### Tests
- [x] Yes


frontend/src/app/features/subsidiary/training-and-qualifications/view-subsidiary-training-and-qualifications.component.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    const override = {
+      withWorkers: false,
+      totalRecords: 0,
+    };
+
+    const { getByTestId, queryByTestId } = await setup(override);
+
+    expect(getByTestId('no-staff-records')).toBeTruthy();
+    expect(queryByTestId('staff-records')).toBeFalsy();
+  });
```
duncanc19, line 116: I don't understand the difference between this and the test above, if they're different cases could we explain the difference in the name? It looks like there's also a duplicate test below. Would we be able to have these tests checking the page content instead of test ids?

ssrome: The duplication has been removed and additional tests have been added

PR #6517, opened 2025-02-27
Feat/1623 cqc parents archived subs

Jonopono123:
#### done
- Filter cqc locations on the contents of the locations table

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/routes/missingCqcProviderLocations/index.js
+++ b/backend/server/routes/missingCqcProviderLocations/index.js
@@ -30,7 +30,7 @@ const missingCqcProviderLocations = async (req, res) => {
     if (provId && provId != 'null' && weeksSinceParentApproval) {
       const childWorkplacesLocationIds = await getChildWorkplacesLocationIds(childWorkplaces.rows);
 
-      const missingCqcLocations = await findMissingCqcLocationIds(this.provId, childWorkplacesLocationIds);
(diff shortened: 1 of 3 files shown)


backend/server/routes/missingCqcProviderLocations/index.js
```diff
@@ -70,12 +66,20 @@ const getChildWorkplacesLocationIds = async (childWorkplaces) => {
...
+const getActiveCQCLocationsForProvider = async (providerId) => {
+  const CQCProviderLocations = await CQCDataAPI.getCQCProviderData(providerId);
+
+  return (await models.location.findMultipleByLocationID(CQCProviderLocations?.locationIds)).map(x => x.locationid);
+}
+
+const findMissingCqcLocationIds = async (provId, childWorkplacesLocationIds) => {
+  const cqcProviderData = await getActiveCQCLocationsForProvider(provId);
+
   let missingCqcLocations = {};
```
duncanc19, line 78: I think we can simplify this function a bit. Could we just filter `cqcProviderData` to remove the matching ones and then return an object instead of creating it and then pushing? Something along the lines of:
```
const missingLocationIds = cqcProviderData.filter((locationId) => !childWorkplacesLocationIds.includes(locationId));

return { missingCqcLocationIds: missingLocationIds, count: missingLocationIds.length }
```

ssrome (approved):

PR #6518, opened 2025-02-28
fix/1628-69572-bu-grouping-errors

ssrome:
#### Work done
- Separate address errors into individual codes for each properties
- Refactor and separate qualification errors
- Add bulk upload lambda related commands to Makefile

#### Tests
- [x] Yes


duncanc19 (approved):

lambdas/bulkUpload/test/unit/classes/workerCSVValidator.spec.js
```diff
@@ -793,6 +826,173 @@ describe('/lambdas/bulkUpload/classes/workerCSVValidator', async () => {
       });
     });
 
+    describe('_validateZeroHourContract', () => {
+      // error = ZERO_HRCONT_ERROR
```
duncanc19, line 830: Some commented out code which I think can probably just be removed. I would normally say that we should test the happy path too but the way this spec file is set up means the happy path is tested for all of the fields as every test is checking all of the validation errors. It's not ideal but I'd say it's out of scope to change the set up here.

ssrome: Has been removed

backend/server/test/unit/routes/establishments/bulkUpload.spec.js
```diff
@@ -520,6 +521,279 @@ describe('/server/routes/establishment/bulkUpload.js', () => {
...
+          BENEFITS: '',
+          SICKPAY: '',
+          PENSION: '',
+          HOLIDAY: '',
+        },
+        2,
+        csvEstablishmentSchemaErrors,
+        [],
+        [
+          {
```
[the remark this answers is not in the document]
ssrome: Now moved into a beforeEach so both tests can use

duncanc19 (approved):

PR #6527, opened 2025-03-10
Fix/1629 training and qualification tab not highlighted

kapppa-joe:
#### Work done
- refactored NewTabsComponent
- when navigating to a page under another section (e.g. staff record, training and qualification), update the current tab in NewTabsComponent

#### Tests
- [x] Yes


frontend/src/app/shared/components/new-tabs/new-tabs.component.ts
```diff
@@ -73,10 +102,15 @@ export class NewTabsComponent implements OnInit, OnDestroy {
     return null;
   }
 
-  public getTabSlugFromSubsidiaryUrl(route: NavigationEnd) {
-    const urlArray = route.urlAfterRedirects.split('/').filter((section) => section.length > 0);
-    if (urlArray.length > 2) {
-      return this.tabs.find((tab) => urlArray[2] === tab.slug);
+  private handleNavigationOfNonDashboardPages(route: NavigationEnd): void {
```
duncanc19, line 105: We've got `route` as the param name for quite a few functions but it's also the name we use for ActivatedRoute, could we rename the params to make it clearer they're unrelated?

frontend/src/app/core/services/tabs.service.ts
```diff
@@ -2,22 +2,43 @@ import { Injectable } from '@angular/core';
...
 
+export const MainDashboardTabs = {
+  homeTab: { title: 'Home', slug: 'home', active: false },
+  workplaceTab: { title: 'Workplace', slug: 'workplace', active: false },
+  staffRecordsTab: { title: 'Staff records', slug: 'staff-records', active: false },
+  tAndQTab: { title: 'Training and qualifications', slug: 'training-and-qualifications', active: false },
+  benchmarksTab: { title: 'Benchmarks', slug: 'benchmarks', active: false },
+};
+
+export const SubsidiaryAccountTabs = {
```
duncanc19, line 19: Could we rename this to something like `SubsidiaryViewTabs`? For me, subsidiary account makes it sound like a sub logging in but this tab shows when a parent is viewing a sub right?

duncanc19 (approved):

PR #6540, opened 2025-03-19
Feat/1641 total number of staff

kapppa-joe:
#### Work done
- Add reusable NumberInputWithButtons component
- Add a UpdateTotalNumberOfStaff component for the new page
- Add `"exclude": ["./cypress.config.ts"]` to frontend `tsconfig.json` to address recent problem of vscode not recognising jasmine syntax in tests

#### Tests
- [x] Yes


frontend/src/app/shared/components/number-input-with-buttons/number-input-with-buttons.component.html
```diff
@@ -0,0 +1,30 @@
+<div class="asc-number-input">
+  <input
```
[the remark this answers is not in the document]
kapppa-joe: false positive as the label of this input is supposed to be out of the component (to allow flexibility in use)

frontend/src/app/features/workers/update-workplace-details-after-staff-changes/update-total-number-of-staff/update-total-number-of-staff.component.ts
```diff
@@ -0,0 +1,120 @@
...
+  onCancel(event: Event): void {
+    event.preventDefault();
+    this.returnToPreviousPage();
+  }
+
+  onSuccess(): void {
+    this.returnToPreviousPage();
+  }
+
+  onError(_error: Error): void {}
```
duncanc19, line 97: Shall we have a server error message?

frontend/src/app/features/workers/update-workplace-details-after-staff-changes/update-total-number-of-staff/update-total-number-of-staff.component.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  public getFirstErrorMessage(item: string): string {
+    const errorType = Object.keys(this.form.get(item).errors)[0];
+    return this.errorSummaryService.getFormErrorMessage(item, errorType, this.formErrorsMap);
+  }
+
+  public returnToPreviousPage() {
+    const returnToUrl = this.establishmentService.returnTo?.url;
+    if (returnToUrl) {
+      this.router.navigate(returnToUrl);
```
duncanc19, line 107: `returnTo` would have to be set from the summary page if we want to use it but I don't think we get much value from it as location.back should do the same thing

PR #6563, opened 2025-04-07
Feat/1638 update leavers

ssrome:
#### Work done
- Create update leavers component
- Add update-leavers-job-roles to routing

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/core/services/update-workplace-after-staff-changes.service.ts
+++ b/frontend/src/app/core/services/update-workplace-after-staff-changes.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { Starter, Vacancy } from '@core/model/establishment.model';
+import { Starter, Vacancy, Leaver } from '@core/model/establishment.model';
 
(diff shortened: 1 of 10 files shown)


frontend/src/app/core/services/update-workplace-after-staff-changes.service.ts
```diff
@@ -48,6 +50,14 @@ export class UpdateWorkplaceAfterStaffChangesService {
   set selectedStarters(updatedStarters: Starter[]) {
     this._selectedStarters = updatedStarters;
   }
+
+  get selectedLeavers(): Vacancy[] {
+    return this._selectedLeavers;
+  }
+
+  set selectedLeavers(updatedVacancies: Vacancy[]) {
```
duncanc19, line 58: Can we update the types of these getters and setters and the name of the parameter?

ssrome: Types have been updated

duncanc19 (approved):

PR #6568, opened 2025-04-10
Fix/update slv validations and alert

kapppa-joe:
#### Work done
- Fix a bug on update SLV validation when user clicked radio button during an error
- Update validation message for starters and leavers ("Select there are no starters or..." --> "Select no staff started or...")
- Remove backlink on "Do you want to delete another worker" page
- Change alert service to remove existing alert on NavigationEnd (instead of NavigationStart)

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/core/services/alert.service.spec.ts
+++ b/frontend/src/app/core/services/alert.service.spec.ts
@@ -5,7 +5,7 @@ import { NavigationEnd, Router } from '@angular/router';
 import { BehaviorSubject } from 'rxjs';
 import { WindowRef } from './window.ref';
 
-fdescribe('BackLinkService', () => {
+describe('BackLinkService', () => {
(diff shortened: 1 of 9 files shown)


ssrome (approved):

frontend/src/app/core/services/alert.service.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { TestBed } from '@angular/core/testing';
+
+import { AlertService } from './alert.service';
+import { NavigationEnd, Router } from '@angular/router';
+import { BehaviorSubject } from 'rxjs';
+import { WindowRef } from './window.ref';
+
+describe('BackLinkService', () => {
```
duncanc19, line 8: This needs updating to be AlertService

PR #6569, opened 2025-04-10
Feat/1662 e2e tests for update workplace after staff changes

ssrome:
#### Work done
- Add tests for both the add and delete staff flows
- Add tests to update number of staff, starters, leavers and vacancies from the workplace summary

#### Tests
- [x] Yes


frontend/cypress/e2e/standAloneEstablishment/EditUser/staffRecordsPage.spec.cy.js
duncanc19, line 1: Looks really good! For the long flow tests it's sometimes a bit difficult to tell what page you're on, I don't think it's needed for all of them but it might be helpful to add a couple of comments above some sections to make it clearer where the user is

frontend/cypress/e2e/standAloneEstablishment/EditUser/staffRecordsPage.spec.cy.js
```diff
@@ -129,6 +135,114 @@ describe('Standalone staff records page as edit user', () => {
...
+      cy.getByLabel('Yes').click();
+      cy.contains('button', 'Continue').click();
+      cy.addJobRoles(jobRolesToAdd, 'type');
+
+      onHomePage.clickTab('Staff records');
+
+      cy.get('a[role="button"]').contains('Add a staff record').click();
+
+      addAndConfirmMandatoryStaffDetails(worker1, contractType, mainJobRole);
+      cy.getByLabel('Yes').click();
```
duncanc19, line 164: From the other comment here's an example maybe, I had to think for a bit about what page this would be (Add another staff record?)

duncanc19 (approved):

PR #6579, opened 2025-04-17
Feat/1674 ripples workplace flow vacancies page 3

kapppa-joe:
#### Work done
- Extract job role numbers table to a shared component (JobRoleNumbersTableComponent)
- Amend HowManyVacanciesComponent according to prototype, using the job role numbers table
- Amend UpdateVacancies component to use the job role numbers table
- Rename a service `UpdateWorkplaceAfterStaffChangesService` --> `VacanciesAndTurnoverService`
- Change workplace flow vacancies page 2 to store selected vacancies in `VacanciesAndTurnoverService` instead of localStorage

(Note: the tests in HowManyStartersComponent and HowManyLeaversComponent are marked as skipped for the moment, due to changes in shared directive)

#### Tests
- [x] Yes


duncanc19 (approved):
Looks really good! Just a thought when we have feature branches, it might be good to do things like service renames in a separate PR just so it's easier to focus on the code changes when you're reviewing

PR #6600, opened 2025-05-01
feat/1684-radio-buttons-highest-level-other-quals

ssrome:
#### Work done
- Change other quals levels answers from select to radios
- Update text on call to action button

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/workers/other-qualifications-level/other-qualifications-level.component.html
+++ b/frontend/src/app/features/workers/other-qualifications-level/other-qualifications-level.component.html
@@ -51,5 +51,5 @@
     </div>
   </div>
 
-  <app-submit-button [return]="!insideFlow" (clicked)="setSubmitAction($event)" callToAction="Save"></app-submit-button>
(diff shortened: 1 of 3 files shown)


frontend/src/app/features/workers/other-qualifications-level/other-qualifications-level.component.html
```diff
@@ -47,5 +51,5 @@ <h1 class="govuk-label-wrapper">
     </div>
   </div>
 
-  <app-submit-button [return]="!insideFlow" (clicked)="setSubmitAction($event)" callToAction="Save"></app-submit-button>
+  <app-submit-button [return]="!insideFlow" (clicked)="setSubmitAction($event)" callToAction="Save and continue"></app-submit-button>
```
duncanc19, line 54: I think Save and continue is the default CTA in the flow so I think we might not need to pass it in here

duncanc19 (approved):

frontend/src/app/features/workers/other-qualifications-level/other-qualifications-level.component.html
ssrome: Remove this

PR #6606, opened 2025-05-08
Add e2e tests for training and quals records

ssrome:
#### Work done
- Add e2e tests to add, update and delete training records
- Add e2e tests for qualifications
- Move and rename existing training and quals e2e tests

#### Tests
- [x] Yes


frontend/cypress/support/commands/trainingAndQualsCommands.js
```diff
@@ -0,0 +1,95 @@
...
+
+  const queryString1 = `SELECT "ID" FROM cqc."Worker"
+  WHERE "EstablishmentFK" = $1
+  AND "NameOrIdValue" = $2`;
+
+  const parameters1 = [establishmentID, workerName];
+
+  cy.task('dbQuery', { queryString: queryString1, parameters: parameters1 }).then((result) => {
+    cy.wrap(result.rows[0]?.ID).as('workerId');
+  });
```
duncanc19, line 61: A lot of the logic in each of these commands is retrieving the workerId and looks to be the same, could we make a helper function to do this? It could maybe go into sharedCommands as I imagine this might be useful in other places too

ssrome: Moved the logic to get the workerId into the worker commands file

PR #6607, opened 2025-05-08
Feat/1699 cwp database changes

Zuhal-Ayob:
#### Work done
- Add a table for the care workforce pathway 
- Insert data for the care workforce pathway  table
- Add Id of the care workforce pathway  table as an foreign key in the worker table.
- Create a property file for the care workforce pathway .
- Add the properties in the worker property file.

#### Tests
- [x] Yes


backend/migrations/20250507150411-createTableForCareWorkforcePathwayRoleCategories.js
```diff
@@ -0,0 +1,133 @@
...
+    Seq: 70,
+    Title: 'Registered manager',
+    Description: 'Is focussed on regulatory and legal requirements, and runs the service',
+    AnalysisFileCode: 7,
+    BulkUploadCode: 7,
+  },
+  { ID: 8, Seq: 80, Title: 'I do not know', Description: null, AnalysisFileCode: -2, BulkUploadCode: 999 },
+  {
+    ID: 9,
+    Seq: 90,
```
duncanc19, line 64: Pretty low risk but is it worth having the sequence numbers higher for the _don't know_ and _none of the above_ options just in case more than 9 options are added to the end of the list?

backend/server/models/classes/worker/properties/careWorkforcePathwayRoleCategoryProperty.js
```diff
@@ -0,0 +1,89 @@
...
+
+  static clone() {
+    return new CareWorkforcePathwayRoleCategoryProperty();
+  }
+
+  async restoreFromJson(document) {
+    if (document.careWorkforcePathwayRoleCategory) {
+      const validatedData = await this._validateOrigin(document.careWorkforcePathwayRoleCategory);
+
+      if (validatedData) {
```
duncanc19, line 21: The default is already null for the `validateOrigin` function except for the last part, so could we return null at the bottom and then you don't need to handle it here?

backend/server/models/classes/worker/properties/careWorkforcePathwayRoleCategoryProperty.js
```diff
@@ -0,0 +1,89 @@
...
+
+    return {
+      careWorkforcePathwayRoleCategory: {
+        currentValue: this.property,
+        ...this.changePropsToJSON(showPropertyHistoryOnly),
+      },
+    };
+  }
+
+  async _validateOrigin(payloadData) {
```
duncanc19, line 67: I'm wondering if there's a better name for this too as with `validateOrigin` it's not too clear that it's returning something. Maybe something along the lines of `getRoleCategoryFromDatabase`

ssrome (approved):

PR #6615, opened 2025-05-15
Feat/1704 feature flag cwp questions

Zuhal-Ayob:
#### Work done
- Add feature flag for CWP component 
- Applied feature flag in qualification and other qualifications component

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/workers/care-workforce-pathway/care-workforce-pathway.component.spec.ts
+++ b/frontend/src/app/features/workers/care-workforce-pathway/care-workforce-pathway.component.spec.ts
@@ -84,6 +84,14 @@ describe('CareWorkforcePathwayRoleComponent', () => {
     expect(component).toBeTruthy();
   });
 
+  it('component should not render when deature flag is on', async () => {
+    const overrides = { cwpQuestionsFlag: true };
+    const { fixture, queryByTestId } = await setup(overrides);
+
+    fixture.detectChanges();
+    expect(queryByTestId('cwp-flag')).toBeFalsy();
+  });
+
   it('should show the heading and caption', async () => {
     const overrides = { cwpQuestionsFlag: false };
     const { getByText, getByTestId, component } = await setup(overrides);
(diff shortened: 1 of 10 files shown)


frontend/src/app/features/workers/care-workforce-pathway/care-workforce-pathway.component.spec.ts
```diff
@@ -52,9 +54,13 @@ describe('CareWorkforcePathwayRoleComponent', () => {
           provide: CareWorkforcePathwayService,
           useClass: MockCareWorkforcePathwayService,
         },
+        { provide: FeatureFlagsService, useClass: MockFeatureFlagsService },
         AlertService,
         WindowRef,
       ],
+      componentProperties: {
+        cwpQuestionsFlag: overrides.cwpQuestionsFlag,
+      },
```
kapppa-joe, line 63: as `cwpQuestionsFlag` is not an `@Input` value for this component, I think we've better avoid to set it as componentProperties, because that doesn't emulate how the component read the value

frontend/src/app/features/workers/care-workforce-pathway/care-workforce-pathway.component.spec.ts
```diff
@@ -79,8 +85,9 @@ describe('CareWorkforcePathwayRoleComponent', () => {
   });
 
   it('should show the heading and caption', async () => {
-    const { getByText, getByTestId } = await setup();
-
+    const overrides = { cwpQuestionsFlag: false };
+    const { getByText, getByTestId, component } = await setup(overrides);
+    component.cwpQuestionsFlag = false;
```
kapppa-joe, line 90: same as above, I think we've better avoid to set the value of `cwpQuestionsFlag` manually, as it doesn't emulate how the component actually read the value
if test can't pass without manually doing this, then it probably indicate something is not quite right in our setup

frontend/src/app/features/workers/other-qualifications-level/other-qualifications-level.component.spec.ts
```diff
@@ -106,15 +122,17 @@ describe('OtherQualificationsLevelComponent', () => {
...
     it('should not render the progress bars when accessed from outside the flow', async () => {
-      const { queryByTestId } = await setup();
+      const overrides = { cwpQuestionsFlag: false, returnUrl: true };
+      const { queryByTestId } = await setup(overrides);
 
       expect(queryByTestId('progress-bar-1')).toBeFalsy();
     });
   });
 
   describe('navigation', () => {
```
kapppa-joe, line 132: should we add some tests for the case when `cwpQuestionsFlag` is true? As we've added the behaviour of hiding the care-workforce-pathway question, we better have test to cover that

kapppa-joe (approved):

PR #6621, opened 2025-05-22
Feat/1701 cwp bu staff file

Zuhal-Ayob:
#### Work done
- Add download functionality for the Care Workforce Pathway.
- Add upload functionality for the Care Workforce Pathway.

#### Tests
- [x] Yes


lambdas/bulkUpload/classes/workerCSVValidator.js
```diff
@@ -2975,6 +3009,12 @@ class WorkerCsvValidator {
       yearOfEntry: this._yearOfEntry ? this._yearOfEntry : undefined,
       healthAndCareVisa: this._healthAndCareVisa ? this._healthAndCareVisa : undefined,
       employedFromOutsideUk: this._employedFromOutsideUk ? this._employedFromOutsideUk : undefined,
+      careWorkforcePathwayRoleCategory: this._careWorkForcePathwayCategory
+        ? {
+            roleCategoryId: this._careWorkForcePathwayCategory,
```
kapppa-joe, line 3014: this probably will not work when the bulk upload code is 998 (None of the above) or 999 (Not known), as the roleCategoryId is supposed to be the CareWorkForceRoleCategoryFK

lambdas/bulkUpload/classes/workerCSVValidator.js
```diff
@@ -2901,6 +2901,32 @@ class WorkerCsvValidator {
...
+            this._generateWarning(
+              'The code you have entered for CWPCATEGORY is incorrect and will be ignored',
+              'CWPCATEGORY',
+            ),
+          );
+        } else {
+          this._careWorkForcePathwayCategory = myValidatedCwpCategory;
+        }
+      }
+    }
```
kapppa-joe, line 2927: I can see two issues in this part:
1. If we are already using `this.BUDI.careWorkforcePathwayRoleCategory()` to convert bulk upload code to ASC code, we shouldn't manually add the special cases of 999 and 998. careWorkforcePathwayRoleCategory() should handle the conversion if it's working correctly

2. Tests are failing on the pipeline with error `TypeError: this.BUDI.careWorkforcePathwayRoleCategory is not a function`. Probably we need to implement careWorkforcePathwayRoleCategory in `lambdas/bulkUpload/classes/BUDI.js` and mock or update the mapping

PR #6622, opened 2025-05-28
Feat/1702 care workforce pathway new page

kapppa-joe:
#### Work done
- Add a `CareWorkforcePathwayWorkersSummaryComponent` for the new page
- Add a `RequireCWPAnswerForSomeWorkersGuard` to redirect user to dashboard home when CWP question is answered for all workers
- Amend `CareWorkforcePathwayRoleComponent` question page to show different text when visited from the CWP worker summary page. Also change the `navigate` method to async ensure alert is added at the correct timing
- Add an e2e test for the journey of answering CWP questions from new page
- Fix a bug in `SummarySectionComponent' component that causes problem when clicking link from flag
- Amend migration file and add line breaks `\n` to the description text to meet design requirements

#### Tests
- [x] Yes


frontend/src/app/core/guards/require-cwp-answer-for-some-workers/require-cwp-answer-for-some-workers.guard.ts
```diff
@@ -0,0 +1,31 @@
...
+import { inject } from '@angular/core';
+import { EstablishmentService } from '@core/services/establishment.service';
+import { ParentSubsidiaryViewService } from '../../../shared/services/parent-subsidiary-view.service';
+
+export const RequireCWPAnswerForSomeWorkersGuard: CanActivateFn = async (route, _state) => {
+  const careWorkforcePathwayService = inject(CareWorkforcePathwayService);
+  const establishmentService = inject(EstablishmentService);
+  const router = inject(Router);
+  const parentSubsidiaryViewService = inject(ParentSubsidiaryViewService);
+
```
duncanc19, line 12: Is there a reason why you've gone for a function rather than a class? I think a class might be a bit cleaner as it's got a few dependencies and most of the guards we've got in the codebase are classes

backend/server/routes/establishments/careWorkforcePathway.js
```diff
@@ -6,22 +6,55 @@ const { hasPermission } = require('../../utils/security/hasPermission');
 const getNoOfWorkersWhoRequireCareWorkforcePathwayRoleAnswer = async (req, res) => {
   const establishmentId = req.establishmentId;
   try {
-    let workers = await models.worker.getAllWorkersWithoutCareWorkforceCategory(establishmentId);
+    let workerCount = await models.worker.countAllWorkersWithoutCareWorkforceCategory(establishmentId);
```
duncanc19, line 9: Looks like this can be a const instead of let

frontend/src/app/features/workers/care-workforce-pathway/care-workforce-pathway.component.ts
```diff
@@ -94,13 +111,49 @@ export class CareWorkforcePathwayRoleComponent extends QuestionComponent {
...
+      case 'summary':
+        return this.router.navigate(this.getRoutePath(''));
+
+      case 'exit':
+        return this.router.navigate(['/dashboard'], { fragment: 'staff-records' });
+
+      case 'return':
+        return this.router.navigate(this.returnUrl);
+    }
+  }
```
duncanc19, line 152: I think it makes sense not updating the line in onSuccess in the question component in case there are side effects, but would it be safe just updating the navigate method to return in the question component? Feels like it should be fine

duncanc19 (approved):

PR #6624, opened 2025-06-02
Feat/1710 cwp workplace new question 1 backend changes

ssrome:
#### Work done
- Create table for care workforce pathway awareness questions
- Add property for care workforce pathway awareness

#### Tests
- [x] Yes


backend/server/routes/careWorkforcePathwayWorkplaceAwarenessAnswers.js
```diff
@@ -0,0 +1,37 @@
...
+    let careWorkforcePathwayWorkplaceAwarenessAnswers = [];
+
+    giveCareWorkforcePathwayWorkplaceAwarenessAnswers.forEach((thisCareWorkforcePathwayWorkplaceAwarenessAnswer) => {
+      careWorkforcePathwayWorkplaceAwarenessAnswers.push({
+        id: thisCareWorkforcePathwayWorkplaceAwarenessAnswer.id,
+        title: thisCareWorkforcePathwayWorkplaceAwarenessAnswer.title,
+      });
+    });
+    return careWorkforcePathwayWorkplaceAwarenessAnswers;
+  }
```
duncanc19, line 31: Could we just get the id and title attributes in the database call and then it wouldn't need any formatting? If not maybe using map would be a bit cleaner than creating an empty array and pushing to it

ssrome: Now updated to only get the id and title attributes. Removed the formatting function

PR #6625, opened 2025-06-03
Feat/1720 cwp workplace question 2 backend

kapppa-joe:
#### Work done
- add migration for a new table for CWP use reasons, a junction table to record n:m relationships and adding a column of CWP use to Establishment
- add a careWorkforcePathwayUse property file to handle conversions between establishment class and database record
- add a `POST /api/establishment/{uid}/careWorkforcePathway/careWorkforcePathwayUse` endpoint for updating the CWP use and reasons 
- add a `GET /api/careWorkforcePathway/useReasons` endpoint for retrieving a list of all available reasons

#### Tests
- [x] Yes

### DIFF
--- a/backend/migrations/20250528140219-createTableAndColumnsForCareWorkforcePathwayUse.js
+++ b/backend/migrations/20250528140219-createTableAndColumnsForCareWorkforcePathwayUse.js
@@ -1,0 +1,142 @@
+'use strict';
+
+const models = require('../server/models/index');
+
+const cwpReasonsTable = { tableName: 'CareWorkforcePathwayReasons', schema: 'cqc' };
+const junctionTable = { tableName: 'EstablishmentCWPReason', schema: 'cqc' };
+const establishmentTable = { tableName: 'Establishment', schema: 'cqc' };
+
+/** @type {import('sequelize-cli').Migration} */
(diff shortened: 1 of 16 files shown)


backend/migrations/20250528140219-createTableAndColumnsForCareWorkforcePathwayUse.js
```diff
@@ -0,0 +1,143 @@
...
+            allowNull: true,
+          },
+        },
+        { transaction },
+      );
+
+      const addColumnToEstablishmentTable = queryInterface.addColumn(
+        establishmentTable,
+        'CareWorkforcePathwayUse',
+        {
```
Zuhal-Ayob, line 75: Do we need the audit columns here?

Zuhal-Ayob (approved):

backend/migrations/20250528140219-createTableAndColumnsForCareWorkforcePathwayUse.js
kapppa-joe: Thanks!! I have added the audit columns just now. 
Also tested in local that the audit columns are updated when we update the value

PR #6628, opened 2025-06-09
Feat/1712 cwp new question in workplace flow

ssrome:
#### Work done
- Add page for care workforce pathway awareness in workplace flow
- Add endpoint to update care workforce pathway awareness on establishment
- Fix incorrect datatypes on establishment table and model

#### Tests
- [x] Yes


backend/server/routes/establishments/careWorkforcePathway/careWorkforcePathwayAwareness.js
duncanc19, line 1: It looks like we're just loading what's passed in the body of the request and there isn't logic specific to `careWorkforcePathwayWorkplaceAwareness`, so I think we might be able to just use the standard `updateSingleEstablishmentField` and not have to create a new endpoint for it

duncanc19 (approved):

PR #6662, opened 2025-07-03
1735 investigate establishment routes that update properties

ssrome:
#### Work done
- Create new endpoints to get and update the requested establishment field
- Updated frontend components to use the new endpoints
- Delete the staff, employerType, shareData and name endpoints

#### Tests
- [x] Yes


backend/server/routes/establishments/establishmentField.js
```diff
@@ -0,0 +1,96 @@
...
+      throw new HttpError('Request is invalid', 400);
+    }
+
+    await thisEstablishment.save(req.username);
+    return res.status(200).json(thisEstablishment.toJSON(false, false, false, true, false, filteredProperties));
+  } catch (error) {
+    console.error('Establishment::%s POST: ', property, error.message);
+    if (error instanceof HttpError) {
+      const safeMessage = error.safe || 'An error occurred.';
+      return res.status(error.statusCode).send(safeMessage);
```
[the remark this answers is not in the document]
ssrome: Suggested package hasn't been updated in 7 years

backend/server/routes/establishments/establishmentField.js
```diff
@@ -0,0 +1,107 @@
...
+      null,
+      error,
+      null,
+      `Failed to retrieve Establishment with id/uid: ${establishmentId}`,
+    );
+
+    if (error instanceof HttpError) {
+      const safeMessage = error.safe || 'An error occurred.';
+      return res.status(error.statusCode).send(safeMessage);
+    }
```
kapppa-joe, line 55: it seems that thisError (type EstablishmentRestoreException)  has .safe but error (type HttpError) does not, so `error.safe` here probably will always be undefined.
Do you want to use `error.message` or `thisError.safe` instead?

backend/server/routes/establishments/establishmentField.js
```diff
@@ -0,0 +1,107 @@
...
+    if (!isValidEstablishment) {
+      throw new HttpError('Request is invalid', 400);
+    }
+
+    await thisEstablishment.save(req.username);
+    return res.status(200).json(thisEstablishment.toJSON(false, false, false, true, false, filteredProperties));
+  } catch (error) {
+    console.error('Establishment::%s POST: ', property, error.message);
+    if (error instanceof HttpError) {
+      const safeMessage = error.safe || 'An error occurred.';
```
kapppa-joe, line 89: similar to the above one, this `error.safe` probably would always be undefined. Do we want to use `error.message` instead?

backend/server/test/unit/routes/establishments/establishmentField.spec.js
```diff
@@ -0,0 +1,176 @@
...
+      expect(res.statusCode).to.equal(500);
+    });
+  });
+
+  describe('/updateEstablishmentFieldWithAudit', () => {
+    let mockRequest;
+    const setupTests = (propertyToRequest = property) => {
+      mockRequest = {
+        method: 'POST',
+        url: `/api/establishment/${establishmentId}/updateEstablishmentFieldWithAudit`,
```
kapppa-joe, line 88: could we check again for the url? because when I test this branch locally, if I answer the share data question, it seems to be calling `POST http://localhost:8080/api/establishment/{uid}/establishmentField/ShareData` for example

backend/server/routes/establishments/establishmentField.js
ssrome: Also updated to use `error.message`

backend/server/test/unit/routes/establishments/establishmentField.spec.js
ssrome: Url has been corrected

backend/server/routes/establishments/establishmentField.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+};
+
+const updateEstablishmentFieldWithAudit = async (req, res) => {
+  const establishmentId = req.establishmentId;
+  const thisEstablishment = new Establishment.Establishment(req.username);
+
+  const property = req.params?.property;
+
+  filteredProperties.push(property);
```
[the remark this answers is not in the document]
ssrome: Updated as suggested

kapppa-joe (approved):

PR #6664, opened 2025-07-11
Feat/1750 parent sort option

ssrome:
#### Work done
- Add sort option to Your other workplaces page
- Update backend to return child workplaces in the order of requested sort
- Add function in WorkplaceService to hold sort value when coming from home page

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/establishment.js
+++ b/backend/server/models/establishment.js
@@ -2281,6 +2281,7 @@ module.exports = function (sequelize, DataTypes) {
     searchTerm = '',
     getPendingWorkplaces = false,
     getAttentionFlags = false,
+    sortBy = 'workplaceNameAsc',
   ) {
     const offset = pageIndex * limit;
     let ustatus;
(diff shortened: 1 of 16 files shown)


frontend/src/app/features/workplace/view-my-workplaces/view-my-workplaces.component.spec.ts
```diff
@@ -278,4 +281,86 @@ describe('ViewMyWorkplacesComponent', () => {
...
   });
+
+  describe('sort by', () => {
+    it('should not show if there are no workplaces', async () => {
+      const { queryByTestId } = await setup({ hasChildWorkplaces: false });
+
+      expect(queryByTestId('sortBy')).toBeFalsy();
+    });
+
+    it('should not show if there is 1 workplace', async () => {
```
duncanc19, line 292: Could we pass childWorkplaces into the overrides so then we don't have to set totalWorkplaceCount directly for these tests?

frontend/src/app/core/services/workplace.service.ts
```diff
@@ -66,4 +72,12 @@ export class WorkplaceService extends WorkplaceInterfaceService {
...
     };
   }
+
+  public setAllWorkplacesSortValue(value: string) {
+    this._allWorkplacesSortValue = value;
+  }
+
+  public getAllWorkplacesSortValue(): string {
+    return this._allWorkplacesSortValue;
+  }
```
duncanc19, line 82: Is the intention that the sort value would always stay until the user next changes it? With this current set up it looks like if they went into a child workplace/did something else unrelated and then went back to the page the sort value would still be set

backend/server/models/establishment.js
```diff
@@ -2298,6 +2299,30 @@ module.exports = function (sequelize, DataTypes) {
...
+      ],
+      workplaceToCheckDesc: [
+        [sequelize.literal("\"Status\" IN ('PENDING', 'IN PROGRESS')"), 'ASC'],
+        ['showFlag', 'DESC'],
+        ['NameValue', 'DESC'],
+      ],
+    }[sortBy] || [
+      [sequelize.literal("\"Status\" IN ('PENDING', 'IN PROGRESS')"), 'ASC'],
+      ['NameValue', 'ASC'],
+    ];
```
duncanc19, line 2324: I think we could simplify this a bit because all of them start with the pending/in progress sort. You could have something like this if you made a separate sortOptions object
```
const order = [[sequelize.literal("\"Status\" IN ('PENDING', 'IN PROGRESS')"), 'ASC'], ...(sortOptions[sortBy] ?? [['NameValue', 'ASC']])]
``` 
or maybe it could be in a helper function which returns the default

frontend/src/app/features/workplace/view-my-workplaces/view-my-workplaces.component.spec.ts
ssrome: Now updated as suggested

PR #6677, opened 2025-07-31
Feat/1756 DHA staff record flow question

kapppa-joe:
#### Work done
- Add migrations for the new columns related to worker delegated healthcare activities question, and a new property to worker class
- Add a new worker question `CarryOutDelegatedHealthcareActivitiesComponent` to frontend. Also update the logic to determine next route in previous question `AdultSocialCareStartedComponent`
- Add a sequelize hook to worker model, which clears the answer of worker DHA question when a worker change to a job role that can't do DHA

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/workers/adult-social-care-started/adult-social-care-started.component.spec.ts
+++ b/frontend/src/app/features/workers/adult-social-care-started/adult-social-care-started.component.spec.ts
@@ -3,6 +3,7 @@ import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { getTestBed } from '@angular/core/testing';
 import { UntypedFormBuilder, ReactiveFormsModule } from '@angular/forms';
 import { ActivatedRoute, Router, RouterModule } from '@angular/router';
+import { Contracts } from '@core/model/contracts.enum';
 import { Worker } from '@core/model/worker.model';
 import { WorkerService } from '@core/services/worker.service';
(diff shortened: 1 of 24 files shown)


frontend/src/app/features/workers/adult-social-care-started/adult-social-care-started.component.spec.ts
```diff
@@ -109,166 +102,310 @@ describe('AdultSocialCareStartedComponent', () => {
...
-      'days-of-sickness',
-    ]);
-  });
-
-  it(`should call submit data and navigate with the 'contract-with-zero-hours' url when 'Save and continue' is clicked and the contract typs is not permanent or temporary`, async () => {
-    const { component, getByText, routerSpy } = await setup(true, 'other');
-
-    const button = getByText('Save and continue');
-    fireEvent.click(button);
+  describe('when inside add new worker workflow', () => {
```
kapppa-joe, line 112: Note: I have moved the existing tests for next route to a function `shouldSkipWorkerDHAQuestion`, so that we can reuse the same tests, because now the conditions to determine next route is more complicated

If should see worker DHA question --> navigate to DHA question
if should not see DHA question --> navigate to `days-of-sickness` or `contract-with-zero-hours`, depends on worker contract type

Zuhal-Ayob (approved):

PR #6678, opened 2025-07-31
Feat/1759 sort on staff records

ssrome:
#### Work done
- Add additional sort values to the staff summary
- Add additional sorting options to the workers endpoint
- Add sortBy Service to store values to maintain staff summary search, pagination and sort

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/core/services/sortBy.service.ts
+++ b/frontend/src/app/core/services/sortBy.service.ts
@@ -1,0 +1,47 @@
+import { Injectable } from '@angular/core';
+import {  NavigationStart, Router } from '@angular/router';
+import {  filter, } from 'rxjs/operators';
+import { TabsService } from './tabs.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SortByService {
+  public navUrl: string;
+  private _totalWorkerCount: number
+
+  constructor(private router: Router, private tabsService: TabsService) {
+    this.navUrl;
+
+    this.router.events.pipe(filter((event) => event instanceof NavigationStart)).subscribe((event: NavigationStart) => {
+      console.log(event)
+
+      const destinationUrl = event.url;
+
+      if (!destinationUrl.includes('staff-record')) {
+        this.clearLocalStorage();
+      }
+    });
+
+    this.tabsService.selectedTab$.subscribe((newTab) => {
+      if (newTab && newTab !== 'staff-records') {
+        this.clearLocalStorage();
+      }
+    });
+  }
+
+  public clearLocalStorage() {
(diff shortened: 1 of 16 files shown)


frontend/src/app/core/services/sortBy.service.ts
duncanc19, line 1: Not a big deal but it's not the usual casing for this and the spec file, we mostly seem to go with the convention of `sort-by.service.ts`/`sort-by.service.spec.ts` and then have the mock service in pascal case like you've got it so it might be good to stick with that

frontend/src/app/core/test-utils/MockSortByService.ts
```diff
@@ -0,0 +1,31 @@
...
+          staffSummarySearchTerm: 'Ma',
+          staffSummaryIndex: '0',
+        }
+      : {
+          staffSummarySortValue: null,
+          staffSummarySearchTerm: null,
+          staffSummaryIndex: null,
+        };
+  }
+}
```
duncanc19, line 31: I think it's a bit more flexible if we use a generic overrides rather than using a boolean to have two options. That way if we want to test another case we can just pass in the values we want. You could change `useLocalStorageValuesForSort` to be an object with it set to the null version, or you could just pass in returnLocalStorageForSort as a function, either would work with generic overrides

frontend/src/app/core/services/sortBy.service.ts
```diff
@@ -0,0 +1,45 @@
+import { Injectable } from '@angular/core';
+import { NavigationStart, Router } from '@angular/router';
+import { filter } from 'rxjs/operators';
+import { TabsService } from './tabs.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SortByService {
+  public navUrl: string;
```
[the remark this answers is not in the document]
ssrome: Removed unused variable

frontend/src/app/core/services/sortBy.service.ts
ssrome: Renamed file

duncanc19 (approved):

frontend/src/app/core/test-utils/MockSortByService.ts
```diff
@@ -0,0 +1,31 @@
+import { Injectable } from '@angular/core';
+import { Router } from '@angular/router';
+import { SortByService } from '@core/services/sort-by.service';
+import { TabsService } from '@core/services/tabs.service';
+
+@Injectable()
+export class MockSortByService extends SortByService {
+  public useLocalStorageValuesForSort = false;
```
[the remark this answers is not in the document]
ssrome: Has been removed

PR #6684, opened 2025-08-05
fix/1770-wording-for-read-only-users-in-home-page-summary

ssrome:
#### Work done
- Add extra message for read only users in summary section

#### Tests
- [x] Yes


duncanc19 (commented):
I'm not sure if this is working as expected. I've gone onto the branch and I can still see the '_Staff total does not match staff records added_' message on my read user. I think by having this only when there is no message, the checks are still happening and in this case it sets the message when it finds a mismatch between number of staff and number of records (which will be undefined). I think we'll need to check if they have permission before doing the message checks.

PR #6693, opened 2025-08-19
Feat/1791 dha question 2 on workplace summary

ssrome:
#### Work done
- Add DHA question 2 to workplace summary. It will only show if "Yes" is answered for question 2
- Add pipe to format answer

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/shared/pipes/format-what-dha.pipe.ts
+++ b/frontend/src/app/shared/pipes/format-what-dha.pipe.ts
@@ -12,12 +12,6 @@ export class FormatWhatDhaPipe implements PipeTransform {
 
     const { knowWhatActivities } = whatDHA;
 
-    switch (knowWhatActivities) {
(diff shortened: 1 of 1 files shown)


frontend/src/app/shared/pipes/format-what-dha.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    switch (knowWhatActivities) {
+      case "Don't know":
+        return 'Not known';
+
+      default:
+        return '-';
+    }
+  }
+}
```
duncanc19, line 23: I think this can be simplified a bit because there's only the 'Don't know' case or the default.
You could maybe have something along the lines of
```
if (whatDHA?.knowWhatActivities === "Don't know") return ...
return '-';
```

duncanc19 (approved):

PR #6698, opened 2025-08-21
Feat/1792 dha clear answers for workplace q1 q2 and workers on main service change

ssrome:
#### Work done
- Added `clearDoDHAWorkplaceOnMainServiceChange` hook to clear DHA workplace q1, q2 and worker dha when the establishment main service is changed to one where `CanDoDelegatedHealthcareActivities` is not true 
- Also handled the case when worker dha has a value but workplace q1 is null

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/models/hooks/establishmentHooks.js
+++ b/backend/server/models/hooks/establishmentHooks.js
@@ -55,41 +55,45 @@ const clearDoDHAWorkplaceOnMainServiceChange = async (establishment, options) =>
       return;
     }
 
-    if (establishment.changed('MainServiceFKValue')) {
-      const staffDoDelegatedHealthcareActivitiesValue = establishment.staffDoDelegatedHealthcareActivities;
-
-      if (!staffDoDelegatedHealthcareActivitiesValue) {
-        return;
-      }
-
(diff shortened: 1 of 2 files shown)


backend/server/models/hooks/establishmentHooks.js
```diff
@@ -47,4 +47,56 @@ const clearDHAWorkplaceAnswerOnChange = async (establishment, options) => {
...
-module.exports = { clearDHAWorkerAnswersOnWorkplaceChange, clearDHAWorkplaceAnswerOnChange };
+const clearDoDHAWorkplaceOnMainServiceChange = async (establishment, options) => {
+  try {
+    const models = establishment.sequelize.models;
+
+    if (!establishment || !models) {
+      return;
+    }
+
+    if (establishment.changed('MainServiceFKValue')) {
```
duncanc19, line 58: It might be nice if we flip this if so it returns if `!establishment.changed('MainServiceFKValue')`, just so the if blocks are less nested

ssrome: Refactored to reduce nesting

duncanc19 (approved):

PR #6710, opened 2025-09-03
Feat/1796 pdf download on deletion

Zuhal-Ayob:
#### Work done
-

#### Tests
- [ ] Yes


frontend/src/app/core/services/pdf-make.service.ts
```diff
@@ -0,0 +1,305 @@
...
+              },
+              layout: {
+                fillColor: '#1d70b8',
+                paddingTop: () => 5,
+                paddingBottom: () => 5,
+                hLineWidth: () => 0,
+                vLineWidth: () => 0,
+              },
+              width: '70%',
+              margin: [5, 20, 0, 20] as [number, number, number, number],
```
kapppa-joe, line 100:
```suggestion
              margin: [5, 20, 0, 20] as Margins,
```
We can replace these [number, number, number, number] with the `Margins` type from the pdfmake library 

```
import { Content, Margins, TDocumentDefinitions } from 'pdfmake/interfaces';
```

PR #6847, opened 2026-03-19
Feat/pay new columns

Zuhal-Ayob:
#### Work done
- Add a new migration to add new columns to the worker table.

#### Tests
- [ ] Yes


backend/migrations/20260318164305-updatePayForMultiStaffPageViewed.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+const workerTable = { tableName: 'Establishment', schema: 'cqc' };
```
ssrome: The variable needs to be renamed and then changed below

ssrome (approved):

PR #6856, opened 2026-03-30
feat/1907 Fast track pay updates by job role

chris-groves:
#### Work done
- New page enables user to update worker pay by job role
- Page route is `../workplace/<workplace-uid>/staff-record/fast-track-pay-updates`
- Create `FastTrackPayUpdatesComponent`
- Update `WorkersGroupedByJobRole` to include the fields `count` and `annualHourlyPay` (for storing updated pay details)
- Create `WorkersByJobRoleResolver` to get the `WorkersGroupedByJobRole` data to display on the page
- Create `workersGroupedByJobRole` field on `WorkerService` to store the updated pay details to be retrieved later in the flow
- The `Continue` button submits the updated pay details to the `WorkerService` but does not navigate as the next page in the flow is not yet developed 
- The `Cancel` button navigates to `/` - this will need to be updated when the previous page in the flow is developed
- e2e tests will need to be added when the page before is developed

<details>
<summary>Screenshot</summary>
image

</details>

#### Tests
- [x] Yes

### DIFF
--- a/frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.html
+++ b/frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.html
@@ -1,0 +1,84 @@
+<form (ngSubmit)="onSubmit()" [formGroup]="form" #formEl>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span data-testid="caption" class="govuk-caption-l">Staff records</span>
+      <h1 data-testid="heading" class="govuk-heading-l govuk-!-margin-bottom-7">Fast-track pay updates by job roles</h1>
+    </div>
+  </div>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <table class="govuk-table govuk-!-margin-bottom-3">
+        <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th data-testid="job-role-heading" scope="col" class="govuk-table__header" style="width: 445px">Job role</th>
+          <th data-testid="pay-heading" scope="col" class="govuk-table__header" style="margin-left: 0;">New hourly pay rate or salary</th>
+        </tr>
+        </thead>
+        <tbody class="govuk-table__body" formArrayName="workers">
+          @for (workers of workersByJobRole.groups; track $index) {
+            <tr class="govuk-table__row" [formGroupName]="$index">
+              <td class="govuk-table__cell" style="vertical-align: top; padding-top: 18px; padding-bottom: 18px;">
+                <div style="line-height: 1.25; display: flex; justify-content: space-between; align-items: flex-start;">
+                  <span>{{ workers.title }} ({{ workers.count }} record{{ workers.count === 1 ? '' : 's' }})</span>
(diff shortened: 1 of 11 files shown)


frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.html
```diff
@@ -0,0 +1,84 @@
...
+                      </label>
+                    </div>
+                    <div class="govuk-radios__item">
+                      <input
+                        class="govuk-radios__input"
+                        [attr.data-testid]="'salary-radio-' + $index"
+                        [formControlName]="'value'"
+                        [id]="'salary-' + $index"
+                        type="radio"
+                        value="Salary">
```
kapppa-joe: for this one, I think the value need to be "Annually" as that is how we are storing it in database

frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.html
```diff
@@ -0,0 +1,84 @@
...
+                      <label class="govuk-label govuk-radios__label" for="salary">
```
kapppa-joe: here I think for= should be `"'salary-' + $index"` in order for browser to match it with the input el

frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.html
```diff
@@ -0,0 +1,84 @@
...
+                  <div class="govuk-radios govuk-radios--inline govuk-radios--small govuk-!-margin-bottom-0">
+                    <div class="govuk-radios__item" style="margin-right: 5px;">
+                      <input
+                        class="govuk-radios__input"
+                        [attr.data-testid]="'hourly-radio-' + $index"
+                        [formControlName]="'value'"
+                        [id]="'hourly-' + $index"
+                        type="radio"
+                        value="Hourly">
+                      <label class="govuk-label govuk-radios__label" for="hourly">
```
kapppa-joe: here I think for= should be `"'hourly-' + $index"` in order for browser to match it with the input el

frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    this.workersByJobRole.groups.forEach((group, index) => {
+      const formEntry = workersFormValues[index];
+
+      group.annualHourlyPay = {
+        value: formEntry.value,
+        rate: formEntry.rate,
+      };
+    });
+
+    this.workerService.setWorkersGroupedByJobRole(this.workersByJobRole);
```
kapppa-joe: For this part, maybe we can have another const to store the updated workers, rather than mutating the existing `workersByJobRole` property in the component. 
We could do something like this to achieve the same result:

```suggestion
    const updatedWorkers = this.workersByJobRole.groups.map((group, index) => {
      const formEntry = workersFormValues[index];
      const annualHourlyPay = { value: formEntry.value, rate: formEntry.rate };

      return { ...group, annualHourlyPay };
    });
    this.workerService.setWorkersGroupedByJobRole({ groups: updatedWorkers });
```

kapppa-joe (approved):

PR #6858, opened 2026-04-10
Feat/fast track confirm page

Zuhal-Ayob:
#### Work done
- Add the functionality for updating the pay confirmation page.

#### Tests
- [x] Yes


frontend/src/app/features/workers/fast-track-confirmation-page/fast-track-confirmation-page.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.workersByJobRole.groups.forEach((group) => {
+      workersFormArray.push(
+        this.formBuilder.group({
+          workerId: group.jobId,
+          value: group.annualHourlyPay?.value || null,
+          rate: group.annualHourlyPay?.rate || null,
+        }),
+      );
+    });
+  }
```
kapppa-joe: Do we need a form for this confirmation page? In the html template I couldn't find an input form

frontend/src/app/features/workers/fast-track-confirmation-page/fast-track-confirmation-page.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    private alertService: AlertService,
+    private establishmentService: EstablishmentService,
+  ) {}
+
+  ngOnInit(): void {
+    this.setBackLink();
+    this.workplace = this.route.snapshot.data.establishment;
+    this.workersByJobRole = this.workerService.getWorkersGroupedByJobRole();
+    this.jobRoleIndex = this.route.snapshot.queryParams['index'];
+    this.jobRoleChanged = this.workersByJobRole?.groups[this.jobRoleIndex];
```
kapppa-joe: Are these two variables in use? I can't see a reference from html template or elsewhere. When I actually open the page these two seems to be undefined

frontend/src/app/features/workers/fast-track-confirmation-page/fast-track-confirmation-page.component.html
```diff
@@ -0,0 +1,75 @@
+@if (submitted && form.invalid) {
+<app-error-summary [formErrorsMap]="formErrorsMap" [form]="form"></app-error-summary>
+}
```
kapppa-joe: do we have validation in this confirmation page? I took a look but can't find validation there, and also can't find a formErrorsMap property as well

frontend/src/app/features/workers/fast-track-confirmation-page/fast-track-confirmation-page.compoenet.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  async function setup(overrides: any = {}) {
+    const establishment = establishmentBuilder() as Establishment;
+    const mockGroups = overrides.groups ?? [
+      {
+        jobId: 1,
+        title: 'Care worker',
+        count: 2,
+        workers: [{ uid: '1' }, { uid: '2' }],
+        annualHourlyPay: { value: 'HOURLY', rate: 10 },
+      },
```
kapppa-joe: Just a small note, the annualHourlyPay probably has to be one of ("Annually", "Hourly", "Don't know"). Seems that the backend would reject all capital "HOURLY".
This is mock data for test so wouldn't cause any error, but might be potentially confusing if we look back later

frontend/src/app/features/workers/fast-track-confirmation-page/fast-track-confirmation-page.compoenet.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    component.onSubmit();
+
+    expect(updateWorkersSpy).toHaveBeenCalled();
+
+    const payload = updateWorkersSpy.calls.mostRecent().args[1];
+
+    expect(payload.length).toBe(2);
+    expect(payload[0]).toEqual({
+      uid: '1',
+      annualHourlyPay: { value: 'HOURLY', rate: 10 },
```
kapppa-joe:
```suggestion
      annualHourlyPay: { value: 'Hourly', rate: 10 },
```

frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.ts
```diff
@@ -22,12 +22,25 @@ export class FastTrackPayUpdatesComponent {
...
+    let serviceData = this.workerService.getWorkersGroupedByJobRole();
+
+    if (!serviceData || !serviceData.groups) {
+      // Initial flow: populate service from resolver
+      serviceData = this.route.snapshot.data.workersByJobRole;
+      this.workerService.setWorkersGroupedByJobRole({ groups: serviceData.groups });
+    }
+
+    // Use service data as source of truth
+    this.workersByJobRole = serviceData;
```
kapppa-joe: Some concern related to this part:

These lines are about prefilling the form with previous value, shall we add a unit test for it?

When we do something like this, I think we need to clear the temp data in workerService at the correct timing. Otherwise it might cause some trouble.
For example, if a user has parent & subsidiary workplace, and then use this form to update all pay for parent workplace and then the subsidiary, then in the subsidiary this could incorrectly load the workersByJobRole from parent account

frontend/src/app/features/workers/fast-track-pay-updates/fast-track-pay-updates.component.ts
```diff
@@ -37,32 +50,65 @@ export class FastTrackPayUpdatesComponent {
...
+          return {
+            ...group,
+            annualHourlyPay: {
+              value: formEntry.value,
+              rate: formEntry.rate,
+            },
+          };
+        }
+        return group; // keep others untouched
+      });
```
kapppa-joe: Hi, I noticed something odd while testing this bit in the browser.
If I use the 'change' link for e.g. Care worker to go back to the form, and then update the pay for several job roles, only the Care worker updates are brought to the confirmation page, and the change to other job roles are dropped silently. That feels a bit confusing. Was that required by the design?

kapppa-joe (approved):

PR #6897, opened 2026-06-24
Feat/1942: Add staff records summary message

Praveena-Prasannan:
#### Work done
- Updated staff records summary message text and wording
- Added new `LastStaffRecordMessageDismissedAt` column in Establishment
- Implemented rule change: “You’ve not added any staff records in the last 12 months” message disappears on click and restarts the countdown
- Fetched the new column value from API response and used it in the UI
- Updated UI logic to determine when the summary message should be shown or hidden
- Added functionality to dismiss summary messages
- Corrected API call used to update message dismissal state
- Updated component and service unit tests to cover new behaviour

#### Tests
- [x] Yes


frontend/src/app/shared/components/summary-section/summary-section.component.ts
```diff
@@ -101,8 +101,23 @@ export class SummarySectionComponent implements OnInit, OnDestroy {
...
+  public async onClick(
+    event: Event,
+    fragment: string,
+    route: string[],
+    skipTabSwitch: boolean = false,
+    message?: string,
+  ): Promise<void> {
     event.preventDefault();
+
+    if (message === 'You’ve not added any staff records in the last 12 months') {
```
kapppa-joe: For this one, the string 'You’ve not added any staff records in the last 12 months' appears 2 times in this file. Could we assign this string to a const in this file and refer to it in both places?  So that in case if this message need to be changed in future we only need to change it in one place.

frontend/src/app/shared/components/summary-section/summary-section.component.ts
```diff
@@ -212,20 +227,22 @@ export class SummarySectionComponent implements OnInit, OnDestroy {
     }
 
     const afterWorkplaceCreated = dayjs(this.workplace.created).add(12, 'M');
+    const lastStaffRecordMessageDismissedAt = dayjs(this.workplace.lastStaffRecordMessageDismissedAt).add(12, 'M');
```
kapppa-joe: this const is actually the dismissedAt date + 12 months, so it may be better to rename it to reflect that

frontend/src/app/shared/components/summary-section/summary-section.component.ts
```diff
@@ -306,9 +323,13 @@ export class SummarySectionComponent implements OnInit, OnDestroy {
...
   private updateSingleEstablishmentField(dataToUpdate: any): void {
-    this.subscriptions.add(
-      this.establishmentService.updateSingleEstablishmentField(this.workplace.uid, dataToUpdate).subscribe(),
-    );
+    this.establishmentService.updateSingleEstablishmentField(this.workplace.uid, dataToUpdate).subscribe((response) => {
+      if (!response?.data) {
+        return;
+      }
+
+      this.workplace.lastStaffRecordMessageDismissedAt = response.data.lastStaffRecordMessageDismissedAt;
```
kapppa-joe: here may be better to add a check that `response.data.lastStaffRecordMessageDismissedAt` is defined before setting it to workplace object, because there are other places which call the updateSingleEstablishmentField() method

frontend/src/app/shared/components/summary-section/summary-section.component.ts
Praveena-Prasannan: Done.

frontend/src/app/shared/components/summary-section/summary-section.component.ts
Praveena-Prasannan: Done

Praveena-Prasannan:
All requested changes addressed. Ready for re-review

kapppa-joe (approved):

PR #6913, opened 2026-07-13
Feat/1958 Update homepage tiles

Praveena-Prasannan:
#### Work done
- Updated homepage tile styling and content for Parent, Subsidiary, and Standalone views.
- Added new Access Funding tile SVG asset.
- Added dynamic year logic for homepage content.
- Updated wording across homepage tiles:
Bulk Upload
Funding Checker
Benchmarks
Benefits Bundle
ASC-WDS Certificates
News
Image 13-07-2026 at 09 58
Image 13-07-2026 at 10 00
Image 13-07-2026 at 10 01

#### Tests
- [x] Yes


frontend/src/assets/scss/components/_asc-card.scss
```diff
@@ -0,0 +1,22 @@
...
+  width: 100%;
+}
+.asc-card-break {
+  display: none;
+}
+
+@media (min-width: 769px) {
+  .asc-card-break {
+    display: inline;
+  }
```
kapppa-joe, line 21: For these css rules, could we put them in one of the shared scss files under `frontend/src/assets/scss/` ? So that we don't have to repeat these in 3 places
(I think we can either put these in one of the existing scss files there, or add a new scss file and then import in `main.scss`)

kapppa-joe (approved):

PR #6920, opened 2026-07-22
Add a new check to pass the validation when there is a new establishment

Zuhal-Ayob:
#### Work done
PR Summary

 **The issue :**

- The transfer process attempted to resolve the destination workplace during validation, before the new workplace had been created in the database. As a result, the destination workplace ID was unavailable and the worker remained associated with the original workplace.

**What should happen:**

- When transferring a staff record to a workplace created within the same bulk upload, the workplace should be created first and the worker should then be transferred to the newly created workplace.

**Fix:**

- Updated the bulk upload completion process so that workers with TRANSFERSTAFFRECORD are matched to newly created workplaces after establishment creation. The newly created establishment ID is then assigned to the worker before saving, allowing the transfer to complete successfully.

#### Tests
- [x] Yes

### DIFF
--- a/backend/server/routes/establishments/bulkUpload/complete.js
+++ b/backend/server/routes/establishments/bulkUpload/complete.js
@@ -68,6 +68,13 @@ const completeUpdateEstablishment = async (
       delete thisEstablishmentJSON.localIdentifier;
 
       await foundCurrentEstablishment.load(thisEstablishmentJSON, true, true);
+      Object.values(foundCurrentEstablishment._workerEntities || {}).forEach((worker) => {
+        const targetEstablishment = onloadEstablishments.find((e) => e.name === worker.transferStaffRecord);
+
+        if (targetEstablishment) {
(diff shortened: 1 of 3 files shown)


Praveena-Prasannan (approved):

backend/server/routes/establishments/bulkUpload/complete.js
```diff
@@ -68,6 +68,13 @@ const completeUpdateEstablishment = async (
       delete thisEstablishmentJSON.localIdentifier;
 
       await foundCurrentEstablishment.load(thisEstablishmentJSON, true, true);
+      Object.values(foundCurrentEstablishment._workerEntities || {}).forEach((worker) => {
+        const targetEstablishment = onloadEstablishments.find((e) => e.name === worker.transferStaffRecord);
```
kapppa-joe: Is it correct to compare with e.name here?  I think the value in TransferStaffRecord is the LOCALESTID, not establishment name

PR #6921, opened 2026-07-23
Feat/1970 error handling when directus is down

kapppa-joe:
#### Work done
- Add a special case in frontend `HttpErrorHandler`, so that it would not navigate to error page when backend GET /cms endpoint return an error response
- Add logic to cache CMS content in Redis ( = valkey / elasticCache) and serve CMS content from the cache when directus is down
- Store CMS base urls in config yaml files instead of hardcoded in backend code

#### Tests
- [x] Yes


backend/server/utils/cacheCMSResponse.js
Zuhal-Ayob, line 1: I think there is a spelling error in the file name , is it mean to be cacheCMSResponse rather than cacheCMSReponse?

Zuhal-Ayob (approved):

PR #6928, opened 2026-08-06
Fix/1972 user being signed out when workplace permission change

kapppa-joe:
#### What is the issue: 
When the parent sets permissions to "no access to data, linked only" for its workplace, the user on the sub workplace gets signed when they refresh and as soon as they try and sign in.  

#### Work done
- Address the above issue, so that refresh will not trigger sign out in the above situation.
- Add a special logout message to better inform user when they are signed out by a 403 response 

#### Notes:
The sign out behaviour is triggered by 403 forbidden response from backend server.
On page refresh, AppHeader try to call `getAllUsersForEstablishment` to determine whether to show a flag on User link, but workplace of no access to data is not allowed to see user list, thus causing the 403 response.
In this PR, The "signed out on refresh" problem is addressed by adding a permission check before calling `getAllUsersForEstablishment`.

However, if instead of page refresh, if the user who got permission taken away tried to click on workplace tab / staff records tab / other links that require permission, they will still get a 403 error. 
This is a legitimate case, and for security reason we shouldn't change this.
A special message is added to logout component to guide user to sign in again in this case.

#### Tests
- [x] Yes


frontend/src/app/features/logout/logout.component.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    const { component } = await setup();
+    expect(component).toBeTruthy();
+  });
+
+  it('should show a message on sign out', async () => {
+    const { getByText } = await setup();
+    expect(getByText('You have been signed out.')).toBeTruthy();
+  });
+
+  it('should show a special message if signed out by 403 response from servr', async () => {
```
Praveena-Prasannan: Minor typo in the test description - servr → server.

Praveena-Prasannan (approved):

